PR #31, opened 2021-06-11
Add Extension Type support

### LINKED ISSUE #26: [FEA] A fallback mechanism
Feature request, make it possible to specify a fallback function that is called when the encoder encounters an unsupported object. 
The fallback function can then handle the unsupported object using only msgspec compatible types. When decoding, a corresponding fallback function are then called to handle the decoding of the object.

#### Motivation
In Dask/Distributed we are discussing the replacement of `msgpack`. The library should be:
- fast
- secure (no arbitrary code execution when decoding)
- support basic Python types
- not convert list to tuples (can be disabled in `msgpack` but with a performance penalty)
- not convert numpy scalars to python scalars
- support a fallback mechanism

jcrist:
This adds support for both encoding and decoding [msgpack extension types](https://github.com/msgpack/msgpack/blob/master/spec.md#ext-format-family). This comes with 2 additional apis:

- A new `msgspec.Ext` type, representing the tuple of `(code: int, data: bytes_like)` that describes a msgpack extension.
- A new `ext_hook` parameter to `msgspec.Decoder`/`msgspec.decode`.

By default msgspec will deserialize all extension objects into `msgspec.Ext` objects. However, if a user specifies a `ext_hook` to `msgspec.Decoder`/`msgspec.decode`, this callback will be used instead to deserialize the extension type. Together with the existing `default` hook in `msgspec.Encoder`/`msgspec.encode`, this lets users fully add extra types to be serialized/deserialized.

For example, say we wanted to define a new extension type with code `42` that contains a buffer of data serialized via `pickle`. If we wanted to apply this to all non-builtin types serialized by msgspec, we could do this as follows:

```python
import msgspec
import pickle
from typing import Any

def default(obj: Any) -> Any:
    """Serialize the object using pickle, and mark it as an extension type with code 42"""
    return msgspec.Ext(42, pickle.dumps(obj))

def ext_hook(code: int, data: memoryview) -> Any:
    if code == 42:
        return pickle.loads(data)
    raise TypeError(f"Unknown extension type with code {code}")

enc = msgspec.Encoder(default=default)
dec = msgspec.Decoder(ext_hook=ext_hook)

buf = enc.encode(range)  # functions normally aren't serializable with msgspec
out = dec.decode(buf)
assert out == range
```

Note that `ext_hook` has the signature `ext_hook(code: int, data: memoryview) -> Any`. Here `data` is a memoryview into larger message buffer. This lets us avoid an extra copy when forwarding the data buffer to `ext_hook`, but it means that if the returned object holds on to a reference to that buffer (instead of doing a copy itself) then the larger message will persist in memory until that object is freed. For most use cases of `ext_hook` I wouldn't expect this issue to come up, but it's worth noting. The performance improvement from avoiding the copy is worth it to add a tiny rare footgun IMO.

Fixes #26.

Still a few todos:

- [x] Make `Ext` accept a `memoryview` for `data`, rather than just `bytes` or `bytearray`
- [x] Add docs

### DIFF
--- a/docs/source/api.rst
+++ b/docs/source/api.rst
@@ -24,6 +24,13 @@ Struct
     :members:
 
 
+Ext
+---
+
+.. autoclass:: Ext
+    :members:
+
+
 Functions
 ---------
 
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -41,7 +41,7 @@ Highlights
   excellent editor integration.
 - ``msgspec`` is **flexible**. Unlike other libraries like ``msgpack`` or
   ``json``, ``msgspec`` natively supports a wider range of Python builtin
-  types.
+  types. Support for additional types can also be added through :ref:`extensions`.
 - ``msgspec`` supports :ref:`"schema evolution" <schema-evolution>`. Messages can
   be sent between clients with different schemas without error.
 
@@ -104,6 +104,10 @@ Msgspec currently supports serializing/deserializing the following types:
 - `set`
 - `enum.Enum`
 - `msgspec.Struct`
+- `msgspec.Ext`
+
+Support for serializing additional types can be added through the use of the
+``default`` callback on the `Encoder`, or by defining custom :ref:`extensions`.
 
 .. _typed-deserialization:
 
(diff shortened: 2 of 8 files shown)


jcrist:
Hmmm, looks like there's a segfault somewhere, haven't hit that locally. Will debug further later.

PR #33, opened 2021-06-22
Add `dec_hook` to `Decoder`/`decode`

jcrist:
Adds a new `dec_hook` callback supporting type conversions between
builtin msgpack types and arbitrary python types. This hook should have
the following signature:

```python
def dec_hook(obj: Any, type: Type) -> Any:
    ...
```

This receives an object composed of builtin msgpack types (e.g. dict,
list, int, ...) and the expected decode type (as described by the
decoder's type parameter). It should attempt to convert the value to the
provided type, or raise a type error appropriately. When combined with
`enc_hook` on the encoder, this lets users serialize arbitrary custom
objects without resorting to extension types.

For example, to serialize/deserialize a `NamedTuple` you might write
the following:

```python
import msgspec
from typing import NamedTuple

class Point(NamedTuple):
    x: int
    y: int

def enc_hook(obj):
    if isinstance(obj, tuple):
        # convert the named tuple to a supported type
        # in this case it's a tuple like `(1, 2)`
        return tuple(obj)
    raise TypeError(f"Type {type(obj).__name__} is not supported")

def dec_hook(obj, type):
    if issubclass(type, tuple):
        # convert the builtin type (e.g. `[1, 2]`) to
        # the custom type (`Point(x=1, y=2)`)
        return type(*obj)
    raise TypeError(f"Type {type.__name__} is not supported")

enc = msgspec.Encoder(enc_hook=enc_hook)
dec = msgspec.Decoder(Point, dec_hook=dec_hook)

msg = enc.encode(Point(1, 2))
x = dec.decode(msg)
print(x)  # Point(x=1, y=2)
```

There's still a few TODO items, but this is mostly functional already
- [x] Tests
~- [ ] Docs~
- [x] Better support `GenericAlias` objects for custom types

### DIFF
--- a/msgspec/core.c
+++ b/msgspec/core.c
@@ -3472,7 +3472,7 @@ mp_decode_type_custom(DecoderState *self, bool generic, TypeNodeObj* type, TypeN
     else {
         out = obj;
     }
-    
+
     /* Generic classes must be checked based on __origin__ */
     if (generic) {
         MsgspecState *st = msgspec_get_global_state();
--- a/msgspec/core.c
+++ b/msgspec/core.c
@@ -112,7 +112,7 @@ msgspec_get_state(PyObject *module)
 /* Find the module instance imported in the currently running sub-interpreter
    and get its state. */
 static MsgspecState *
-msgspec_get_global_state()
+msgspec_get_global_state(void)
 {
     return msgspec_get_state(PyState_FindModule(&msgspecmodule));
 }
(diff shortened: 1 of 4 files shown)


msgspec/core.c
```diff
@@ -582,10 +587,15 @@ to_type_node(PyObject * obj, bool optional) {
     }
 
     /* Attempt to extract __origin__/__args__ from the obj as a typing object */
-    origin = PyObject_GetAttrString(obj, "__origin__");
-    if (origin == NULL) goto done;
-    args = PyObject_GetAttrString(obj, "__args__");
-    if (args == NULL) goto done;
+    if ((origin = PyObject_GetAttrString(obj, "__origin__")) == NULL ||
+            (args = PyObject_GetAttrString(obj, "__args__")) == NULL) {
```
jcrist, line 591: TODO: pull these out into interned string objects

jcrist:
TODO: switch the order of parameters in `dec_hook`. Since the user will most likely be dispatching on the type, it'd be nice if this parameter was first to mirror the convention in other dispatch decorators (e.g. `functools.singledispatch`).

jcrist:
Putting off writing docs til a later PR. This is good for now. Merging.

PR #37, opened 2021-07-02
Optimize buffer usage in `encode`/`encode_into`

jcrist:
This applies a few buffer-oriented optimizations to speedup encoding (by
up to a factor of ~2x).

- Move resizing of write buffer to a separate function, and inline
`mp_write`. This allows gcc to do a better job of optimizing small
memcpy operations with little code bloat.

- When returning a `bytes` object from `encode`, don't hit realloc to
shrink the buffer down to the exact output size. Instead, we just shrink
the view presented to Python. This requires a bit of mucking in the
internals of the bytes object, but nothing too onerous. Since the output
of `encode` will usually be immediately written to a socket/file and
then dropped, this shouldn't result in an increase in application memory
usage.

### DIFF
--- a/msgspec/core.c
+++ b/msgspec/core.c
@@ -24,6 +24,22 @@
     (PyType_IS_GC(Py_TYPE(x)) && \
      (!PyTuple_CheckExact(x) || IS_TRACKED(x)))
 
+/* Fast shrink of bytes & bytearray objects. This doesn't do any memory
+ * allocations, it just shrinks the size of the view presented to Python. Since
+ * outputs of `encode` should be short lived (immediately written to a
+ * socket/file then dropped), this shouldn't result in increased application
+ * memory usage. */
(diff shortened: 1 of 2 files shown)


jcrist:
Benchmark:

```python
import gc
import timeit
import msgspec


def time(func, data):
    gc.collect()
    timer = timeit.Timer("func(data)", globals={"func": func, "data": data})
    n, t = timer.autorange()
    return t / n


data = b"x" * 10
msg = [data] * 100000

enc = msgspec.Encoder()

t = time(enc.encode, msg)
print(f"msgspec took {t * 100000} us per call")
```

**Master**

```
$ python bench.py
msgspec took 124.0671865089098 us per call
```

**This PR**

```
$ python bench.py
msgspec took 73.82942020194605 us per call
```

PR #44, opened 2021-07-09
Use a freelist for struct types

jcrist:
This adds a freelist for "small" struct objects (currently defined as
`1 <= nfields <= 10`). Structs are added to the freelist during
deallocation, up to a per-size cap (currently 2000). Structs can then be
allocated from the freelist (if non-empty), skipping the cost of a
malloc.

To prevent keeping old pages around, the freelist is cleared every major
GC collection. This is done by a hook added to the module `m_traverse`
(since modules should persist for the lifetime of the program, they'll
usually be in the largest generation).

During high load, this leads to a measurable (4-6%) performance
improvement on deserialization.

### DIFF
--- a/msgspec/core.c
+++ b/msgspec/core.c
@@ -729,6 +729,129 @@ static PyTypeObject StructMixinType;
 #define OPT_TRUE 1
 #define STRUCT_MERGE_OPTIONS(opt1, opt2) (((opt2) != OPT_UNSET) ? (opt2) : (opt1))
 
+/* To reduce overhead of repeatedly allocating & freeing messages (in e.g. a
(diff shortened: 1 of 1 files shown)


jcrist:
Running the benchmark script gives:

**master**

```
- msgspec...
  dumps: 224.36 us
  loads: 688.29 us
- msgspec asarray...
  dumps: 154.70 us
  loads: 453.03 us
```

**This PR**

```
- msgspec...
  dumps: 222.62 us
  loads: 622.18 us
- msgspec asarray...
  dumps: 152.51 us
  loads: 435.89 us
```

For 10% and 4% speedups on load time respectively.
```

PR #46, opened 2021-07-12
[WIP] JSON support

jcrist:
This is a WIP pass at adding JSON support to msgspec. Currently there's only an encoder.

TODO:
- [ ] Encoder
- [ ] Decoder
- [ ] Add tests


jcrist:
I've done _some_ optimization work already, but there's likely more that can still be done. Currently it looks like this is on-par with orjson - some messages msgspec is faster, and some orjson is faster. I haven't identified the slowdowns on our end yet, but I'm sure they're solvable. Note that struct-based messages serialize much faster than dicts (due to limitations on python identifiers, we can take a fast path for serializing keys).

jcrist:
Superseded by #56. Closing.

PR #49, opened 2021-07-20
Json encoder optimizations

llllllllll:
Full explanation here: https://gist.github.com/llllllllll/b3ccc744e88c5bf0ee4e2c16c6fcdebf

I realize that these changes make the code harder to read, so I understand if you don't think the minor performance improvements are worth it. I figured I would share what I did in case you did want it.


jcrist:
Thanks for this (and the very thorough writeup)! I'm going to merge this in now and iterate on it from there.

PR #55, opened 2021-11-01
Support serializing memoryview objects

### LINKED ISSUE #52: Add support for `typing.ByteString`/`collections.abc.ByteString`
When parsing type signatures for structs, it'd be useful to support a more abstract bytes-like object. For encoding this would result in no major changes (encoding only deals with concrete types). For decoding, this would decode the type as `bytes` (unless an extension type was used).

jcrist:
We support encoding memoryview objects as binary blobs, but don't support decoding into `memoryview` objects for typed decoders. We could add support pretty easily, but since:

- memoryviews are usually an optimization to avoid copying (and a copy is necessary on decode)
- mypy will accept passing a `memoryview` to a signature annotated `bytes` already, so encode-side annotations are fine.
 
I don't see decode support as being necessary now. For all uses decoding into a `bytes` or `bytearray` makes more sense IMO.

Fixes #52 (well, kind of. I don't think the original request there makes sense anymore).

### DIFF
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -40,6 +40,7 @@ types:
 - `str`
 - `bytes`
 - `bytearray`
+- `memoryview`
 - `tuple`
 - `list`
 - `dict`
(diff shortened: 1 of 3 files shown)


jakirkham:
cc @madsbk (for awareness)

PR #66, opened 2022-02-11
[DOC] Fix typo in performance tips

fjetter:
I assume this indentation is wrong in the examples


jcrist:
No, this is correct as is. Instead of reallocating a new buffer for every message in the stream, we use a single shared buffer to encode into and write out for every message.

fjetter:
oh, I got it. I assume you somehow reset the buffer since otherwise we should receive truncated data somewhere, shouldn't we?

PR #77, opened 2022-02-28
Micro-optimize `decode(type=Struct)`

jcrist:
For the common case of `msgspec.json.decode(msg, type=Struct)`, we can
allocate the `TypeNode` on the stack and have a faster init path,
reducing overhead per call. This still isn't as fast as creating a
`Decoder` once and using `Decoder.decode`, but it brings things closer.
On my machine, decoding a simple `Point(x=1, y=2)` struct takes the
following times:

- `decode` (master): 240 ns
- `decode` (this PR): 150 ns
- `Decoder.decode`: 120 ns

### DIFF
--- a/msgspec/_core.c
+++ b/msgspec/_core.c
@@ -6600,9 +6600,15 @@ msgspec_msgpack_decode(PyObject *self, PyObject *const *args, Py_ssize_t nargs, 
     }
     state.ext_hook = ext_hook;
 
-    /* Only build TypeNode if required */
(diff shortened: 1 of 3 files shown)


jcrist:
Going to sit on this one for a bit, I'm still not sure I want the (mild) added complexity in favor of the (mild) performance improvement.

jcrist:
Upon further thought, I do think this is worth it. Merging.

PR #80, opened 2022-02-28
Upgrade cibuildwheel

jcrist:
Maybe this will fix the missing Python 3.10 wheel builds.


jakirkham:
Yeah I think Python 3.10 was added in a newer version. Went through a similar thing with Numcodecs

PR #83, opened 2022-03-01
Tagged Unions

jcrist:
This PR adds tagged union support, enabling decoding into a union of struct types.

If either `tag` or `tag_field` are non-None (the default), then a struct is considered "tagged". In this case, a new field (the `tag_field`) is added to its serialized representation, with a unique corresponding value (the `tag`) per struct type. During decoding the `tag` is decoded first and is used to lookup the struct type in an internal hashmap - this type is then used to decode the remainder of the fields. This is both efficient (especially if the tag is the first field in the message) and unambiguous (unlike the YOLO method some other libraries employ where each type is tried in order until one succeeds).

A user can pass in `tag=True` to enable the default tagging configuration. In this case, `tag_field="type"` and `tag` is the class name. These values can be overridden by manually passing in strings for either (e.g. `tag_field="some other field name"`). `tag` also accepts a callable, mapping the class name to a new string to use for the tag (e.g. `tag=lambda name: name.lower()` to use the class name but lowercase).

Example:

```python
from msgspec import Struct
from typing import Literal, Union


class Get(Struct, tag_field="op"):
    key: str


class Put(Struct, tag_field="op"):
    key: str
    val: str


class Del(Struct, tag_field="op"):
    key: str


msgspec.json.encode(Get("foo"))
# b'{"op": "Get", "key": "foo"}'

msgspec.json.decode(
    b'{"op": "Put", "key": "foo", "val": "bar"}',
    type=Get | Put | Del
)
# Put(key="foo", val="bar")
```


jcrist:
cc @goodboy, I think you asked about something like this a few months ago.

goodboy:
Woot, yeah i think so too. Will def check it out stat!

goodboy:
Yeah, thankfully still had this tab 😎 

I need this exactly right now 😂

PR #104, opened 2022-04-26
Adding Linux and Mac ARM wheels

### LINKED ISSUE #103: Build wheels for Linux/Mac ARM
With the Apple M1 using ARM, it would be valuable to support Linux and Mac ARM builds in the CI. I can take a look at this when I get a chance. I suspect is should be a small change to add the builds in the GitHub actions.

cjermain:
This resolves #103 by adding steps for Linux and Mac ARM wheels in the CI build. I also bumped the version of pypa/cibuildwheel to v2.4.0.

Mac support: https://cibuildwheel.readthedocs.io/en/stable/faq/#apple-silicon
Linux support: https://cibuildwheel.readthedocs.io/en/stable/faq/#emulation

I don't have a great way of testing this. :smile:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,13 +52,22 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
+      - name: Set up QEMU
+        if: runner.os == 'Linux'
+        uses: docker/setup-qemu-action@v1
+        with:
+            platforms: all
+
       - name: Build & Test Wheels
-        uses: pypa/cibuildwheel@v2.3.1
+        uses: pypa/cibuildwheel@v2.4.0
         env:
           CIBW_TEST_REQUIRES: "pytest msgpack"
           CIBW_TEST_COMMAND: "pytest {project}/tests"
           CIBW_BUILD: "cp38-* cp39-* cp310-*"
           CIBW_SKIP: "*-win32 *_i686 *_s390x *_ppc64le"
+          CIBW_ARCHS_MACOS: x86_64 arm64
+          CIBW_ARCHS_LINUX: x86_64 aarch64
+          CIBW_TEST_SKIP: "*_arm64"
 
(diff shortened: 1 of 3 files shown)


cjermain:
@jcrist, what do you think?

jcrist:
Thanks @cjermain! Apologies for the delay here - it looks like there are some test failures on the new architectures. I'm not sure what's going on - hopefully will find some time to debug before the next release. If you're interested in debugging here, I'd welcome the help, though so far this has been mostly a single-dev codebase and the internals are mostly undocumented.

cjermain:
It looks like the CI issues were related to 2 changes in the main branch since I forked. I've merged upstream into this, and it would be great to kick off the builds again.

cjermain:
It looks like there are still some errors when testing the Linux ARM wheel:
https://github.com/jcrist/msgspec/runs/6528746892?check_suite_focus=true#step:4:821

solsen2142:
After testing with the other architectures, it looks the only issue is with aarch64 for Linux. If we only want MacOS for now, we could come back to Linux aarch64 later.

cjermain:
Thanks @solsen2142! I'm hoping to run this on a Linux ARM VM on top of a Mac M1 chip, so for me the Linux build is a priority.

jcrist:
Thanks for your work here @cjermain - I've pushed up a patch that should (hopefully) fix the linux ARM builds. The main issue was that apparently C leaves it up to the implementation whether `char` is a signed or unsigned type. On x86 it's generally a signed type, but on linux ARM it's unsigned. OSX ARM defines it as signed for compatibility with x86 OSX.

jcrist:
Tests are green, merging. Unfortunately, this makes the linux builds ~8x slower. I'll push a followup PR to trim down the builds for non-releases. We only really need to test on linux once for every python version + one build on aarch64 to catch architecture-specific bugs.

PR #109, opened 2022-05-30
Use `asan` and `ubsan` in CI

jcrist:
- Add `asan`/`ubsan` build & tests to CI
- Fixes bugs found when running with `asan`/`ubsan`, as well as silencing some warnings found by running clang's static analysis locally.


jcrist:
This currently disables the leak sanitizer in asan. I'd like to use the leak sanitizer as well, but Python leaks memory on exit normally, so there's a lot of false positives in the logs.

PR #121, opened 2022-06-17
Use dict str cache for msgpack decoder

jcrist:
Previously we added a cache for short ascii string keys when decoding
dicts using JSON. We now expand this optimization to the msgpack decoder
as well.


jcrist:
With this change, the standard twitter.json dataset loads ~30% faster when using the untyped decoder. We still recommend using `Struct` types when data has a known schema, but this optimization dramatically speeds up decoding when loading structured data (dicts have a common, repeated structure) but no schema is provided.

PR #141, opened 2022-07-07
Clarify `__hash__` behavior when `frozen=True`

### LINKED ISSUE #139: Freezing a `Struct` containing a sub-nonhashable
I think I've more or less had the exact same issue with `frozen=True` in `pydantic` as well but this is the issue:

```python
[ins] In [1]: import msgspec

[ins] In [2]: class Msg(msgspec.Struct, frozen=True):
         ...:     l: list[str]
         ...:

[ins] In [3]: msg = Msg([str(i) for i in range(10)])

[ins] In [4]: {msg: 10}
TypeError                                 Traceback (most recent call last)
Input In [4], in <cell line: 1>()
----> 1 {msg: 10}

TypeError: unhashable type: 'list'
```

I realize that there might not be an easy away around this (without some kind of nested hashable obj scanner) but I figured I'd report since there's [no mention of this restriction in the docs](https://jcristharif.com/msgspec/structs.html#frozen-instances).

jcrist:
Fixes #139.

### DIFF
--- a/docs/source/structs.rst
+++ b/docs/source/structs.rst
@@ -210,8 +210,9 @@ Frozen Instances
 ----------------
 
 A struct type can optionally be marked as "frozen" by specifying
-``frozen=True``. This disables modifying attributes after initialization,
-and adds a ``__hash__`` method to the class definition.
+``frozen=True``. This disables modifying attributes after initialization, and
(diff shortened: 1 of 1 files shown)


goodboy:
Yup looks good to me!

PR #144, opened 2022-07-14
Add codecov

jcrist:


.github/workflows/ci.yml
```diff
@@ -27,7 +27,7 @@ jobs:
 
       - name: Install Dependencies
         run: |
-          pip install pytest mypy pyright msgpack
+          pip install pytest mypy pyright msgpack coverage
```
jcrist, line 30: Install `coverage` to provide coverage for python files

.github/workflows/ci.yml
```diff
@@ -55,7 +56,14 @@ jobs:
           ASAN_OPTIONS: "detect_leaks=0"
         run: |
           export LD_PRELOAD=`gcc -print-file-name=libasan.so`
-          py.test -s -m "not mypy and not pyright"
+          coverage run -m pytest -s -m "not mypy and not pyright"
+          coverage xml
+          gcov -abcu `find build/ -name *.o`
```
jcrist, line 61: Here we:

- Run pytest via `coverage` to enable python coverage support
- Generate a `coverage.xml` file containing coverage for all Python source files
- Use `gcov` (which comes standard with gcc) to generate `.gcov` files for every source file in our build.

.github/workflows/ci.yml
```diff
@@ -55,7 +56,14 @@ jobs:
...
+
+      - name: Upload Codecov
+        uses: codecov/codecov-action@v3
+        with:
+          files: coverage.xml,_core.c.gcov,atof.h.gcov,ryu.h.gcov
```
jcrist, line 66: This uses the github actions codecov action to upload coverage files to codecov.io.

We manually specify the list of files, since `gcov` generates a few additional files based on inline functions found in the cpython header files (e.g. one for `cpython/Include/object.h`). There may be a cleaner way of doing this, but `msgspec` only has 3 C source files, so manually listing them seems fine.

setup.py
```diff
@@ -5,15 +5,19 @@
...
 extra_compile_args = []
 extra_link_args = []
-if DEBUG:
-    extra_compile_args.extend(["-O0", "-g", "-UNDEBUG"])
 if SANITIZE:
     extra_compile_args.extend(["-fsanitize=address", "-fsanitize=undefined"])
     extra_link_args.extend(["-lasan", "-lubsan"])
+if COVERAGE:
+    extra_compile_args.append("--coverage")
+    extra_link_args.append("-lgcov")
```
jcrist, line 18: Here we add compile and link args to enable `gcov` support. These could have been added in CI only using `CFLAGS` and `LDFLAGS` respectively, but I find adding a top-level option in `setup.py` easier for me to use locally during development.

This lets you build the package with gcov support using:

```
MSGSPEC_COVERAGE=true python setup.py develop
```

jcrist (commented):
Adding a bunch of review comments to document the changes necessary here to support adding codecov.io coverage tracking to a Python project containing a C extension.

PR #150, opened 2022-07-21
maybe we accept subclass of dict or list

wuhanck:
sometimes we use data:
it is subclass of dict or list and the its element may also be subclass of dict or list.

can we loose the check to allow encode that data?
I dont know if there is performance penalty.


jcrist:
Maybe? This is something I've contemplated allowing, but am not sure if I want to support it.

The reason we currently don't is for consistency. Right now, any type that works by default when passed to an `encode` method can also be decoded when passed to `decode(msg, type=your_type)`. We can fairly easily support subclasses of `dict` or `list` for `encode`, but they're not easy to support for `decode`. This creates an inconsistency.

Currently you can use the `enc_hook` kwarg to support your custom types (see [extending docs](https://jcristharif.com/msgspec/extending.html)):

```python
In [1]: class mydict(dict):
   ...:     pass
   ...: 

In [2]: import msgspec

In [3]: data = mydict(a=1, b=2)

In [4]: msgspec.json.encode(data)  # errors by default
TypeError                                 Traceback (most recent call last)
Input In [4], in <cell line: 1>()
----> 1 msgspec.json.encode(data)

TypeError: Encoding objects of type mydict is unsupported

In [5]: def enc_hook(obj):
   ...:     if isinstance(obj, mydict):
   ...:         return dict(obj)
   ...:     raise TypeError(f"Encoding {obj} is unsupported")
   ...: 

In [6]: msgspec.json.encode(data, enc_hook=enc_hook)  # works now
Out[6]: b'{"a":1,"b":2}'
```

This works, but performance would be higher if we supported encoding subclasses of `dict`/`list` natively. Let me think on it for a bit, I don't want to add this feature and then regret it later.

Can you say more about why you use `dict`/`list` subclasses? What methods/features do your subclasses add?

wuhanck:
dict are only the access pattern we treat our data.
other aspect like validation, like fields value association, etc , need to subclass dict.
encode and decode really dont need to be symmetrical in type info, and consider language neutrality.
msgspec may  just assure dict/list/int/str/int/float encode to bytes and convert to dict/list/int/float/etc.
dict or list may get extra logical layer which handled by business logical...layer is layed here.
it is practical......

jcrist:
Thanks. I pushed a few fixups to this PR in #160. This is now in. I'm hoping to cut the next release in the next week or so.

wuhanck:
you are so nice. thank you very much.  some people still work in old unix favor, we use msgspec just encode/decode dict/list/int/float/str/bytes back and forth; other logic is handled by other layer. thanks for your understanding again.

PR #156, opened 2022-07-26
Add some naive cpython api linters

jcrist:
This adds some naive linting tests for easy to detect errors when using
the cpython capi. It also fixes some bugs exposed by these tests:

- Adds a test that every successful call to `Py_EnterRecursiveRepr`
doesn't return without calling `Py_LeaveRecursiveCall`
- Adds a test that every successful call to `Py_ReprEnter` doesn't
return without calling `Py_ReprLeave`


jcrist:
Failure is due to code coverage of the patch alone, which doesn't seem quite right.

PR #162, opened 2022-07-28
Fix decoding Optional[custom type]

### LINKED ISSUE #159: optional doesn't work with a custom class
i'm trying to decode a jsonrpc response which has an optional value which i need to process in `dec_hook`.

to do this i subclass `address` class from `str`, but get a decoding error when the value is `null`. i tried both `Optional[address]` and `address | None`. it works if i change to `Optional[str]`, but i can't do the processing.

the doc says "Unions with custom types are unsupported beyond optionality (i.e. Optional[CustomType])", which implies this should be possible.

i extracted a small [sample here](https://gist.github.com/banteg/49b45f574e850f13b46e609cba143806) which includes one working and one failing case. for the failing case i get:
```python
msgspec.DecodeError: Expected `address`, got `NoneType` - at `$.to`
```

a more complete spec of what im building [is here](https://gist.github.com/banteg/cee923b6616a8f70db1e18a61b91e39c) if it's useful.

jcrist:
This used to work, but lapsed. Looks like there never was a test that
hit this case :/.

Fixes #159.

### DIFF
--- a/msgspec/_core.c
+++ b/msgspec/_core.c
@@ -5743,6 +5743,8 @@ ms_decode_custom(PyObject *obj, PyObject *dec_hook, bool generic, TypeNode* type
 
     if (obj == NULL) return NULL;
 
+    if (obj == Py_None && type->types & MS_TYPE_NONE) return obj;
+
     custom_obj = TypeNode_get_custom(type);
 
     if (dec_hook != NULL) {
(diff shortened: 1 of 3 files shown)


jcrist:
codecov is being a bit too finicky on the patch coverage (one line decrease due to randomness!!?!). Ignoring, I'll push up a patch later to decrease the strictness here.

PR #175, opened 2022-08-26
[UNTESTED] Add support for renaming to kebap-case

ljrk0:
Hi,

this should implement builtin renaming to kebap-case.  I didn't test the code due to me virtually never working with Python and less with its C API, so it'd be good if someone else tested it.  Also, I only added one test case where I saw fit/understood the testing harness, more could be added but I wouldn't know how.

### DIFF
--- a/msgspec/__init__.pyi
+++ b/msgspec/__init__.pyi
@@ -41,7 +41,9 @@ class Struct(metaclass=__StructMeta):
         tag: Union[None, bool, str, int, Callable[[str], Union[str, int]]] = None,
         tag_field: Union[None, str] = None,
         rename: Union[
-            None, Literal["lower", "upper", "kebap", "camel", "pascal"], Callable[[str], str]
+            None,
+            Literal["lower", "upper", "camel", "pascal", "kebab"],
+            Callable[[str], str],
         ] = None,
         omit_defaults: bool = False,
         frozen: bool = False,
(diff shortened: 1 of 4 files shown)


msgspec/__init__.pyi
```diff
@@ -41,7 +41,7 @@ class Struct(metaclass=__StructMeta):
         tag: Union[None, bool, str, int, Callable[[str], Union[str, int]]] = None,
         tag_field: Union[None, str] = None,
         rename: Union[
-            None, Literal["lower", "upper", "camel", "pascal"], Callable[[str], str]
+            None, Literal["lower", "upper", "kebap", "camel", "pascal"], Callable[[str], str]
```
jcrist, line 44: TIL some people spell this "kebap" instead of "kebab". Since other tools like serde use the "kebab" spelling (https://serde.rs/container-attrs.html#rename_all), I think we should go with that one.

jcrist (commented):
Thanks for the PR! This mostly looks good to me, I can finish up the few final tweaks needed.

>  I didn't test the code due to me virtually never working with Python and less with its C API, so it'd be good if someone else tested it.

Out of curiosity - if you never work with Python and didn't test this feature locally, is this a feature you needed yourself? Or just something you noticed was missing?

msgspec/__init__.pyi
ljrk0: Ah, it seems like kebap is a bit more common in Germany as this is the spelling I see at my local Döner Kebap :-)

But I have absolutely nothing against naming it kebab, especially if this is more common

ljrk0:
> Thanks for the PR! This mostly looks good to me, I can finish up the few final tweaks needed.

Great!  Thanks for your work on maintaining this piece of software!

> > I didn't test the code due to me virtually never working with Python and less with its C API, so it'd be good if someone else tested it.
> 
> Out of curiosity - if you never work with Python and didn't test this feature locally, is this a feature you needed yourself? Or just something you noticed was missing?

I simply helped out a colleague who was struggling to find the right place in the API and defined them a simple `kebap_rename=lambda s: s.replace('_', '-')` for their usage, however, since a built-in was missing, I decided to quickly hack on it (they simply used the python implementation though) :-)

jcrist:
Thanks!

Issue #125, opened 2022-06-19
Support json-schema generation


old-ocean-creature:
It would be helpful if JSON schema generation was supported

```python
class User(msgspec.Struct):
    """A new type describing a User"""
    name: str
    groups: Set[str] = set()
    email: Optional[str] = None

schema = User.json_schema()
```

Similar to the functionality seen in https://github.com/s-knibbs/dataclasses-jsonschema

jcrist:
Thanks for opening this. This is definitely in scope, but isn't something I plan on spending much time on immediately.

Note that to avoid polluting the `Struct` method namespace, this should probably be a top-level method in the `msgspec.json` namespace instead. Something like `msgspec.json.json_schema` perhaps?

adriangb:
I think supporting json schema would be critical for msgspec to be an alternative to Pydantic

> to avoid polluting the Struct method namespace, this should probably be a top-level method in the msgspec.json namespace instead. Something like msgspec.json.json_schema perhaps?

👍  on this, I think having these methods be top-level is a much better idea. Ideally `Struct` would have no methods to avoid name collisions (e.g. someone wants a field called `json_schema`).

jcrist:
I'm not an active `json-schema` user - a couple questions/comments:

- We can fairly easily support generating a simple `json-schema` spec from a `msgspec` annotated type. I expect this to be no more than a couple hours work to get this functioning + tests + docs.
- We don't currently have a mechanism for adding extra info to the generated json-schema (`title`, `description`, ...). How important are those (or, more directly, if we release a version of `json_schema` that doesn't support those, will anyone find it still useful)? How do other systems handle this? What metadata fields like these should we support?

adriangb:
I think an initial implementation without the metadata would be a good place to start.

As to what metadata to include, I think it would be good to support title and description. There is also other metadata like "pattern" or "format", but that dovetails with constraints/validation. As to _how_ to express that metadata in Python, I would recommend PEP593 (see #154).

jcrist:
> As to how to express that metadata in Python, I would recommend PEP593 (see https://github.com/jcrist/msgspec/issues/154).

In your mind, how would the non-functional annotations (title, description) be specified on the schema (do these go on fields? On struct types themselves?)? An example user session would be helpful here.

adriangb:
maybe something like:

```python
PhoneNumber = Annotated[str, Pattern(r"\+1\d{10}"), Description("Phone number. US only. No dashes")]

[...]

PR #176, opened 2022-08-27
Support constrained type validations

### LINKED ISSUE #154: Support annotated-types for metadata and constraint specification
My understanding is that msgspec currently only supports basic type/schema validation, but not more complex validations like a regex pattern or arbitrary user functions, which Pydantic and other validators do support.

I think it would be really interesting if msgspec adopted [annotated-types](https://github.com/annotated-types/annotated-types) (maybe as an optional dependency), which would not only enable an API for these sorts of constraints but also:
- Improve interoperability with other libraries (Pydantic V2 will support annotated-types, so users could go back and forth or share types between the two libraries)
- Get automatic support for other parts of the ecosystem like Hypothesis

jcrist:
This adds support for constrained types, implementing the following constraints from [json-schema](https://json-schema.org/draft/2020-12/json-schema-validation.html):

- Numeric constraints `ge`, `gt`, `le`, `lt`, and `multiple_of` correspond to `minimum`, `exclusiveMinimum`, `maximum`, `exclusiveMaximum`, and `multipleOf` respectively
- `pattern` corresponds to `pattern`
- `min_length` and `max_length` correspond to `minLength`/`maxLength` or `minItems`/`maxItems` or `minProperties`/`maxProperties` depending on if the annotated type is a string, sequence, or mapping.

Constraints are added by wrapping a type with `Annotated`, and the special `msgspec.Meta` annotation. This is nice, since it works natively with `mypy` or `pyright`, and can be applied anywhere in the type hierarchy, not just on a struct field. These can be used to build up some fairly complicated validations.

For example, here we define a type with:

- a field `x` that takes a list of at least 3 integers greater than or equal to 0
- a field `y` that takes a string matching a regex for a valid unix username

```python
from typing import Annotated, List

from msgspec import Struct, Meta, json


class Example(Struct):
    x: Annotated[
        List[Annotated[int, Meta(ge=0)]],
        Meta(min_length=3),
    ]
    y: Annotated[
        str,
        Meta(pattern="^[a-z_]([a-z0-9_-]{0,31}|[a-z0-9_-]{0,30}\$)$")
    ]


dec = json.Decoder(Example)
```

Validation happens during decoding, and except for the `pattern` validation has negligible overhead. Users should feel comfortable adding constraints to types as needed without worrying about performance impacts.

Errors are raised for values not matching the specified constraints, just as they are for value not matching the specified types:

```python
>>> dec.decode(b'{"x": [1, 2, 3], "y": "alice"}')  # ok
Example(x=[1, 2, 3], y='alice')

>>> dec.decode(b'{"x": [1, 2], "y": "ben"}')  # not enough items
msgspec.ValidationError: Expected `array` of length >= 3 - at `$.x`

>>> dec.decode(b'{"x": [1, 2, -1], "y": "carol"}')  # items must be >= 0
msgspec.ValidationError: Expected `int` >= 0 - at `$.x[2]`

>>> dec.decode(b'{"x": [1, 2, 3], "y": "dave\'s username"}')  # doesn't match regex
msgspec.ValidationError: Expected `str` matching regex '^[a-z_]([a-z0-9_-]{0,31}|[a-z0-9_-]{0,30}\\$)$' - at `$.y`
```

The `Meta` objects also contain support for json-schema metadata annotations (`title`, `description`, ...), but currently nothing is done with it. Generation of a json-schema spec from a type (#125) will happen in a follow-up PR.

Fixes #154 (kind of). I'm reading the bulk of that issue as "support value constraints", which this adds. The remaining bit of supporting the [annotated-types](https://github.com/annotated-types/annotated-types) library isn't done, and I'm not 100% sure I want to support it yet. I'll open a follow-up issue. to address that, rather than continuing discussion in #154.

Still needs docs, but otherwise this should be done.


jcrist:
cc @adriangb. This is _part_ of what you asked for.

We still don't integrate with the `annotated-types` library itself. Right now I'm +0.1 on making the `Meta` objects produce valid `annotated-types` (using `__iter__` or something), but am not sold on `msgspec` consuming `annotated-types` annotations yet. There's a maintenance and understandability cost on supporting multiple ways of spelling something, and I'm not fully convinced it's worth it here. I'd want to hear a request from other users before spending time on this.

adriangb:
First thing I'll say is that this is amazing. The C part is _way_ over my head. But it's really cool that that you got so much of this working at once.

w.r.t. annotated-types, I'm curious to hear the cons to integrating with it. I can imagine several, but I want to make sure we're on the same page as to which before I try to think of ways to mitigate them.

> There's a maintenance and understandability cost on supporting multiple ways of spelling something, and I'm not fully convinced it's worth it here.

From the developers perspective or the user's? I defiantly understand from a developers perspective, but I think it'd be transparent for a user's perspective.

Also were you thinking of having msgspec understand annotated-types annotations, having it create annotated-types annotations (via `Meta.__iter__`) or both?

jcrist:
> First thing I'll say is that this is amazing. The C part is way over my head. But it's really cool that that you got so much of this working at once.

Thanks! It's been fun to work on. There's some nuance to making these validations fast, but I'm really happy with how this has turned out.

> Also were you thinking of having msgspec understand annotated-types annotations, having it create annotated-types annotations (via `Meta.__iter__`) or both?

I'd be open to making `msgspec` create annotated-types annotations (via `Meta.__iter__`), since this is self contained and easier to do. I'm less open to making it understand annotated-types annotations, since this would complicate the implementation and provide multiple ways to spell the same thing, which may confuse users (more on this below). I'd want to see other users requesting annotated-types integration before spending any time on this myself, since I'm not convinced of the use case.

> I'm curious to hear the cons to integrating with it. I can imagine several, but I want to make sure we're on the same page as to which before I try to think of ways to mitigate them.

There are two main concerns:

### Multiple ways of spelling the same thing.

I want `msgspec` to be easy to use and understand, which means limiting the number of concepts and configurations. We intentionally don't support all the flexibility of `pydantic`, since the number of config options there can be overwhelming and confusing to users. If we support `annotated-types`, then we need to demonstrate their usage in the docs, and explain why a user might choose one or the other.

`annotated-types` also supports some constraints we don't (`Predicate`), and doesn't support some annotations we do (the json-schema metadata things like `title`, `description`, ...). This would force a user to import several different annotations from different places, which doesn't seem as ergonomic to me as a single `Meta` import.

There's also some incompatibilities in meaning that might trip a user up (`max_length` here is inclusive, while it's exclusive in `annotated-types`).

These are surmountable issues given proper documentation, but it'd require some work to present in a clear and concise way.

### Complicated implementation

I want `msgspec` classes to be fast to define (see the [import time benchmark](https://jcristharif.com/msgspec/benchmarks.html#benchmark-structs)). This means limiting the amount of work done at import time, and limiting the amount of sub-packages imported. We also do (almost) all conversion from python type annotations into `TypeNode` objects in C. If we support annotated-types, then this implementation would get a lot more complicated. The easiest way to do it would be to define a utility in python code that converts annotated-types annotations into our own internal `Meta` objects, so the C code only needs to understand `Meta` annotations. This would be possible, but would require some care to ensure it doesn't slow down type construction time.

Neither of these issues is a for-sure "no", but I'm hesitant to spend time on this for now until more users ask for it.

jcrist:
Not sure why coverage is failing, if you look at the diff, all of this patch is covered 🤷. Merging!

PR #181, opened 2022-09-05
Generate JSON Schemas from types

### LINKED ISSUE #125: Support json-schema generation

jcrist:
This adds two functions:

- ``msgspec.json.schema`` for generating a single JSON Schema from a
  given type.
- ``msgspec.json.schema_components`` for generating multiple schemas
  (and a corresponding components mapping) from multiple types.

~Still needs docs and tests.~

Fixes #125.

### DIFF
--- a/docs/source/api.rst
+++ b/docs/source/api.rst
@@ -62,6 +62,16 @@ MessagePack
 .. autofunction:: decode
 
 
+JSON Schema
+-----------
+
+.. currentmodule:: msgspec.json
+
+.. autofunction:: schema
+
+.. autofunction:: schema_components
+
+
 Exceptions
 ----------
 
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -159,6 +159,7 @@ is required.
     structs.rst
     constraints.rst
     extending.rst
+    jsonschema.rst
     perf-tips.rst
     benchmarks.rst
     examples/index.rst
--- a/docs/source/jsonschema.rst
+++ b/docs/source/jsonschema.rst
@@ -1,0 +1,98 @@
+JSON Schema
+===========
+
+``msgspec`` provides a few utilities for generating `JSON Schema`_
+specifications from msgspec-compatible :ref:`types <supported-types>` and
+:doc:`constraints <constraints>`.
+
+- `msgspec.json.schema`: generates a complete JSON Schema for a single type.
+- `msgspec.json.schema_components`: generates JSON schemas for multiple types,
+  along with a corresponding ``components`` mapping. This is mainly useful when
+  generating multiple schemas to include in a larger specification like OpenAPI_.
+
+
+The generated schemas are compatible with `JSON Schema`_ 2020-12 and OpenAPI_
+3.1.
+
+
+Example
+-------
+
+
+.. code-block:: python
+
+    import msgspec
+    from msgspec import Struct, Meta
+    from typing import Annotated, Optional
+
+
+    # A float constrained to values > 0
+    PositiveFloat = Annotated[float, Meta(gt=0)]
+
+
+    class Dimensions(Struct):
+        """Dimensions for a product, all measurements in centimeters"""
+        length: PositiveFloat
+        width: PositiveFloat
+        height: PositiveFloat
+
+
+    class Product(Struct):
+        """A product in a catalog"""
+        id: int
+        name: str
+        price: PositiveFloat
+        tags: set[str] = set()
+        dimensions: Optional[Dimensions] = None
+
+
+    # Generate a schema for a list of products
+    schema = msgspec.json.schema(list[Product])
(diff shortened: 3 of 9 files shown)


jcrist:
A few open questions:

- What types should be extracted into a shared `"$defs"`/`"definitions"`/`"components"` field? Right now I pull out all `Struct`, `TypedDict`, `NamedTuple`, and `Enum` types.
- How should `title` be set if not explicitly set by annotating with a `Meta` object? Right now it's the class name for shared definitions in `$defs`, and the field name (converted to a title, e.g. `"field_one" -> "Field one"`) for fields. All other nodes in the schema have no title.
- How should `description` be set if not explicitly set by annotating with a `Meta` object? Right now it's the `__doc__` for shared definitions in `$defs`, and nothing otherwise.

jcrist:
An example:


```python
import enum
from typing import Union

import msgspec
import rich


class Role(enum.Enum):
    """A role for a team member"""
    manager = "manager"
    lead = "lead"
    developer = "developer"
    other = "other"


class TeamMember(msgspec.Struct):
    """A team member in a team"""
    name: str
    skills: list[str] = []
    email: Union[str, None] = None
    role: Role


class Team(msgspec.Struct):
    """A team in the organization"""
    name: str
    members: list[TeamMember] = []


rich.print(msgspec.json.schema(list[Team]))
```

And the output:

```
{
    'type': 'array',
    'items': {'$ref': '#/$defs/Team'},
    '$defs': {
        'Team': {
            'title': 'Team',
            'description': 'A team in the organization',
            'type': 'object',
            'required': ['name'],
            'properties': {
                'name': {'title': 'Name', 'type': 'string'},
                'members': {
                    'title': 'Members',
                    'default': [],
                    'type': 'array',
                    'items': {'$ref': '#/$defs/TeamMember'}
                }
            }
        },
        'TeamMember': {
            'title': 'TeamMember',
            'description': 'A team member in a team',
            'type': 'object',
            'required': ['name', 'role'],
            'properties': {
                'name': {'title': 'Name', 'type': 'string'},
                'role': {'$ref': '#/$defs/Role'},
                'skills': {
                    'title': 'Skills',
                    'default': [],
                    'type': 'array',
                    'items': {'type': 'string'}
                },
                'email': {
                    'title': 'Email',
                    'default': None,
                    'anyOf': [{'type': 'string'}, {'type': 'null'}]
                }
            }
        },
        'Role': {
            'title': 'Role',
            'description': 'A role for a team member',
            'enum': ['developer', 'lead', 'manager', 'other']
        }
    }
}
```

jcrist:
cc @adriangb for thoughts ^^

adriangb:
I'm a bit confused about `msgspec.json.schema_components`: what is the difference with calling `msgspec.json.schema` multiple times (other than performance)?

Can the name for `$defs` be customized in `schema()`? Or could `schema()` return a `Tuple[SchemaType, Mapping[str, SchemaType]]` or something like that?

jcrist:
> what is the difference with calling msgspec.json.schema multiple times (other than performance)?

- Performance (no repeating work building schemas for components shared across two different calls)
- The shared components are returned separately in `schema_components`, rather than set as the `$defs` key in the output.

In general `schema` is what a user probably wants if they're only generating a schema for a single document. `schema_components` is what they want if they're doing anything more complicated (e.g. generating OpenAPI specs).

> Can the name for `$defs` be customized in `schema()`?

The name can be customized in `schema_components`, but not in `schema`. Why do you want to customize it? Is `$defs` not a good default (it's the recommended value in the json-schema spec)?

> Or could `schema()` return a `Tuple[SchemaType, Mapping[str, SchemaType]]` or something like that?

That's basically the return type of `schema_components`, except that it's not variadic.

---

Are these APIs sufficient for your use cases? Based on my understanding of json-schema and OpenAPI I _think_ everything a user may want to do should be possible with the functions added here, but I confess I'm not a heavy user of either of these tools.

adriangb:
I think these APIs work and are sufficient. I don't see myself using `schema()` but it doesn't hurt either.

jcrist:
Ok, this is in, tests, docs, and all. I've checked with some simple examples that the schemas are handled properly by swagger, but there might still be some incompatibilities with other tools. I'm mostly going off my reading of the JSON Schema and OpenAPI docs. It'd be excellent if you (or others) could try this out and see if the generated schemas are sufficient for your use cases, and play well with other tools.

adriangb:
https://github.com/adriangb/xpresso/pull/106

PR #218, opened 2022-12-01
dataclass support

### LINKED ISSUE #57: Support for `dataclass`es
Given the close similarity between [`msgspec.Struct`]( https://jcristharif.com/msgspec/api.html#msgspec.Struct ) and [`dataclasses.dataclass`]( https://docs.python.org/3/library/dataclasses.html ), am curious to what extent `dataclass`es can be used with `msgspec` and what this entails.

jcrist:
This PR adds encoding/decoding support for dataclasses, mirroring our support for `typing.NamedTuple`/`collections.namedtuple`/`typing.TypedDict`. Dataclasses are encoded to/decoded from mappings containing all attributes assigned to the object (union of `__dict__` and `__slots__`).

I do not intend to add support for additional configuration (e.g. renaming fields using `camelCase`) - if the user wants more customization they should switch to using structs. Structs support all the common features of dataclasses, but are more performant to create/use/encode/decode, and are more customizable. The intent of adding dataclass support is:

- Compatibility with what `orjson` is capable of encoding efficiently. Initial benchmarks show the implementation here encodes measurably faster than the one in `orjson`, especially if `slots=True` is used, see below.
- More general support for common python object types.

Todo:

- [x] JSON encoding support
- [x] MessagePack encoding support
- [x] JSON decoding support
- [x] MessagePack decoding support
- [x] Docs

Fixes #57.

### DIFF
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -119,6 +119,7 @@ Most combinations of the following types are supported (see
 - `typing.Literal`
 - `typing.NamedTuple` / `collections.namedtuple`
 - `typing.TypedDict`
+- `dataclasses.dataclass` types
 - `msgspec.msgpack.Ext`
 - `msgspec.Raw`
 - `enum.Enum` subclasses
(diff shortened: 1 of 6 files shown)


jcrist:
A simple encoding benchmark:

```python
In [1]: import msgspec, orjson

In [2]: from dataclasses import dataclass

In [3]: enc = msgspec.json.Encoder()

In [4]: @dataclass
   ...: class NoSlots:
   ...:     field_one: int
   ...:     field_two: int
   ...: 

In [5]: @dataclass(slots=True)
   ...: class Slots:
   ...:     field_one: int
   ...:     field_two: int
   ...: 

In [6]: no_slots = [NoSlots(i - 1, i + 1) for i in range(10000)]

In [7]: with_slots = [Slots(i - 1, i + 1) for i in range(10000)]

In [8]: %timeit enc.encode(no_slots)  # msgspec, no slots
561 µs ± 2.04 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)

In [9]: %timeit orjson.dumps(no_slots)  # orjson, no slots
834 µs ± 1.69 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)

In [10]: %timeit enc.encode(with_slots)  # msgspec, with slots
779 µs ± 20 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)

In [11]: %timeit orjson.dumps(with_slots)  # orjson, with slots
3.71 ms ± 90.1 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [12]: class Struct(msgspec.Struct):
    ...:     field_one: int
    ...:     field_two: int
    ...: 

In [13]: structs = [Struct(i - 1, i + 1) for i in range(10000)]

In [14]: %timeit enc.encode(structs)  # msgspec structs
356 µs ± 307 ns per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
```

For these type definitions on my machine:

- msgspec encodes dataclasses to JSON 1.5x faster than orjson
- msgspec encodes dataclasses with `slots=True` to JSON 5x faster than orjson
- dataclasses with `slots=True` are slower to encode than `slots=False`. This has to do with object layouts and what information is efficiently accessible on the type definition.
- msgspec encodes `Struct` types 1.5x faster than dataclasses

PR #228, opened 2022-12-06
Fix typo in docs for runtime definition of Struct

physinet:
Noticed this small typo in the Struct documentation. The example appears to be valid now.


jcrist:
Thanks!

PR #258, opened 2023-01-08
Add `msgspec.to_builtins` method

jcrist:
This adds a new `msgspec.to_builtins` function, for transforming objects composed of all the types `msgspec` supports to objects composed only of types standard python serializers support (`list`, `tuple`, `int`, `str`, ...).

While this method *may* be useful directly to users, the primary intended use is for wrapping it with additional serialization APIs within `msgspec` itself (for example, a new `msgspec.yaml` module). While JSON may require high-performance and a builtin encoder, config-files like YAML, TOML, ... usually aren't perforamnce sensitive, so the extra copies here shouldn't matter.

Once this is in, we'll follow up with a `msgspec.from_builtins` for handling the "decoding" side of things.


jcrist:
Example of usage:

```python
In [1]: import msgspec                         

In [2]: class User(msgspec.Struct):
   ...:     name: str                          
   ...:     groups: set[str] = set()
   ...:     email: str | None = None
   ...:                                        

In [3]: alice = User("alice", groups={"admin", "engineering"}, email="alice@company.com")

In [4]: alice                                  
Out[4]: User(name='alice', groups={'engineering', 'admin'}, email='alice@company.com')

In [5]: msg = msgspec.to_builtins(alice)  # convert to simple types

In [6]: msg                                    
Out[6]:                                        
{'name': 'alice',                              
 'groups': ['engineering', 'admin'],
 'email': 'alice@company.com'}

In [7]: import yaml                            

In [8]: print(yaml.safe_dump(msg, sort_keys=False))  # can pass to downstream libraries
name: alice                                    
groups:                                        
- engineering                                  
- admin                                        
email: alice@company.com
```

Like I said above, ideally most users won't ever need to directly access this API - we should instead use it to wrap common downstream libraries (`yaml`, `toml`, ...) in high-level APIs mirroring the existing `msgspec.json`/`msgspec.msgpack` modules.

PR #262, opened 2023-01-13
Add `msgspec.replace` function

### LINKED ISSUE #261: Feature request: evolving a frozen struct (with sample code)
The popular attrs library has a utility function [`evolve`](https://www.attrs.org/en/stable/api.html#attrs.evolve). It is used to make a copy of a frozen object with certain changes applied.

Example:

```python
class Point(msgspec.Struct, frozen=True):
    x: int
    y: int

p = Point(x=150, y=150)
p2 = evolve(p, y=250)
assert p2.x == 150
assert p2.y == 250
```

I created an implementation for my own project, but you may wish to add additional error handling, or perhaps implement it in the C core. I am also not sure which namespace to which it ought to be added.

```python
def evolve(obj: msgspec.Struct, **changes):
    cls = obj.__class__
    for field_name in cls.__struct_fields__:
        if field_name not in changes:
            changes[field_name] = getattr(obj, field_name)
    return cls(**changes)
```

jcrist:
This adds a new `msgspec.replace` function, mirroring the functionality of `dataclasses.replace`/`NamedTuple._replace`/`attrs.evolve`. It takes an existing `Struct` instance, and returns a copy with any fields passed as `**changes` replacing the original values.

Note that:
- This works on any struct (including frozen structs), since this returns a new instance rather than mutating the old instance.
- This doesn't do a deepcopy of any of the field values. If the input struct had a list on a field, the output struct will have the same instance of that list (unless that field was passed in `**changes`).

Fixes #261.

### DIFF
--- a/docs/source/api.rst
+++ b/docs/source/api.rst
@@ -11,6 +11,8 @@ Struct
 
 .. autofunction:: defstruct
 
+.. autofunction:: replace
+
 
 Meta
 ----
--- a/msgspec/__init__.py
+++ b/msgspec/__init__.py
@@ -1,5 +1,6 @@
 from ._core import (
     Struct,
+    replace,
     defstruct,
     Raw,
     Meta,
--- a/msgspec/__init__.pyi
+++ b/msgspec/__init__.pyi
@@ -61,6 +61,9 @@ class Struct(metaclass=__StructMeta):
     ) -> None: ...
     def __rich_repr__(self) -> Iterable[Tuple[str, Any]]: ...
 
+S = TypeVar("S", bound=Struct, covariant=True)
+
+def replace(struct: S, /, **changes: Any) -> S: ...
 def defstruct(
     name: str,
     fields: Iterable[Union[str, Tuple[str, Type], Tuple[str, Type, Any]]],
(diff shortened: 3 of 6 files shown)


jcrist:
A quick benchmark against the pure python version provided in the original issue:

```python
In [1]: %load test.py

In [2]: # %load test.py
   ...: from msgspec import defstruct, replace
   ...: 
   ...: 
   ...: Test = defstruct("Test", [f"field_{i}" for i in range(10)])
   ...: 
   ...: 
   ...: def evolve(obj, **changes):
   ...:     cls = obj.__class__
   ...:     for field_name in cls.__struct_fields__:
   ...:         if field_name not in changes:
   ...:             changes[field_name] = getattr(obj, field_name)
   ...:     return cls(**changes)
   ...: 
   ...: 
   ...: orig = Test(*range(10))
   ...: changes = {"field_1": 100, "field_6": 250}
   ...: 

In [3]: %timeit evolve(orig, **changes)
1.33 µs ± 4.42 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)

In [4]: %timeit replace(orig, **changes)
246 ns ± 0.883 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)

```

PR #274, opened 2023-01-23
Add support for `default_factory` on `Struct` types

### LINKED ISSUE #259: Support default factories for Struct types
Hi there,

thank you for your great library! We are investigating a switch from pydantic to msgspec. Our main usecase is messaging (commands, events) through a message broker. Is there a way to supply factories to fields? E.g. every user-generated message in our system gets a timestamp (tz-aware, which makes the init a little more annoying in python).

Did I miss the documentation or would you consider adding support for factory functions?

Thanks!

jcrist:
This PR does a few things:

- It adds a new `msgspec.field` function, which returns an *opaque* object used for configuring fields. So far this only exposes `default` and `default_factory`, both of which match the `dataclasses` kwargs of the same name.
- It adds support for init-time generated default values in `msgspec.Struct` types, through use of the new `default_factory` setting (fixes #259).

Example:

```python
import msgspec
import uuid

class Example(msgspec.Struct):
    id: uuid.UUID = msgspec.field(default_factory=uuid.uuid4)
```

### Breaking Change: mutable default values in structs are now handled differently

This also changes how mutable default values are handled. Previously msgspec's semantics were that a default value for a field would be deepcopied on `__init__` if that field wasn't provided (the implementation would avoid a deepcopy for most common types, but semantically this was the same). This proved problematic in the face of custom types, led to some unnecessary slowdowns, and didn't match the behavior of other similar libraries like `dataclasses` or `attrs`. Since we now can support arbitrarily complex default values through the `default_factory` function, we drop the old `deepcopy` behavior.

A default value on a `Struct` now has the following rules:
- Common mutable containers (list, dict, set, bytearray) will error if used as the default value *if* they aren't empty. The error points users to use a `default_factory` instead.
- Empty common mutable containers (`[]`, `{}`, `set()`, `bytearray()`) are accepted as default values. These are treated as syntactic sugar, and are automatically converted to the equivalent `default_factory` notation (`[]` -> `field(default_factory=list)`).
- Mutable struct types will error if used as default values. The error points users to use a `default_factory` instead.
- Frozen struct types are accepted as a default value, and are used directly without deepcopying. 
- All other values are used directly without deepcopying

Since most mutable default values are empty collections, I don't expect this breaking change to affect most users. To reiterate, the following type definition is still valid:

```python
class Example(Struct):
    x: list[int] = []
    y: set[int] = set()
    z: dict[str, int] = {}
```

while this one will error, and should use a `default_factory` instead:

```python
class Bad(Struct):
    x: list[int] = [1, 2, 3]

# should be
class Good(Struct):
    x: list[int] = field(default_factory=lambda: [1, 2, 3])
```

Todo:
- [x] Tests
- [x] Docs
- [x] Ensure all examples are updated to the newer syntax

### DIFF
--- a/docs/source/api.rst
+++ b/docs/source/api.rst
@@ -9,6 +9,8 @@ Struct
 .. autoclass:: Struct
     :members:
 
+.. autofunction:: field
+
 .. autofunction:: defstruct
 
 .. autofunction:: replace
--- a/docs/source/structs.rst
+++ b/docs/source/structs.rst
@@ -30,8 +30,7 @@ required field and two optional fields.
   `None` if no value is provided.
 
 - ``groups`` is an *optional* field expecting a `set` of `str`. If no value is
-  provided, it defaults to the empty set (note that mutable default values are
-  deep-copied before use).
+  provided, it defaults to the empty set.
 
 Struct types automatically generate a few methods based on the provided type
 annotations:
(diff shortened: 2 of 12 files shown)


jcrist:
A quick benchmark of `__init__` costs after this change, comparing msgspec to a few other dataclass-like libraries: https://gist.github.com/jcrist/9bfe44f60533225d5f8383791f2fe734. In short - msgspec is _fast_. Allocating a new struct *even in the presence of default values* now is as fast as allocating a tuple.

PR #282, opened 2023-01-28
Add new `msgspec.structs` module

### LINKED ISSUE #275: expose annotation on field value
Hi,

So I am working on integrating Struct inside our internals, but there is a problem - I need access to the raw annotation on struct fields. Currently all I can get is the tree like structure of parsed fields, and if I need an annotation I need to do rather funky manipulation. 

So, can you please expose the raw annotation value on the struct field as part of the field data?

```python
class Field(msgspec.Struct):
    """A record describing a field in an object-like type.

    Parameters
    ----------
    name: str
        The field name as seen by Python code (e.g. ``field_one``).
    encode_name: str
        The name used when encoding/decoding the field. This may differ if
        the field is renamed (e.g. ``fieldOne``).
    type: Type
        The field type.
    required: bool, optional
        Whether the field is required. Note that if `required` is False doesn't
        necessarily mean that `default` or `default_factory` will be set -
        optional fields may exist with no default value.
    default: Any, optional
        A default value for the field. Will be `UNSET` if no default value is set.
    default_factory: Any, optional
        A callable that creates a default value for the field. Will be
        `UNSET` if no ``default_factory`` is set.
    """
    
    annotation: Any # <--- can also be typed with `type` if preferable
    name: str
    encode_name: str
    type: Type
    required: bool = True
    default: Any = UNSET
    default_factory: Any = UNSET
```

jcrist:
This adds a new `msgspec.structs` module. The purpose of this module is to include struct-related utilities that aren't worth surfacing to the top-level namespace, but don't really belong anywhere else either.

So far this just includes a new `msgspec.structs.info` function (and return types) for inspecting information about a struct type. This is a shallow-version of the full-featured `msgspec.inspect.type_info` function:

- It only works for struct types
- It doesn't recurse into the field types

Most use cases where you're inspecting msgspec types you probably want to use `msgspec.inspect.type_info` instead.

(Maybe) fixes #275.

### DIFF
--- a/docs/source/api.rst
+++ b/docs/source/api.rst
@@ -155,6 +155,16 @@ Inspect
 .. autodata:: UNSET
 
 
+Struct Utilities
+----------------
+
+.. currentmodule:: msgspec.structs
+
+.. autoclass:: StructInfo
+.. autoclass:: StructField
+.. autofunction:: info
+
+
 Exceptions
 ----------
 
--- a/msgspec/__init__.py
+++ b/msgspec/__init__.py
@@ -18,6 +18,7 @@ from . import json
 from . import yaml
 from . import toml
 from . import inspect
+from . import structs
 
 
 def field(*, default=UNSET, default_factory=UNSET):
--- a/msgspec/__init__.pyi
+++ b/msgspec/__init__.pyi
@@ -182,5 +182,6 @@ from . import json
 from . import yaml
 from . import toml
 from . import inspect
+from . import structs
 
 __version__: str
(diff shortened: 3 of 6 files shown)


Goldziher:
Hi, so Im testing this now. I think I can simply use the work-around I use, which is to access the `__annotations__` field on the class and use the values from there. So thanks a lot for your work! But I don't think this one is really crucial. Up to you to decide of course.

jcrist:
With #328 exposing all struct class-level configuration, I'm going to close and rework this to something like:

```python
def fields(struct_instance_or_class) -> tuple[FieldInfo]:
    ...

class FieldInfo:
    name: str
    encode_name: str
    type: Any
    default: Any = None
```

Issue #316, opened 2023-02-13
Integrate with the attrs ecosystem


Tinche:
### Description

Hello! I think this is a cool library, and I think you should consider integrating with the attrs ecosystem.

A small disclaimer first: I'm a major contributor to attrs (behind @hynek), the author of cattrs, and at this point I'm slowly getting comfortable starting to say I'm one of the mypy attrs plugin maintainers.

I haven't had the chance to use msgspec for real yet. You've done some very impressive things efficiency-wise, but I feel there's a lot of overlap between some work here and some work in attrs/cattrs. So now I'm thinking of ways to integrate.

Strategically speaking, I think msgspec is cool but limited in a number of ways. From what I understand, you don't support generics and PyPy (maybe other constraints I'm not aware of, not sure if you support dict classes?), which would limit my personal use of the library. Also I'm attached to the way cattrs does customization of de/serialization.

However, here's the thing: if we integrate the way I'm thinking, there would be no need for you to actually do these things (you could if you wanted to, or you could focus on other aspects). We could integrate msgspec classes (er, structs) into the attrs system and msgspec de/serialization into cattrs.

A simple idea would be having msgspec as an attrs backend that would be opt-in - so if your class isn't generic, and you don't use PyPy, you'd choose to use msgspec, otherwise you'd use the default attrs backend and have an attrs class. The resulting class would contain `__attrs_attrs__`, and so would essentially be an attrs class. 

What I mean by the attrs ecosystem: for example, Rich supports attrs classes, although I know you support Rich in msgspec. Another example is an OpenAPI generator I'm working on for attrs classes, which would just work for msgspec classes too. A third example is work I'm doing for better typing of `attrs.fields` for use in ORMs/ODMs; this would also just work.

The first and most important question, however, is if you are even open to doing this? If yes, we can bounce ideas.

jcrist:
Hi! Thanks for opening this. A few thoughts inline:

> From what I understand, you don't support generics and PyPy (maybe other constraints I'm not aware of, not sure if you support dict classes?), which would limit my personal use of the library.

Generic structs (#193) are currently unsupported, but are at the top of my list to implement. This is a solvable problem, and is something I'd expect to be resolved in the next 3 months (hopefully sooner).

By "dict classes" I assume you mean Structs with a `__dict__` defined? Those are unplanned and unlikely to happen without a strong use case. Do you have a use case for those over relying on `__slots__`?

[...]

PR #323, opened 2023-02-16
Support encoding and decoding attrs types

### LINKED ISSUE #51: Attrs Support?
Any consideration being given to supporting Attrs (https://www.attrs.org/en/stable/index.html) as a mechanism for Structs? It would be nice to leverage already existing attrs classes instead of refactoring to completely supporting the Structs. There is a workaround to just leveraging the `attr.as_dict` method to serialize into msgpack via `encode`; then deserialize into a `from_dict` classmethod; but that precludes taking advantage of the serialization/deserialization speedup documented for using structs over dicts.

jcrist:
This adds support for encoding/decoding attrs types. It's mostly built off of our existing `dataclasses` functionality, and has the same restrictions:

- When encoding, any *object* attribute without a leading underscore is encoded using the attribute name, even if it's not marked as an attribute as part of the class definition. This is for performance reasons, accessing `__attrs_attrs__` at encoding time comes at a perf cost.
- When decoding, the `__init__` method on the class is not called. This is for both efficiency and correctness reasons. As such, classes that define their own `__init__` or `__attrs_init__` will not have this called during the decoding process. `__attrs_pre_init__` and `__attrs_post_init__` methods are called at the proper times though.
- Currently default factories with `takes_self=True` are not supported. We could support this, but it was more work than I wanted to do right now.

Note that right now neither attrs' validators or convertors are run on decode. This is fixable, but would require a lot more work since this would diverge from the existing `dataclasses` implementation.

All in all this was pretty straightforward to get working. It's nice that `dataclasses` and `attrs` implementations haven't diverged too much here.

This also fixes a bug in the existing dataclass decoder that prevented decoding dataclasses with `frozen=True` configured.

Fixes #51.

Related to (but doesn't resolve) #316.

### DIFF
--- a/docs/source/supported-types.rst
+++ b/docs/source/supported-types.rst
@@ -56,6 +56,10 @@ Most combinations of the following types are supported (with a few restrictions)
 - `collections.abc.Mapping` / `typing.Mapping`
 - `collections.abc.MutableMapping` / `typing.MutableMapping`
 
+**Third-Party Libraries**
+
+- attrs_ types
+
 Additional types may be supported through :doc:`extensions <extending>`.
 
 Note that except where explicitly stated, subclasses of these types are not
@@ -558,6 +562,47 @@ additional features.
       File "<stdin>", line 1, in <module>
     msgspec.ValidationError: Expected `int`, got `str` - at `$.age`
 
+``attrs``
+---------
+
+attrs_ types map to objects/maps in all protocols.
+
+During encoding, all attributes without a leading underscore (``"_"``) are
+encoded.
+
+During decoding, any extra fields are ignored. An error is raised if a field's
+type doesn't match or if any required fields are missing.
+
+If the ``__attrs_pre_init__`` or ``__attrs_post_init__`` methods are defined on
+the class, they are called as part of the decoding process.
+
+When possible we recommend using `msgspec.Struct` instead of attrs_ types for
+specifying schemas - :doc:`structs` are faster, more ergonomic, and support
+additional features.
+
+.. code-block:: python
+
+    >>> from attrs import define
+
+    >>> @define
+    ... class Person:
+    ...     name: str
+    ...     age: int
+
+    >>> carol = Person(name="carol", age=32)
+
+    >>> msg = msgspec.json.encode(carol)
+
+    >>> msgspec.json.decode(msg, type=Person)
+    Person(name='carol', age=32)
+
+    >>> wrong_type = b'{"name": "doug", "age": "thirty"}'
+
+    >>> msgspec.json.decode(wrong_type, type=Person)
+    Traceback (most recent call last):
+      File "<stdin>", line 1, in <module>
+    msgspec.ValidationError: Expected `int`, got `str` - at `$.age`
+
 ``Struct``
 ----------
 
@@ -801,8 +846,7 @@ Union restrictions are as follows:
   issue <https://github.com/jcrist/msgspec/issues>`__.
 
 - Unions may contain at most one type that encodes to an object (`dict`,
-  `typing.TypedDict`, `dataclasses.dataclass`, `Struct` with
-  ``array_like=False``)
+  `typing.TypedDict`, dataclasses_, attrs_, `Struct` with ``array_like=False``)
 
 - Unions may contain at most one type that encodes to an array (`list`,
   `tuple`, `set`, `frozenset`, `typing.NamedTuple`, `Struct` with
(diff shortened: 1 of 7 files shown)


jcrist:
A quick example:

```python
In [1]: import attrs                           

In [2]: import msgspec                         

In [3]: @attrs.define                          
   ...: class User:                            
   ...:     name: str                          
   ...:     groups: list[str] = []
   ...:     email: str | None = None
   ...:                                        

In [4]: alice = User("alice", groups=["admin", "engineering"])

In [5]: msg = msgspec.json.encode(alice)

In [6]: msg                                    
Out[6]: b'{"email":null,"groups":["admin","engineering"],"name":"alice"}'

In [7]: msgspec.json.decode(msg, type=User)
Out[7]: User(name='alice', groups=['admin', 'engineering'], email=None)

In [8]: msgspec.json.decode(b'{"name": 123}', type=User)
ValidationError                           Traceback (most recent call last)
Cell In [8], line 1                            
----> 1 msgspec.json.decode(b'{"name": 123}', type=User)

ValidationError: Expected `str`, got `int` - at `$.name`
```

ltworf:
> When decoding, the __init__ method on the class is not called.

Does this still work if the field had defined a converter?

jcrist:
> Does this still work if the field had defined a converter?

Converters and validators aren't currently supported with the above model. I'm not an attrs user, so I'm not sure what the expected behavior is here - I was hoping to get some feedback on this from @tinche on what's expected before moving forward with this.

Tinche:
Looks pretty cool to me! I personally don't use validators, and converters sparingly so this is a good first step.

Why do you skip attributes with a leading underscore though? We use MongoDB a lot (through our own ODM based on attrs ;) and Mongo documents require an `_id` field. So all my database models have that, and I'd expect it to be in the payload.

jcrist:
> Looks pretty cool to me! I personally don't use validators, and converters sparingly so this is a good first step.

Thanks for the review! Just to clarify - do you think this is good to merge as is, and would be useful for you (or others) without immediate further changes?

> Why do you skip attributes with a leading underscore though? 

For performance reasons, we don't touch `__attrs_attrs__`/`__dataclass_fields__` at all during encode time - all attributes without a leading underscore found in either `__dict__` or `__slots__` on an object are encoded. FWIW this is the same behavior as `orjson` has for dataclasses, and seems to work well enough in practice. We drop the leading underscore attributes since sometimes users stash private state on an object with a leading-underscore attribute, but in my experience it's rare to have a true dataclass field have a leading underscore (I don't use attrs, so no experience there).

> Mongo documents require an `_id` field. So all my database models have that, and I'd expect it to be in the payload.

For struct types, we support [renaming fields](https://jcristharif.com/msgspec/structs.html#renaming-field-names) for this use case. The assumption is that user code probably wants to work with python-friendly attribute names (e.g. `id`), and the name used for encoding is metadata on the field. This isn't exposed at the `msgspec.field` level yet (`msgspec.field` itself is pretty new), so currently you'd do this as:

```python
In [1]: import msgspec

In [2]: class Person(msgspec.Struct, rename={"id": "_id"}):
   ...:     id: int
   ...:     name: str
   ...: 

In [3]: alice = Person(1, "alice")

In [4]: alice.id  # attribute is original field name
Out[4]: 1

In [5]: alice
Out[5]: Person(id=1, name='alice')

In [6]: msgspec.json.encode(alice)  # encoding uses renamed name
Out[6]: b'{"_id":1,"name":"alice"}'

In [7]: msgspec.json.decode(_, type=Person)  # as does decoding
Out[7]: Person(id=1, name='alice')
```

Currently this feature is only exposed for `Struct` types, since dataclasses/attrs don't have an equivalent existing config option. We _could_ expose this through the arbitrary `metadata` option if needed, but I'd rather not until someone asks for it. My recommendation for users needing this feature is to switch to `msgspec.Struct` types. They're more performant, and have more config options.

---

In summary:

- The current implementation of what's *encoded* on an `attrs`/`dataclass` type is done for performance/ease of implementation. It doesn't look at `__attrs_attrs__`/`__dataclass_fields__` at all. We _could_ change this to respect the defined list of fields, but it would complicate things. Happy to do it if asked, but not until a user (or you) asks for it.
- msgspec also supports configuring alternative names for encoding on `msgspec.Struct` types only. We could expose this for dataclasses/attrs if needed, but for now I'd recommend users to switch to using a `msgspec.Struct` in those cases. It's not clear to me how we'd expose this config option for dataclasses/attrs anyway.

PR #330, opened 2023-02-24
Add `msgspec.structs.fields` utility

jcrist:
This mirrors the functionality in `dataclasses.fields`, returning a tuple of `FieldInfo` objects providing information about the individual struct fields.

Supersedes #282.


jcrist:
When coupled with #328, this now exposes all information about a `Struct` type (or instance) in a public and documented way.

PR #333, opened 2023-02-28
Benchmark typedload

ltworf:
Following the inclusion of a comparison to msgspec in typedload's documentation: https://github.com/ltworf/typedload/pull/390

Here is the inclusion of a comparison with typedload in msgspec's documentation.

The validation benchmark is very simple, it should at least hit some unions to become a bit more meaningful, in my opinion. Anyway that's something for the future.


ltworf:
> The purpose of these benchmarks is to compare msgspec against other libraries that are either more popular or more performant.

Well typedload has more monthly downloads than msgspec, and more features. So a comparison does seem worthwhile :)

But from our brief reddit chat, I had understood your type :)

Best

PR #350, opened 2023-03-23
Use `msgspec.UNSET` for tracking unset fields

### LINKED ISSUE #344: Losslessly represent omitted fields
# Description

I sometimes have JSON objects where the presence or absence of a field is semantically distinct from whether that field's value is `null`.

In keeping with `msgspec`'s principles of strictness and correctness, I'd like for the library to be able to losslessly and reversibly encode and decode these objects.

# Example use case

Suppose you have a database of software bugs. Each bug has a `title`, and, optionally, a single `assignee`.

You want to use JSON to represent search filters.

This filter would mean "find all issues titled 'App crashes' that are assigned to `@JohnDoe`":

```json
{
    "title": "App crashes"
    "assignee": "@JohnDoe"
}
```

And this would mean "find all issues titled 'App crashes' that **don't have an assignee**":

```json
{
    "title": "App crashes"
    "assignee": null
}
```

And this would mean "find all issues titled 'App crashes' **regardless of assignee**":

```json
{
    "title": "App crashes"
}
```

In JSON Schema, I would represent this like this:

```json
{
    "$ref": "#/$defs/SearchFilter",
    "$defs": {
        "SearchFilter": {
            "title": "SearchFilter",
            "type": "object",
            "properties": {
                "title": {
                    "type": "string"
                },
                "assignee": {
                    "anyOf": [
                        { "type": "string" },
                        { "type": "null" }
                    ]
                }
            }
            "required": []
        }
    }
}
```

# Proposed API design

Using `msgspec`, I would want to implement the example above something like this:

```python
class SearchFilter(msgspec.Struct):
    # OMITTED and OMITTED_TYPE do not exist today.
    title: str | msgspec.OMITTED_TYPE = msgspec.OMITTED
    assignee: str | None | msgspec.OMITTED_TYPE = msgspec.OMITTED


def filter(all_bugs: list[Bug], filter: SearchFilter) -> list[Bug]:
    def matches_filter(bug: Bug) -> bool:
        title_matches = filter.title == msgspec.OMITTED or filter.title == bug.title
        assignee_matches = filter.assignee == msgspec.OMITTED or filter.assignee == bug.assignee
        return title_matches and assignee_matches
    return [bug for bug in all_bugs if matches_filter(bug)]
```

* Two new symbols are added to `msgspec`: `OMITTED` and `OMITTED_TYPE`.
  * `OMITTED` is a unique sentinel value, distinct from `None`.
  * `OMITTED_TYPE` is the type of `OMITTED`, probably an alias of `typing.Literal[OMITTED]`.
  * Perhaps values of `OMITTED` evaluate to falsey, like `None` does.
* When `encode()` encounters a value of `OMITTED`, it omits the entire key-value pair from the output.
* The `decode()` behavior is unchanged. When it decodes a message that has a certain field missing, it returns that field's default value. In this case, that value can happen to be the sentinel value `OMITTED`.

# Prior art in other libraries

## [PEP 655](https://peps.python.org/pep-0655/)

[PEP 655](https://peps.python.org/pep-0655/) discusses the same problem for `TypedDict`s.

They solve it in a different way. Instead of having a special sentinel value like I'm proposing, they introduce the wrapping types `typing.Required[T]` and `typing.NotRequired[T]`. (These appear to be pass-throughs to `T` at run time. I guess you're supposed to use the `in` operator to gate any potentially unsafe accesses, but [mypy doesn't enforce this today](https://github.com/python/mypy/issues/12094).)

They also explicitly reject the name "omittable."

It feels to me like a lot of their arguments don't make sense when applied to `msgspec.Struct`s, as opposed to `dict`s. But I haven't spent the time to try these things out, or to read the mailing lists and dig into their thinking. Maybe they're right.

## Pydantic

Pydantic has historically badly conflated these concepts. [Planned changes for v2.0](https://docs.pydantic.dev/blog/pydantic-v2/#required-vs-nullable-cleanup) look like they'll bring Pydantic to parity with `msgspec` as it exists today, but they won't address the problem I'm describing here.

jcrist:
`msgspec.UNSET` is a singleton object used to indicate that a field has no set value. This is useful for cases where you need to differentiate between a message where a field is missing and a message where the field is explicitly `None`.

```python
>>> from msgspec import Struct, UnsetType, UNSET, json

>>> class Example(Struct):
...     x: int
...     y: int | None | UnsetType = UNSET  # a field, defaulting to UNSET
```

During encoding, any field containing `UNSET` is omitted from the message.

```python
>>> json.encode(Example(1))  # y is UNSET
b'{"x":1}'

>>> json.encode(Example(1, None))  # y is None
b'{"x":1,"y":null}'

>>> json.encode(Example(1, 2))  # y is 2
b'{"x":1,"y":2}'
```

During decoding, if a field isn't explicitly set in the message, the default value of `UNSET` will be set instead. This lets downstream consumers determine whether a field was left unset, or explicitly set to `None`

```python
>>> json.decode(b'{"x": 1}', type=Example)  # y defaults to UNSET
Example(x=1, y=UNSET)

>>> json.decode(b'{"x": 1, "y": null}', type=Example)  # y is None
Example(x=1, y=None)

>>> json.decode(b'{"x": 1, "y": 2}', type=Example)  # y is 2
Example(x=1, y=2)
```

`UNSET` fields are supported for `msgspec.Struct`, `dataclasses`, and `attrs` types. It is an error to use `msgspec.UNSET` or `msgspec.UnsetType` anywhere other than a field for one of these types.

For ease of use, `UNSET` is falsey and a singleton - usage in predicates should be fairly similar to how Python users already check for `None`:

```python
def on_example(ex: Example):
    if ex.y is UNSET:
        # Handle y being unset
        ...
    elif ex.y is None:
        # Handle y being explicitly None
        ...
    else:
        # Handle y with actual values
        ...
```

Note that this repurposes the existing `msgspec.UNSET` singleton, and is thus a **breaking change**. Anyone who was previously making use of `msgspec.UNSET` should move to using `msgspec.NODEFAULT` instead. My hope is this breakage affects no users, as the previous `msgspec.UNSET` was fairly new, and was a pretty internal API.

Fixes #344.


msgspec/__init__.pyi
```diff
@@ -19,10 +19,15 @@ from . import inspect, json, msgpack, structs, toml, yaml
 
 T = TypeVar("T")
 
-class _Unset:
+class UnsetType:
     pass
 
-UNSET = _Unset()
+UNSET = UnsetType()
```
SyntaxColoring, line 25: If `UNSET` is implemented as an object of an empty `UnsetType` class, static type checkers will unfortunately have a hard time doing type narrowing on it. I'd consider instead making `UNSET` the single value of an enum, so `UnsetType` can be `Literal[UNSET]` and type narrowing will work.

[Example of the problem and how the enum trick fixes it.](https://gist.github.com/SyntaxColoring/d940e6bc5cd37e499779ae8d527c9da1)

This is [GVR's suggested workaround](https://github.com/python/typing/issues/689#issuecomment-561425237). Hopefully someday [PEP 661](https://peps.python.org/pep-0661/) will give us standard sentinel objects, and we'll be able to get this to work without enum trickery. :)

docs/source/supported-types.rst
```diff
@@ -687,6 +688,54 @@ Type checking also still applies.
...
+    >>> class Example(Struct):
+    ...     x: int
+    ...     y: int | None | UnsetType = UNSET  # a field, defaulting to UNSET
+
+During encoding, any field containing ``UNSET`` is omitted from the message.
+
+.. code-block:: python
+
+    >>> json.encode(Example(1))  # y is UNSET
+    b'{"x":1}'
```
SyntaxColoring, line 712: Would this additional example be helpful?

```suggestion
    >>> json.encode(Example(1))  # y is UNSET
    b'{"x":1}'
    
    >>> json.encode(Example(1, UNSET))  # y is UNSET
    b'{"x":1}'
```

SyntaxColoring (approved):
Docs make sense to me! Thanks. A couple of suggestions.

docs/source/supported-types.rst
jcrist: Thanks, added!

PR #385, opened 2023-04-21
Add generic args to typing.Type

mikeshardmind:
Allows this to work better with strict use of pyright, see
  https://github.com/microsoft/pyright/issues/4988

mypy seems to accept either as equivalent, whereas pyright does not.

### DIFF
--- a/msgspec/json.pyi
+++ b/msgspec/json.pyi
@@ -15,7 +15,7 @@ from typing import (
 T = TypeVar("T")
 
 enc_hook_sig = Optional[Callable[[Any], Any]]
-dec_hook_sig = Optional[Callable[[Type[Any], Any], Any]]
+dec_hook_sig = Optional[Callable[[type, Any], Any]]
 
 class Encoder:
     enc_hook: enc_hook_sig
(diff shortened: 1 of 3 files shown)


mikeshardmind:
mypy did not seem to like that particular change in CI, I've gone with the other option of using `type` here, but I'm switching this to a draft until I figure out if there are other discrepancies this causes.

jcrist:
Thanks for the PR! The failure you ran into before was an unrelated spurious `mypy` issue that just started appearing (that I can't reproduce locally). I fixed the failures and added a few additional `Type -> type` changes to finish the PR up in #387. Normally I would push directly to the PR with small fixups like this, but for some reason I wasn't able to here.

Anyway, thanks again for the PR. I hope to cut a release in the next week or two - if you want to use the fix before then you can [install from github](https://jcristharif.com/msgspec/install.html#installing-from-github).

mikeshardmind:
Thanks for the follow-up, I didn't notice you had other commits noting some unrelated mypy issues before coming back to look at it more this morning. I appreciate it.

PR #397, opened 2023-05-01
Dedent docstrings

trendels:
Strip leading whitespace from docstrings using [`textwrap.dedent()`](https://docs.python.org/3.11/library/textwrap.html?t#textwrap.dedent). This fixes markdown rendering of multiline docstrings.

Without this, the leading whitespace before each line causes markdown renderers to interpret the entire docstring as a code block, and inline formatting does not work.

### DIFF
--- a/msgspec/_json_schema.py
+++ b/msgspec/_json_schema.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import re
+import textwrap
 from collections.abc import Iterable
 from typing import Any
 
@@ -142,6 +143,7 @@ def _get_doc(t: mi.Type) -> str:
     doc = getattr(cls, "__doc__", "")
     if not doc:
         return ""
+    doc = textwrap.dedent(doc)
     if isinstance(t, mi.EnumType):
         if doc == "An enumeration.":
             return ""
--- a/msgspec/_json_schema.py
+++ b/msgspec/_json_schema.py
@@ -143,7 +143,7 @@ def _get_doc(t: mi.Type) -> str:
     doc = getattr(cls, "__doc__", "")
     if not doc:
         return ""
-    doc = textwrap.dedent(doc)
+    doc = textwrap.dedent(doc).strip("\r\n")
     if isinstance(t, mi.EnumType):
         if doc == "An enumeration.":
             return ""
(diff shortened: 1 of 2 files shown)


msgspec/_json_schema.py
```diff
@@ -142,6 +143,7 @@ def _get_doc(t: mi.Type) -> str:
     doc = getattr(cls, "__doc__", "")
     if not doc:
         return ""
+    doc = textwrap.dedent(doc)
```
jcrist, line 146: Should we also call `.strip()` here to strip leading/trailing whitespace?

```suggestion
    doc = textwrap.dedent(doc).strip()
```

I _think_ this makes sense. If so, you'll also want to update the test below to match.

tests/test_schema.py
```diff
@@ -1159,3 +1159,29 @@ class Ex2(msgspec.Struct):
...
+        last line.
+        """
+
+        pass
+
+    assert msgspec.json.schema(Example) == {
+        "$ref": "#/$defs/Example",
+        "$defs": {
+            "Example": {
+                "description": "\nfirst line\n\n    indented\n\nlast line.\n",
```
jcrist, line 1180: If we do want to strip leading/trailing whitespace (see above), we'll need to update this test too:

```suggestion
                "description": "first line\n\n    indented\n\nlast line.",
```

tests/test_schema.py
```diff
@@ -1159,3 +1159,29 @@ class Ex2(msgspec.Struct):
             "required": ["b"],
         },
     }
+
+
+def test_multiline_docstrig():
```
jcrist, line 1164: typo

```suggestion
def test_multiline_docstring():
```

jcrist (commented):
Thanks for the fix! Overall this looks good to me, just a few small comments.

msgspec/_json_schema.py
trendels: I think `strip()` is too much: If the docstring starts with an indented code block, like this:

```
"""
    some code

regular paragraph.
"""
```
it would be turned into `"some code\n\nregular paragraph"`.

It should be save to remove just the leading and trailing newlines, though.

trendels: I've pushed a commit that strips newlines and updated the test.

jcrist (approved):
Thanks!

Issue #359, opened 2023-04-02
Improve the type hints on `msgspec.structs.asdict` and `msgspec.structs.astuple`


guacs:
### Description

Would it be possible to add better type hints on the `asdict` and `astuple` helper methods? Right now the type is `dict` which is resolved as `dict[Unknown, Unknown]` by type checkers like Pyright. This then causes an error as shown below:

```python
from msgspec import Struct
from msgspec import structs

class Point(Struct):

    x: int
    y: int

p = Point(1, 2)

# Causes pyright to complain the following:
# Type of "p_dict" is partially unknown
# Type of "p_dict" is "dict[Unknown, Unknown]"
p_dict = structs.asdict(p)
```

I was thinking a simple fix would be to mark the return type as `dict[str, Any]` and `tuple[Any, ...]` as shown below.

```python
# Current version
def asdict(struct: Struct) -> dict: ...
def astuple(struct: Struct) -> tuple: ...

# Proposed version
def asdict(struct: Struct) -> dict[str, Any]: ...
def astuple(struct: Struct) -> tuple[Any, ...]: ...
```

Is there something I'm missing in my thought process, or would this be possible?

jcrist:
Makes sense to me! Should be fixed by #360.

PR #418, opened 2023-05-27
Support `set` & `frozenset` as inputs to `from_builtins`

### LINKED ISSUE #417: Deserializing sets fails in from_builtin
### Description

When a set is used in a dictionary and that is used in the deserialization using the `from_builtins` API, a `ValidationError` is raised. The following is a reproduction of the issue:

```python
import msgspec
from msgspec import Struct

class Foo(Struct):

    bar: set[int]

foo = Foo({1, 2})
foo_dict = msgspec.structs.asdict(foo)

decoded_foo = msgspec.from_builtins(foo_dict, type=Foo)
```

This results in the following error:
```python
msgspec.ValidationError: Expected `array`, got `set` - at `$.bar`
```

jcrist:
Previously this was intentionally omitted, as no known Python serialization libraries will return a `set` object from a `loads` method. However, it's easy to support, and if users are using `from_builtins` to coerce from other sources then a `set`/`frozenset` type might pop up.

Fixes #417.

### DIFF
--- a/docs/source/converters.rst
+++ b/docs/source/converters.rst
@@ -76,8 +76,8 @@ Since all protocols are different, `to_builtins` and `from_builtins` have a few
 configuration options:
 
 - ``builtin_types``: an iterable of additional types to treat as builtin types,
-  beyond the standard `dict`, `list`, `tuple`, `str`, `int`, `float`, `bool`,
-  and `None`.
+  beyond the standard `dict`, `list`, `tuple`, `set`, `frozenset`, `str`,
+  `int`, `float`, `bool`, and `None`.
 
 - ``str_keys``: whether the wrapped protocol only supports strings for object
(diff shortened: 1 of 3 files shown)


guacs:
@jcrist you're incredible man. Such a quick response and such a quick fix is incredible.

jcrist:
No problem, happy to help - thanks for reporting the issue!

jcrist:
This *almost* broke the project record for fastest turnaround time on an issue (#359 is the current record at 21 mins open -> close), which I just realized that was also your issue :sweat_smile: . Hopefully I'm not setting an unrealistic expectation here - I'm not _always_ online, you've just gotten lucky twice :).

guacs:
Haha, next time I'll be expecting even faster times :P

PR #419, opened 2023-05-30
Add `attributes` option to `from_builtins`

### LINKED ISSUE #416: from_builtins, but for objects?  (e.g instantiate from SQLAlchemy models)
### Description

Is there an equivalent to `from_builtins`, but for objects?  Or maybe `msgspec.obj.encode` / `msgspec.obj.decode` ?

I'd like to be able to encode/decode into regular Python objects, specifically SQLAlchemy.

Given:

`tables.py`

```python
from sqlalchemy import ForeignKey
from sqlalchemy.orm import DeclarativeBase, Mapped, mapped_column, relationship
from typing import List


# SQLAlchemy 2.0 Models
class Base(DeclarativeBase):
    pass


class GrommetTable(Base):
    __tablename__ = 'grommet'

    id: Mapped[int] = mapped_column(primary_key=True)
    name: Mapped[str]
    widget_id: Mapped[int] = mapped_column(ForeignKey("widget.id"))

    widget: Mapped["WidgetTable"] = relationship(back_populates="grommets")


class WidgetTable(Base):
    __tablename__ = 'widget'

    id: Mapped[int] = mapped_column(primary_key=True)
    name: Mapped[str]

    grommets: Mapped[List["GrommetTable"]] = relationship(back_populates="widget")
```

 and:

`models.py`

```python
from msgspec import Struct


class Grommet(Struct):
    id: int
    name: str
    widget_id: int


class Widget(Struct):
    id: int
    name: str
    grommets: List[Grommet]
```

and (create test data):

```python
from sqlalchemy import create_engine
from sqlalchemy.orm import Session

from models import Grommet, Widget
from tables import GrommetTable, WidgetTable

engine = create_engine("sqlite+pysqlite:///:memory:", echo=True)

# Create table structure
Base.metadata.create_all(engine)

# Create demo data
g1 = GrommetTable(id=1, name="Grommet 1")
g2 = GrommetTable(id=2, name="Grommet 2")
g3 = GrommetTable(id=3, name="Grommet 3")
g4 = GrommetTable(id=4, name="Grommet 4")
g5 = GrommetTable(id=5, name="Grommet 5")
g6 = GrommetTable(id=6, name="Grommet 6")

w1 = WidgetTable(id=1, name="Widget 1", grommets=[g1, g2, g3])
w2 = WidgetTable(id=2, name="Widget 2", grommets=[g4, g5, g6])

with Session(engine) as session:
    session.add_all([w1, w2])
    session.commit()
```

It would be great to be able to do something like the below:

```python
from sqlalchemy import select
from sqlalchemy.orm import selectinload

from models import Widget
from tables import WidgetTable

with Session(engine) as session:
    query = select(WidgetTable).options(selectinload(WidgetTable.grommets))
    result = session.execute(query)

    for widget in result.scalars():
        widget = from_obj(widget, Widget)   # NEW FEATURE
        print(widget)
```

Output:

```python
Widget(id=1, name='Widget 1', grommets=[Grommet(id=1, name='Grommet 1', widget_id=1), Grommet(id=2, name='Grommet 2', widget_id=1), Grommet(id=3, name='Grommet 3', widget_id=1)])
Widget(id=2, name='Widget 2', grommets=[Grommet(id=4, name='Grommet 4', widget_id=2), Grommet(id=5, name='Grommet 5', widget_id=2), Grommet(id=6, name='Grommet 6', widget_id=2)])
```

The current way to do this would be something like the below implementation, which does work, but feels like there should be something more elegant.

```python
from msgspec import from_builtins
from msgspec.inspect import ListType, type_info

def from_obj(obj, type, _data_only=False):
    data = {}
    for field in type_info(type).fields:
        if isinstance(field.type, ListType):
            value = [from_obj(o, field.type.item_type.cls, _data_only=True) for o in getattr(obj, field.name)]
        else:
            value = getattr(obj, field.name)

        data[field.name] = value

    if _data_only:
        return data
    else:
        return from_builtins(data, type)
```

I think it's most closely related to the Encode/Decode because the input could have more attributes than the `Struct`, just as JSON input could have more keys than the `Struct` when utilizing [Schema Evolution](https://jcristharif.com/msgspec/schema-evolution.html)

This feature would be useful because it allows a common, very fast internal data structure to be used when reading/writing from common data serialization formats (`json`, `toml`, `yaml` `msgpack`), but also from a large number of common data storage engines (`SQLite`, `MySQL` / `MariaDB`, `PostgreSQL`, etc..) 

Pydantic v1 has a similar system with [from_orm()](https://docs.pydantic.dev/latest/usage/models/#recursive-orm-models), which I believe has now been replaced with [model_validate()](https://docs.pydantic.dev/dev-v2/usage/models/#recursive-models_1) in v2.

jcrist:
This adds a new `attributes` option to `from_builtins`. If enabled, arbitrary objects may be coerced to `Struct`/`dataclasses`/`attrs` types by extracting attributes from the input object matching fields in the output type. One use case is converting database query results (ORM or otherwise) to msgspec structured types.

Fixes #416.

### DIFF
--- a/docs/source/converters.rst
+++ b/docs/source/converters.rst
@@ -86,6 +86,12 @@ configuration options:
   supports string values. This is useful for completely untyped protocols like
   URL querystrings, where only string values exist.
 
+- ``attributes``: `from_builtins` only. If True, input objects may be coerced
+  to ``Struct``/``dataclass``/``attrs`` types by extracting attributes from the
+  input matching fields in the output type. One use case is converting database
+  query results (ORM or otherwise) to msgspec structured types. The default is
+  False.
+
 - ``enc_hook``/``dec_hook``: the standard keyword arguments used for
   :doc:`extending` msgspec to support additional types.
 
--- a/msgspec/__init__.pyi
+++ b/msgspec/__init__.pyi
@@ -157,6 +157,7 @@ def from_builtins(
     str_keys: bool = False,
     str_values: bool = False,
     builtin_types: Union[Iterable[type], None] = None,
+    attributes: bool = False,
     dec_hook: Optional[Callable[[type, Any], Any]] = None,
 ) -> T: ...
 @overload
@@ -166,6 +167,7 @@ def from_builtins(
     *,
     str_keys: bool = False,
     builtin_types: Union[Iterable[type], None] = None,
+    attributes: bool = False,
     dec_hook: Optional[Callable[[type, Any], Any]] = None,
 ) -> Any: ...
 
--- a/msgspec/_core.c
+++ b/msgspec/_core.c
@@ -4625,6 +4625,15 @@ ms_validation_error(const char *got, TypeNode *type, PathNode *path) {
     return NULL;
 }
 
+static void
+ms_missing_required_field(PyObject *field, PathNode *path) {
+    ms_raise_validation_error(
+        path,
+        "Object missing required field `%U`%U",
+        field
+    );
+}
+
 static PyObject *
 ms_invalid_cstr_value(const char *cstr, Py_ssize_t size, PathNode *path) {
     PyObject *str = PyUnicode_DecodeUTF8(cstr, size, NULL);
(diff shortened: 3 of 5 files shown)


dmckeone:
I've been playing with this a little and it's been working well for every case I've tried.

jcrist:
Glad to hear it!

PR #457, opened 2023-06-28
Improve JSON decimal encode performance

jcrist:
This improves the performance of encoding JSON strings, mostly by rearranging some work and unrolling some loops.


jcrist:
oop, borked my local branch. Will fix and reopen.

PR #459, opened 2023-06-28
Improve JSON str encode performance

jcrist:
This improves the performance of encoding JSON strings, mostly by rearranging some work and unrolling some loops.

### DIFF
--- a/msgspec/_core.c
+++ b/msgspec/_core.c
@@ -51,6 +51,9 @@ ms_popcount(uint64_t i) {                            \
 
 #define DIV_ROUND_CLOSEST(n, d) ((((n) < 0) == ((d) < 0)) ? (((n) + (d)/2)/(d)) : (((n) - (d)/2)/(d)))
 
+/* These macros are used to manually unroll some loops */
+#define repeat8(x) { x(0) x(1) x(2) x(3) x(4) x(5) x(6) x(7) }
+
 #define is_digit(c) (c >= '0' && c <= '9')
 
 /* Easy access to NoneType object */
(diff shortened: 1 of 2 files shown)


jcrist:
A quick benchmark of string performance:

```python
import random
import string
from time import perf_counter_ns

from orjson import dumps
from msgspec.json import encode

CHARS = string.ascii_letters
CHARS_WITH_ESCAPES = "\x01\n'\"" + string.ascii_letters


def randstr(chars):
    k = random.randint(2, 64)
    return "".join(random.choices(chars, k=k))


N = 1000
M = 10000
random.seed(42)

for header, chars in [("No Escapes:", CHARS), ("With Escapes:", CHARS_WITH_ESCAPES)]:
    print(header)
    data = [randstr(chars) for _ in range(N)]

    start = perf_counter_ns()
    for _ in range(M):
        encode(data)
    stop = perf_counter_ns()
    print(f"- msgspec: {(stop - start) / (1000 * M):.1f} us")

    start = perf_counter_ns()
    for _ in range(M):
        dumps(data)
    stop = perf_counter_ns()
    print(f"- orjson: {(stop - start) / (1000 * M):.1f} us")
```

**Before**

```
$ python bench.py
No Escapes:
- msgspec: 29.7 us
- orjson: 21.5 us
With Escapes:
- msgspec: 53.1 us
- orjson: 67.0 us
```

**This PR**

```
$ python bench.py
No Escapes:
- msgspec: 18.6 us
- orjson: 22.0 us
With Escapes:
- msgspec: 34.2 us
- orjson: 68.1 us
````

Notes:

- This PR makes msgspec ~1.6x faster for encoding all strings
- `orjson` is optimized towards the common case of encoding strings that don't have characters needing escaping (`\n`, `\t`, `\r`, `\f`, `\b`, `"`, `'`, `\`, and anything in `\x01` - `\x32`). In the case of strings where escape codes _are_ needed, orjson's performance tanks. `msgspec` was already faster than `orjson` for encoding strings that require escape codes, now we're ~2x faster in that case. We're also now faster in the common case.

Even though this PR unrolls some loops, it actually results in a (negligible) *decrease* in binary size since we delete a non-inlined helper function for handling the escape code case.

PR #460, opened 2023-06-30
Wrap dec_hook errors in `ValidationError`

### LINKED ISSUE #456: Wrap errors raised by `dec_hook` in `ValidationError`
### Description

Currently if `dec_hook` raises an error, the error is raised as is by the corresponding `decode` call. This is the only place where a `msgspec.ValidationError` or `msgspec.DecodeError` isn't raised by `decode`. To fix this, we should:

- Update the docs to recommend raising `NotImplementedError` in a `dec_hook` when a type isn't handled by the `dec_hook`
- wrap any `TypeError` or `ValueError` raised by `dec_hook` in a `ValidationError` with additional context. Ideally we could also used chained exceptions here, although the c-api makes handling this a bit tricky.

jcrist:
This modifies decoding of custom types to wrap errors raised by a `dec_hook` with `ValidationError`. Only `TypeError` or `ValueError` (and subclasses) exceptions are wrapped, all other exceptions are raised directly.

If an exception is wrapped with a `ValidationError`, the original exception is set as the `__cause__` (and `__context__`). This means the original exception will still show up in tracebacks, and is available for introspection if needed.

This also changes the recommendation for raising `TypeError` on unsupported types in `dec_hook`/`enc_hook` to raising a `NotImplementedError`. No previous functionality was based on this recommendation, so this isn't a breaking change in anyway, just a change in recommendation of how best to use `msgspec`.

Fixes #456.

### DIFF
--- a/docs/source/extending.rst
+++ b/docs/source/extending.rst
@@ -24,7 +24,12 @@ These should have the following signatures:
 
     def dec_hook(type: Type, obj: Any) -> Any:
         """Given a type in a schema, convert ``obj`` (composed of natively
-        supported objects) into an object of type ``type``"""
+        supported objects) into an object of type ``type``.
+
+        Any `TypeError` or `ValueError` exceptions raised by this method will
+        be considered "user facing" and converted into a `ValidationError` with
+        additional context. All other exceptions will be raised directly.
+        """
         pass
 
     def ext_hook(code: int, data: memoryview) -> Any:
@@ -78,8 +83,8 @@ objects to/from objects, which are then natively handled by ``msgspec``.
             # convert the complex to a tuple of real, imag
             return (obj.real, obj.imag)
         else:
-            # Raise a TypeError for other types
-            raise TypeError(f"Objects of type {type(obj)} are not supported")
+            # Raise a NotImplementedError for other types
+            raise NotImplementedError(f"Objects of type {type(obj)} are not supported")
 
 
     def dec_hook(type: Type, obj: Any) -> Any:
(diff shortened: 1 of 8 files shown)


jcrist:
Demo of behavior

```python
In [1]: import msgspec, pydantic

In [2]: class Point(pydantic.BaseModel):
   ...:     x: int
   ...:     y: int
   ...: 

In [3]: class Test(msgspec.Struct):
   ...:     point: Point
   ...: 

In [4]: def dec_hook(typ, val):
   ...:     if issubclass(typ, pydantic.BaseModel):
   ...:         return typ.model_validate(val)
   ...:     raise NotImplementedError
   ...: 

In [5]: dec = msgspec.json.Decoder(type=Test, dec_hook=dec_hook)

In [6]: dec.decode(b'{"point": {"x": 1, "y": 2}}')  # works
Out[6]: Test(point=Point(x=1, y=2))

In [7]: try:
   ...:     dec.decode(b'{"point": {"x": "a", "y": "b"}}')
   ...: except msgspec.ValidationError as exc:
   ...:     error = exc  # work around auto-del of exc
   ...: 

In [8]: error
Out[8]: msgspec.ValidationError("2 validation errors for Point\nx\n  Input should be a valid integer, unable to parse string as an integer [type=int_parsing, input_value='a', input_type=str]\n    For further information visit https://errors.pydantic.dev/0.39.0/v/int_parsing\ny\n  Input should be a valid integer, unable to parse string as an integer [type=int_parsing, input_value='b', input_type=str]\n    For further information visit https://errors.pydantic.dev/0.39.0/v/int_parsing - at `$.point`")

In [9]: error.__cause__
Out[9]: 
2 validation errors for Point
x
  Input should be a valid integer, unable to parse string as an integer [type=int_parsing, input_value='a', input_type=str]
    For further information visit https://errors.pydantic.dev/0.39.0/v/int_parsing
y
  Input should be a valid integer, unable to parse string as an integer [type=int_parsing, input_value='b', input_type=str]
    For further information visit https://errors.pydantic.dev/0.39.0/v/int_parsing
```

If the `str(exc)` for the wrapped error has multiple lines (like pydantic errors) then the path suffix (the `"- at $.point"` bit) looks a little weird, but for now I think that's fine. For normal one-line exceptions the wrapped errors look identical to the builtin errors, which is the desired effect.

PR #461, opened 2023-07-02
Migrate from `flake8` & `isort` to `ruff` and update pre-commit config

samiashi:
Migrate from `flake8` & `isort` to `ruff` and update pre-commit config

https://github.com/astral-sh/ruff


jcrist:
Thanks!

PR #465, opened 2023-07-04
Support encoding decimals as numbers

### LINKED ISSUE #440: Option to JSON encode Decimal types as numbers, not strings
> We like that msgspec handles decimal however they are embedded as strings. Here is some finding from another discussion around handling decimals.

> https://github.com/meltano/sdk/issues/1046#issuecomment-1535493163

> Do we think that msgspec might output decimals like simplejson without the double quotes in the future?

_Originally posted by @s7clarke10 in https://github.com/jcrist/msgspec/issues/349#issuecomment-1586444502_

jcrist:
Previously we always encoded `decimal.Decimal` values as strings. While this is the best format for these types (prevents precision loss, widely supported), sometimes there is a need to encode `Decimal` objects the same as numbers. This adds a new `decimal_format` option to all `Encoder` classes. This value defaults to `"string"` (to encode decimals as strings), but may be set to `"number"` to encode them the same as the protocol's numeric types.

We opt not to support this configuration in the top-level `encode` functions for now, since this is a less-common setting.

Fixes #440.

### DIFF
--- a/docs/source/supported-types.rst
+++ b/docs/source/supported-types.rst
@@ -419,8 +419,8 @@ When decoding, both hyphenated and unhyphenated forms are supported.
 -----------
 
 `decimal.Decimal` values are encoded as their string representation in all
-protocols. This ensures no precision loss during serialization, as would happen
-with a float representation.
+protocols by default. This ensures no precision loss during serialization, as
+would happen with a float representation.
 
 .. code-block:: python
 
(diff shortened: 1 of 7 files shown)


jcrist:
A quick demo:

```python
In [1]: import msgspec, decimal

In [2]: x = decimal.Decimal("1.3")

In [3]: msgspec.json.encode(x)  # by default encodes as a string
Out[3]: b'"1.3"'

In [4]: encoder = msgspec.json.Encoder(decimal_format="number")  # create an encoder to override

In [5]: encoder.encode(x)  # now encodes as a number
Out[5]: b'1.3'

In [6]: import json

In [7]: decimal.Decimal(json.loads(b'1.3'))  # downstream tools may have trouble loading decimals from numbers without precision loss
Out[7]: Decimal('1.3000000000000000444089209850062616169452667236328125')

In [8]: msgspec.json.decode(b'1.3', type=decimal.Decimal)  # msgspec has no issues with this though, but other tools may
Out[8]: Decimal('1.3')
```

jcrist:
If you wish to try this out before the next release, please see our instructions for [installing from GitHub](https://jcristharif.com/msgspec/install.html#installing-from-github).

s7clarke10:
Brilliant, thank you very much for taking this on board and providing the ability to output as a decimal. I do take you point that there may be a loss of precision. I will utilize configuration to select whether we wish to output as a string or decimal using this feature.

PR #470, opened 2023-07-06
Support `__post_init__` for `Struct` types

### LINKED ISSUE #442: Post-init hook
### Description

One of the features that allows for easy runtime validation in dataclasses.dataclass is the post init hook. By providing an opt-in option for Structs, most use cases for libraries like attrs / dataclass could be entirely replaced, and presumably with better performance.

This could be done as a reserved \__post_init__ as on data class, or following attrs, \__msgspec_post_init__. By default this method simply returns (and could probably be elided entirely), should not cost much performance.

### Example
Currently we (Bystro team) use Struct for serialization/deserialization (and the validation at the IO boundary that this enables), but need attrs to enable runtime validation, for instance https://github.com/bystrogenomics/bystro/blob/4dfdeee8c592a69f840013ea95cfd12a5fc3ae99/python/python/bystro/ancestry/ancestry_types.py#L137.
- It would be possible for us to wrap Struct in a custom class that carries this post init hook, but that is somewhat less ergonomic.
- In most, but I don't think all cases, it would be possible to do the validation as a dec_hook. Similarly, for enc_hook
  - For instance json -> Obj1 -> {Obj2 -> model1, Obj3 -> model2} -> Result -> json, where we want to validate Obj2 and Result. Obj2 and Obj3 may be derived from Obj1, but need to validate the content of the values, not just the types. It would be  awkward to require the dec_hook for Obj1 to have knowledge of Obj2 and Obj3.
- We could also continue to use Struct + attrs, but there is nothing I want in attrs besides post-init validation and requiring a 2nd library just for that feature is heavy.

You've put a whole lot of thought into this library, and it may be that you already have a mechanism for handling this, I would welcome feedback. I really love what you've done here, and see Struct as the container that I want to use for all of my data modeling needs.

Thanks very much for considering!

jcrist:
This adds support for an optional `__post_init__` hook method on `msgspec.Struct` types. If defined, this method is called at the end of the generated `__init__` method. It can be used to customize the builtin `__init__` as needed, as well as provides a spot to do some additional validation.

As with `dec_hook` errors, any `ValueError` or `TypeError` raised in `__post_init__` will be wrapped in a `ValidationError` when decoding. In the long run I plan to add support for more granular (and obviously named) custom validator support, but for now doing some validation in `__post_init__` is a fine and simple solution. There are other use cases for `__post_init__` besides validation, so we'd want to add this ability either way.

Fixes #442.

### DIFF
--- a/docs/source/structs.rst
+++ b/docs/source/structs.rst
@@ -67,7 +67,8 @@ annotations:
     True
 
 Note that it is forbidden to override ``__init__``/``__new__`` in a struct
-definition, but other methods can be overridden or added as needed.
+definition, but other methods can be overridden or added as needed. If you need
+to customize the generated ``__init__``, see :ref:`struct-post-init`.
 
 The struct fields are available via the ``__struct_fields__`` attribute (a
 tuple of the fields in argument order ) if you need them. Here we add a method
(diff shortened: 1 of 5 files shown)


jcrist:
A quick example:

```python
>>> import msgspec

>>> class Interval(msgspec.Struct):
...     low: float
...     high: float
...
...     def __post_init__(self):
...         if self.low > self.high:
...             raise ValueError("`low` may not be greater than `high`")

>>> Interval(1, 2)  # valid interval
Interval(low=1, high=2)

>>> Interval(2, 1)  # invalid interval
  File "<stdin>", line 6, in __post_init__
ValueError: `low` may not be greater than `high`

>>> msgspec.json.decode(b'{"low": 2, "high": 1}', type=Interval)  # invalid interval from JSON
  File "<stdin>", line 6, in __post_init__
ValueError: `low` may not be greater than `high`

The above exception was the direct cause of the following exception:

msgspec.ValidationError: `low` may not be greater than `high`
```

akotlar:
Amazing, thank you!

PR #473, opened 2023-07-06
Fixed recompilation in mashumaro

Fatal1ty:
Hi, author of mashumaro here 👋

I appreciate your work and I don't think anyone could outrun msgspec in speed but I found a bug in how mashumaro is benchmarked. Due to [how mashumaro works](https://github.com/Fatal1ty/mashumaro#how-does-it-work), it tries to compile methods `from_*` / `to_*` at import time by default but if it fails because of forward references, then these methods will consist of the code that will compile the real methods and replace with them on demand. In the original benchmark there were forward references due to `from __future__ import annotations` which were leading to expensive compilation on each call of `from_*` / `to_*`.

I fixed this with lambda functions and added ~[discriminated unions](https://github.com/Fatal1ty/mashumaro#discriminator) and~ [explicit lazy compilation](https://github.com/Fatal1ty/mashumaro#lazy_compilation-config-option) to reduce import time and speed it up slightly.

mashumaro + orjson is still slower that msgspec but not by an order of magnitude. I'm leaving the numbers to you because my environment is different from the one used to run the benchmark for the documentation.  I also encourage you to update benchmark results with the new version of mashumaro that has some new optimizations. Just for reference, I got this:

```
Benchmark - 1000 objects with validation
msgspec
  dumps: 133.01 us
  loads: 277.36 us
  total: 410.37 us
pydantic
  dumps: 13646.91 us
  loads: 14558.10 us
  total: 28205.02 us
cattrs
  dumps: 1801.90 us
  loads: 1976.82 us
  total: 3778.72 us
mashumaro
  dumps: 675.22 us
  loads: 1277.56 us
  total: 1952.78 us
```

P.S. With only two variants in Union I don't see performance boost with using `Discriminator` feature but if there were more variants, it would take place. It's up to you whether to use it, you might want to alter your benchmarks in the future. Anyway, it will look like this:
```python
contents: List[
    Annotated[
        Union[FileMashumaro, DirectoryMashumaro],
        Discriminator(field="type", include_supertypes=True),
    ]
]
```

### DIFF
--- a/benchmarks/bench_validation.py
+++ b/benchmarks/bench_validation.py
@@ -130,6 +130,9 @@ class FileMashumaro(DataClassORJSONMixin):
     nbytes: int
     type: Literal["file"] = "file"
 
+    class Config:
+        lazy_compilation = True
+
 
 @dataclasses.dataclass
 class DirectoryMashumaro(DataClassORJSONMixin):
@@ -140,13 +143,16 @@ class DirectoryMashumaro(DataClassORJSONMixin):
     contents: List[Union[FileMashumaro, DirectoryMashumaro]]
     type: Literal["directory"] = "directory"
 
+    class Config:
+        lazy_compilation = True
+
 
 def bench_mashumaro(n):
     return bench(
-        lambda x: x.to_jsonb(),
-        DirectoryMashumaro.from_json,
+        lambda x: x.to_json(),
+        lambda x: DirectoryMashumaro.from_json(x),
         n,
-        DirectoryMashumaro.from_dict,
+        lambda x: DirectoryMashumaro.from_dict(x),
     )
 
 


jcrist:
Thanks @Fatal1ty!

> I fixed this with lambda functions and added [discriminated unions](https://github.com/Fatal1ty/mashumaro#discriminator) and [explicit lazy compilation](https://github.com/Fatal1ty/mashumaro#lazy_compilation-config-option) to reduce import time and speed it up slightly.

It looks like both these features were added in the last few weeks? I'd be happy to update the benchmarks to include them if they represent how a typical Mashumaro user would write them. No sense in having a flawed benchmark.

I did play around with your PR a bit and found that removing the `lazy_compilation = True` setting didn't cause a regression, the speedup seems to be 100% due to the delayed attribute access? From reading through mashumaro's code, I think that makes sense here. Is there something mashumaro could do to avoid this potential footgun? Perhaps caching compilation on a private `_from_dict` method and having the `from_dict` method itself call that so a strong reference to `from_dict` before compilation wouldn't result in this issue?

Either way, happy to merge as is. I'll push up new benchmark results later before the upcoming release.

jcrist:
Also, would you be open to adding `msgspec` to your benchmarks? I'd be happy to push up a PR if so.

Fatal1ty:
> It looks like both these features were added in the last few weeks? I'd be happy to update the benchmarks to include them if they represent how a typical Mashumaro user would write them. No sense in having a flawed benchmark.

Yes, they were added recently. As you wrote at the top of benchmarks documentation, benchmarks are hard. I think it's up to the user to choose the way to deal with their structures. There is no silver bullet here and each case requires special handling if you want to have the best performance. In case of your _current_ benchmarks using discriminator rather makes if worse at least in my environment.

> I did play around with your PR a bit and found that removing the lazy_compilation = True setting didn't cause a regression

Yes, you're right. Explicit delayed compilation doesn't improve deserialization since it's already delayed due to the use of `from __future__ import annotations` at the top. I just added `lazy_compilation = True` to make it explicit for future edits of the benchmark.

> Is there something mashumaro could do to avoid this potential footgun? Perhaps caching compilation on a private _from_dict method and having the from_dict method itself call that so a strong reference to from_dict before compilation wouldn't result in this issue?

It will certainly eliminate this potential footgun but at the cost of an additional call. I'm not ready to sacrifice performance at this point. Perhaps, it's worth highlighting this edge case in the documentation.

> Also, would you be open to adding msgspec to your benchmarks? I'd be happy to push up a PR if so.

Honestly, I wouldn't be open for this and I hope you will understand the reason. Like many other projects that include benchmarks in their repo, I do it to promote my own project. I do acknowledge that msgspec's Structs in C are way faster than dataclasses and adding comparison to your project will be against the goal. I understand if you remove my library from your benchmark but in any case you have to show a comparison with other libraries just to promote your own library.

Also a little off topic but there are already so many serialization libraries that it seems like the time has come for an independent benchmark to appear. I imagine how cool it would be to have something similar to [Web Framework Benchmarks](https://www.techempower.com/benchmarks/).

jcrist:
> Honestly, I wouldn't be open for this and I hope you will understand the reason. Like many other projects that include benchmarks in their repo, I do it to promote my own project. I do acknowledge that msgspec's Structs in C are way faster than dataclasses and adding comparison to your project will be against the goal.

I'm sorry to hear that, but honestly can't agree with that reasoning. IMHO while a benchmark *may* be used for marketing purposes, their primary purpose is education. They can help educate users on the tradeoffs libraries make - some are better for one use case and some better for a different use case. And they can help educate the developer on how their project compares on various metrics.

I'd consider intentionally leaving a library out of a benchmark just because it's faster than your project equivalent to lieing through omission. There are many things that mashumaro does well (and better than msgspec!) that you could continue to use to make a pitch for your library. You could say it's the "fastest pure python library". Or "more flexible than msgspec while maintaining some of its speed". But making an argument based on yours being the fastest in your benchmarks (simply because you've omitted any competition that is faster) is frankly misleading to users.

Anyway, it being your project you're welcome to do what you want with it. I'll respect your wishes. But I disagree with your reasoning here and can't consider it ethical behavior as a developer to willfully mislead.

Issue #492, opened 2023-07-26
Add support for repr customisation


aedify-swi:
### Description

As mentioned in https://github.com/jcrist/msgspec/issues/491, it would be great to be able to change the repr of specific fields. [Attrs does this.](https://www.attrs.org/en/stable/examples.html#other-goodies)

I usually do this for sensitive fields or subclasses, where I want to reduce the print output (e.g. I am only interested in a field of that subclass for logging).

I'd suggest adding this to Meta:

```python

Password = Annotated[str, msgspec.Meta(repr=lambda x: "***")]

class Credentials(msgspec.Struct):
    user: str
    password: Password

print(Credentials("Jim", "secret"))
# Credentals(user="Jim", password="***")
```
What do you think?

jcrist:
I think this could be in scope. I already had plans to support excluding fields from a repr, from looking at the attrs docs they also support passing a callable to `repr` to override how a field is repr'd. Seems like a natural extension.

The api would be as follows:

```python
class Example(msgspec.Struct):
    x: int   # always repr'd like normal
    y: int = msgspec.field(repr=False)  # excluded from the repr
    z: int = msgspec.field(repr=hex)  # uses the passed callable to repr the field
```

Julienh:
Hello,
I just found this library which I plan to use for a project, performances are great!
However I was looking for a solution to exclude fields from the output, this solution would be perfect!
Do you plan to add it in a future release please?

lmmx:
I contributed a PR for this in #543, apologies I forgot to comment here to update those subscribed to the issue

Lawouach:
That'd be fantastic indeed.

I'm using msgspec for a cli and its settings. Being able to hide sensitive data from being dump at will would be greatly appreciated.

PR #543, opened 2023-09-06
feat: add `repr` kwarg to `field` function and `Field` C struct

lmmx:
I picked up ticket #492 to help contribute this feature, as I'd really like to use this library for a project which requires excluding fields (a linked list with mutual references which would produce infinite loop if `repr`'d).

To begin, I have done the following:

- Extended the field Python function to accept a new keyword argument `repr` which controls the inclusion of the field in the generated `__repr__` string for the struct.
- Added a new attribute `repr` to the `Field` C struct to store the value of the `repr` keyword argument.
- Updated the `Field` C struct docstring to include details about the new `repr` kwarg.

Then to complete the feature I followed the lead of the other lookup dictionaries (ending in `_lk`) like `defaults_lk`, and used a list that was converted to a tuple, as I'm sure you're familiar. For boolean configuration I have only allowed this to be `Py_True` or `Py_False`, and defaulting to `Py_True` if the value type is not a `Field` object. With a little extra logic here this could be extended to handle arbitary callables if desired.

I tried to follow the example of how to represent `repr` in C from where other arguments use `int` to represent booleans. Please take a careful look at this to ensure it fits the house style.

```py
>>> class A(msgspec.Struct):
...     a: int = msgspec.field(repr=False)
...     b: int
... 
>>> A(a=1, b=2)
A(b=2)
```

**Update** I've added tests and remembered to also implement this feature in the `__rich_repr__` method internal too.
**Update** I've implemented support for callables as described in the original ticket
**Update** I reviewed the memory management (incrementing/decrementing), refcount tests continue to pass


lmmx:
I've implemented support for callable arguments and added all the test coverage I think this feature needs now, looks ready to review @jcrist :tada: Please let me know of any changes you'd like me to make

dhirschfeld:
This would be super valuable for e.g. not accidentally exposing secrets in logs.

lmmx:
Would you like any changes to this to shepherd it along @jcrist ? :slightly_smiling_face:

PR #561, opened 2023-10-05
Fix module init for Python 3.12

jcrist:
Python 3.12 changed module init slightly (this may actually be a CPython 3.12 bug, it's not 100% clear). Before Python 3.12 the module wouldn't be available via `PyState_FindModule` until it was fully initialized. Now the module is immediately available, but in an invalid state leading to segfaults. The workaround is to manually call `PyState_AddModule` to ensure the module is always available.

In the long run we should move to using multiphase module init which should avoid this problem entirely.

### DIFF
--- a/msgspec/_core.c
+++ b/msgspec/_core.c
@@ -6043,15 +6043,10 @@ StructMeta_new_inner(
     /* Fill in struct offsets */
     if (structmeta_construct_offsets(&info, cls) < 0) goto cleanup;
 
-    /* Cache access to __post_init__ (if defined).
-     * XXX: When StructMeta is defined, the module hasn't finished initializing
-     * yet so `mod` will be NULL here. */
-    if (mod != NULL) {
-        cls->post_init = PyObject_GetAttr((PyObject *)cls, mod->str___post_init__);
+    /* Cache access to __post_init__ (if defined). */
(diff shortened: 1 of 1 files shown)


jcrist:
For additional context, the segfault here:

- Only occurs on CPython 3.12
- Only has a faulty memory access at import time. If `msgspec` successfully imports then you won't be hit by this issue.
- Doesn't segfault if there has been sufficient code run before a `msgspec` import (I'm _guessing_ this is due to the faulty memory access being in a region that was already allocated and used previously in this case). In particular, this means that the test suite never segfaults due to this issue, which is how it went uncaught until the conda-forge builds started failing ([here](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=795634&view=logs&j=517fe804-fa30-5dc2-1413-330699242c05&t=c7d0fcb4-3a10-5b26-dfea-e879a93fc2ab)).

Rather than add a test for this esoteric issue, I've:
- Manually validated that the fix here resolved it
- More finely read the [relevant docs](https://docs.python.org/3/c-api/module.html#c.PyState_AddModule) which confirm that the code here is as intended
- Read the relevant CPython source to also confirm this behavior

I'll cut a new release immediately after merging this.

akotlar:
Thanks

PR #586, opened 2023-11-18
docs: udpate `defstruct` docs with mention of field spec; add example;

alexei:
Update `defstruct` docs to mention it's possible to use `msgspec.field`.

Addresses https://github.com/jcrist/msgspec/issues/580


alexei:
@jcrist I suppose the documentation site is generated automatically?

jcrist:
Thanks for the PR (and apologies for the delayed response)! I've pushed a small fixup, but otherwise this looks good to me.

> I suppose the documentation site is generated automatically?

Yes, it's built and deployed as part of CI. If you want to build locally see https://github.com/jcrist/msgspec/blob/main/.github/CONTRIBUTING.md#documentation.

PR #624, opened 2024-01-04
feat: support deserializing to memoryview

SunDoge:
The implementation enables zero-copy functionality when decoding msgpack by supporting deserialization directly to a memoryview. This is very useful for decoding large numpy array.

### DIFF
--- a/docs/source/api.rst
+++ b/docs/source/api.rst
@@ -150,6 +150,7 @@ Inspect
 .. autoclass:: StrType
 .. autoclass:: BytesType
 .. autoclass:: ByteArrayType
+.. autoclass:: MemoryViewType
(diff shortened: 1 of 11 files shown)


jcrist:
Thanks for the PR! I've taken the liberty to finish it up - adding support for new types needs to be done in several places.

Note that zero-copy deserialization of buffers can also be handled through [Ext](https://jcristharif.com/msgspec/extending.html#defining-a-custom-extension-messagepack-only) types. When decoding an `Ext`, the buffer passed to `ext_hook` is a `memoryview` into the original backing message. This was intended to support decoding large buffer-backed objects (like numpy arrays) in a zero-copy fashion.

PR #687, opened 2024-05-15
Preserve typing args when inspecting custom types

dcwatson:
The other inspected types carry through the typing args, seems like custom types should as well. My specific use-case is wanting to use `msgspec.inspect.type_info` on `Struct`s which may have custom `Generic` members, without having to augment it with a separate call to `msgspec.structs.fields` and parsing out the type info myself.

### DIFF
--- a/msgspec/inspect.py
+++ b/msgspec/inspect.py
@@ -1002,4 +1002,4 @@ class _Translator:
             )
             return out
         else:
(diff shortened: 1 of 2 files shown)


xiaohuanshu:
this is very useful, why wasn't it merged.
maybe should add both cls and original type in CustomType

Issue #683, opened 2024-05-08
Support numeric constraints for Decimal values


corentin-regent:
### Description

I am currently using msgspec for parsing financial (cryptocurrency) data from various exchanges. Precision in this context is crucial, so I use [Decimal](https://docs.python.org/3/library/decimal.html#decimal-objects) values quite a lot. Though for safety and documentation, I would also like to set [constraints](https://jcristharif.com/msgspec/constraints.html) for these values, however msgspec currently does not support it.

Supporting numeric constraints for the `Decimal` type would add consistency with the `int` and `float` types:

```python
>>> msgspec.json.decode('123.456', type=Annotated[float, Meta(gt=0)])
123.456
>>> msgspec.json.decode('123.456', type=Annotated[Decimal, Meta(gt=0)])
TypeError: Can only set `gt` on a numeric type - type `typing.Annotated[decimal.Decimal, msgspec.Meta(gt=0)]` is invalid
```

Additionally (may be a separate feature), supporting the `multiple_of` constraint with a `Decimal` argument (e.g. `Meta(multiple_of=Decimal('0.1'))`) would provide users with a workaround to the false negatives due to precision loss, from [this issue](https://github.com/json-schema-org/json-schema-spec/issues/312).

Siyet:
Resolved in the community fork [msgspec-arise](https://github.com/Siyet/msgspec-arise): [issue #8](https://github.com/Siyet/msgspec-arise/issues/8), released in [v0.20.2](https://github.com/Siyet/msgspec-arise/releases/tag/0.20.2).

PR #692, opened 2024-05-20
Support numeric constraints for Decimal types

corentin-regent:
This is my first time playing around with C, so please feel free to tell me if more micro-optimizations can be achieved. Same for compiler hints, notably for the `MS_INLINE` and `MS_NOINLINE` near `ms_check_decimal_constraints(...)`: I implemented it in a manner similar to what was done for floats for example, though I must admit that I did not quite understand the pattern for when to use each.

Also, I added a DevContainer for easier setup for newcomers to the project.

Closes https://github.com/jcrist/msgspec/issues/683.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,26 @@
+{
+    "name": "jcrist/msgspec",
+    "image": "mcr.microsoft.com/devcontainers/python:3.12-bookworm",
+    "postCreateCommand": "scripts/install.sh",
+    "customizations": {
+        "vscode": {
+            "extensions": [
+                "charliermarsh.ruff",
+                "ms-python.python",
+                "ms-vscode.cpptools"
+            ],
+            "settings": {
+                "C_Cpp.default.includePath": [
+                    "/usr/local/include/**"
+                ],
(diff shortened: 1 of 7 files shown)


Siyet:
Hi @cocorigon! I've run into the same issue (#683) in production - `Annotated[Decimal, Meta(gt=0)]` raises `TypeError` when used with schema_components.

Are you still planning to rebase this PR and continue working on it? If not, I'd be happy to pick it up - would open a fresh PR based on your approach, with credit of course.

No rush, just want to make sure the effort isn't duplicated.

cocorigon:
Hey @Siyet! I don't really do much Python anymore, so no, I will not work on this PR again in the near future.

So feel free to use this code as you wish!

Siyet:
Hi @cocorigon, thanks again for the original work here and for the explicit handover note above.

I've picked this up in #1006, which builds on the same approach (C-level constraint check, `multiple_of` via Decimal arithmetic) and is rebased against the current `src/msgspec/` + `tests/unit/` layout. Closing this one as superseded so review attention can stay on a single thread. Credit to you is in the #1006 description.

Closes #683 will move to #1006.

Issue #695, opened 2024-05-28
msgspec.json.decode parsing a float value


Oakington:
### Description

msgspec==0.18.6
python 3.8.10
Very odd behavior. Parsing of a stringified JSON with a float value results in the following

`
import json
import msgspec
x = '{"a": 3322.1755149730398}'
msgspec.json.decode(x)
>{'a': 9.871031767461413e-178}

json.loads(x)
>{'a': 3322.1755149730398}
`

jcrist:
Thanks for opening this. I'm unable to reproduce the issue, I suspect this is platform specific. Here's what I see on an x86 linux machine:

```python
In [1]: import msgspec

In [2]: msg = '{"a": 3322.1755149730398}'

In [3]: msgspec.json.decode(msg)
Out[3]: {'a': 3322.1755149730398}
```

Can you provide some extra platform information (OS, hardware, etc...)? The following script should output some relevant info:

```python
import platform
import sys

uname = platform.uname()

print(f"python: {platform.python_version()}")
for attr in ["machine", "processor", "system", "release"]:
    print(f"{attr}: {getattr(uname, attr)}")
print(f"libc: {platform.libc_ver()}")
print(f"byteorder: {sys.byteorder}")
```

Oakington:
python: 3.8.10
machine: AMD64
processor: Intel64 Family 6 Model 165 Stepping 2, GenuineIntel
system: Windows
release: 10
libc: ('', '')
byteorder: little

jcrist:
Hmmm, we do test on a platform that _should_ be similar enough to this one to catch errors. A few more questions:

- How did you install `msgspec`? Did you install it with `pip` from pypi? Using conda? Something else?
- Do you know if you used a prebuilt wheel for your platform (as available on pypi or through conda), or did you compile it yourself from source?

Oakington:
Installed via pip. I don't know if it was a prebuilt as I just updated my venv to 3.11, upgraded pip wheel and setuptools, re-installed msgspec and it's now working.....Very strange. Let me see if I can go back and get specific builtins versions....Will update shortly.

Oakington:
OK, reproduced with the following details. Let me know if there's anything else I can provide to track this down.

`python`
`Python 3.8.10 (tags/v3.8.10:3d8993a, May  3 2021, 11:34:34) [MSC v.1928 32 bit (Intel)] on win32`
`py -m pip --version `
[...]

PR #696, opened 2024-05-28
Error when compiling for 32-bit windows builds

jcrist:
Add a build time error that would have caught #695.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,9 +1,28 @@
+import sys
 import os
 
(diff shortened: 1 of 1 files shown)


jcrist:
FWIW we _could_ support windows 32 with some testing and bugfixes. But given it's 2024 I don't expect that platform to have much use and supporting it would cost some devtime. A compile-time error for now seems simpler.

PR #700, opened 2024-06-06
Improve `Meta` overloads

adriangb:
Also add some overloads to Meta to make it more type safe so e.g. `Meta(gt=0, pattern=r'abc')` is not allowed


sobolevn:
I will finish this :)

Siyet (approved):
The overloads look right. I checked them against the runtime constraint validation across the full combination matrix (single constraints, every `gt`/`ge` x `lt`/`le` pairing, numeric mixed with `pattern`/`min_length`/`max_length`/`tz`, and metadata-only) and the type-checker verdict matches `Meta.__init__`'s runtime `ValueError`s exactly, with no false positives or negatives: every combo the runtime rejects is a `call-overload` error, and every runtime-valid combo type-checks. mypy and pyrefly are both clean on `basic_typing_examples.py`; pyright is covered by CI plus the inline ignore markers. Nice that this lifts the runtime checks to type-check time.

One coordination note: #1116 (open) also edits the `from typing_extensions import ...` line (it adds `disjoint_base`), so whichever of these two lands second will need a trivial one-line rebase on that import.

Approving. One scope question, since the title mentions PEP 746: this diff is the type-safe `Meta` constructor overloads only, not the `Annotated`-metadata (`__supertype__`) side of PEP 746. Is that part still planned here, or is this PR now just the constructor overloads?

sobolevn:
Yes, I renamed the PR. Full PEP-746 support is comming later on, I am working on it :)

PR #709, opened 2024-07-12
Fix memory leak in Raw.copy().

tjgq:
Raw_New borrows a reference to its PyBytesObject argument, so Raw_Copy must still dispose of it.

The following is a minimal repro for this issue:

```
import msgspec

DECODER = msgspec.msgpack.Decoder(type=msgspec.Raw)
PAYLOAD = b"\xda\xff\xff" + (65535 * b"\0") # array of 65535 zeros

for _ in range(1000000):
  raw = DECODER.decode(PAYLOAD)
  raw.copy()
```


jcrist:
Thanks!

PR #724, opened 2024-08-22
Do not encode UNSET values in struct arrays.

### LINKED ISSUE #723: JSON encoding fails to skip UNSET values in array_like Struct
### Description

Per the [documentation of UNSET](https://jcristharif.com/msgspec/supported-types.html#unset):

> During encoding, any field containing UNSET is omitted from the message.

Unfortunately, this fails for Structs with `array_like=True`, resulting in the following error:

```plain
TypeError: Encoding objects of type msgspec.UnsetType is unsupported
```

This appears to be an oversight in the function [json_encode_struct_array](https://github.com/jcrist/msgspec/blob/0.18.6/msgspec/_core.c#L14103), which does _not_ check for (and ignore) UNSET values, whereas [json_encode_struct_object _does_ ignore UNSET values](https://github.com/jcrist/msgspec/blob/0.18.6/msgspec/_core.c#L14069), as expected.

I suspect this _might be intentional_ since there can be cases where _decoding_ in the face of UNSET values can present ambiguity, as least in the face of roundtripping, when attributes that can be UNSET are interspersed with those that cannot be UNSET.

However, I would argue that the developer should make judicious use of `array_like=True` to avoid such ambiguity. In other words, it seems reasonable to me that msgspec should indeed skip UNSET values in this case anyway, documenting that this is sort of a "use at your own risk of ambiguity" for the developer, particularly since _decoding_ works without issue.

For example, consider this struct:

```python
class Position(msgspec.Struct, array_like=True, frozen=True, forbid_unknown_fields=True):
    longitude: float
    latitude: float
    altitude: Union[float, msgspec.UnsetType] = msgspec.UNSET
```

Decoding works perfectly well:

```
>>> pos = msgspec.json.decode("[-3.703790, 40.416775]", type=Position)
>>> pos
Position(longitude=-3.70379, latitude=40.416775, altitude=UNSET)
```

Unfortunately, encoding does not:

```
>>> msgspec.json.encode(pos)
...
TypeError: Encoding objects of type msgspec.UnsetType is unsupported
```

Instead of that `TypeError`, I expect encoding to result in `b'[-3.70379, 40.416775]'` since `altitude` is `UNSET`.

When I remove `array_like=True` (or set it explicitly to `False`), decoding (a JSON object in place of a JSON array, _without_ a value for `altitude`) _and_ encoding work as expected.

chuckwondo:
Fixes #723

### DIFF
--- a/tests/test_common.py
+++ b/tests/test_common.py
@@ -4061,6 +4061,17 @@ class TestUnset:
             sol = proto.encode(y)
             assert res == sol
 
+    def test_unset_encode_struct_array_like(self, proto):
+        class Ex(Struct, array_like=True):
+            x: Union[int, UnsetType] = UNSET
+            y: Union[int, UnsetType] = UNSET
+            z: int = 0
+
+        for x, y in [(Ex(), [0]), (Ex(x=1), [1, 0]), (Ex(y=2), [2, 0])]:
+            res = proto.encode(x)
+            sol = proto.encode(y)
+            assert res == sol
+
 
 class TestOrder:
     def test_encoder_order_attribute(self, proto):
--- a/msgspec/_core.c
+++ b/msgspec/_core.c
@@ -12702,17 +12702,36 @@ mpack_encode_struct_array(
     int tagged = tag_value != NULL;
     PyObject *fields = struct_type->struct_encode_fields;
     Py_ssize_t nfields = PyTuple_GET_SIZE(fields);
-    Py_ssize_t len = nfields + tagged;
+    Py_ssize_t len = nfields + tagged, actual_len = len;
 
     if (Py_EnterRecursiveCall(" while serializing an object")) return -1;
 
+    Py_ssize_t header_offset = self->output_len;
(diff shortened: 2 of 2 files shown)


tests/test_common.py
```diff
@@ -4061,6 +4061,17 @@ class Ex(Struct, omit_defaults=True):
...
+        class Ex(Struct, array_like=True):
+            x: Union[int, UnsetType] = UNSET
+            y: Union[int, UnsetType] = UNSET
+            z: int = 0
+
+        for x, y in [(Ex(), [0]), (Ex(x=1), [1, 0]), (Ex(y=2), [2, 0])]:
+            res = proto.encode(x)
+            sol = proto.encode(y)
+            assert res == sol
+
```
CharString: We need another assertion here:

```python
roundtrip = proto.decode(res, type=Ex)
assert roundtrip == res
```

And add a `z` case for good measure.

What good is it to speak, if we will be misunderstood?

PR #760, opened 2024-10-24
MNT: account for CPython 314 changes

tacaswell:
https://github.com/python/cpython/pull/125251 renamed _PY_IMMORTAL_REFCNT -> _PY_IMMORTAL_INITIAL_REFCNT

### DIFF
--- a/msgspec/_core.c
+++ b/msgspec/_core.c
@@ -72,6 +72,12 @@ ms_popcount(uint64_t i) {                            \
 #define MS_UNICODE_EQ(a, b) _PyUnicode_EQ(a, b)
 #endif
 
+#if PY314_PLUS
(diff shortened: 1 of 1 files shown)


tacaswell:
fair enough, but I would go with the new name rather than the old one (I think the new one is clearer).

tacaswell:
(sorry, could not keep my bikeshedding comments to my self).

jcrist:
Fair enough! Out of curiousity - are you using Python dev w/ msgspec for something? How did you find this compatibility issue?

jcrist (approved):

tacaswell:
I have a project that tries to build the development branch of everything (https://github.com/tacaswell/build_the_world) with the goal of catching these things as early as possible.  I then try to work in that environment for my day-to-day development.  I was playing with msgspec a while ago so it ended up in my build list.

ofek:
Really great find, thanks!

PR #767, opened 2024-11-06
remove copyright year in docs

chirizxc:
https://www.sphinx-doc.org/en/master/usage/configuration.html#config-copyright

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -19,7 +19,7 @@ except ImportError:
 
 
 project = "msgspec"
-copyright = "%Y, Jim Crist-Harif"
+copyright = "Jim Crist-Harif"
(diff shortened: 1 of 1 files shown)


docs/source/conf.py
```diff
@@ -19,7 +19,7 @@ class UnsetType:
 
 
 project = "msgspec"
-copyright = "2022, Jim Crist-Harif"
+copyright = "%Y, Jim Crist-Harif"
```
jcrist, line 22: You don't need a copyright year, and if anythiing it should be the year first published, not the current year. Better off to just remove the year entirely.

```suggestion
copyright = "Jim Crist-Harif"
```

jcrist:
Thanks!

PR #773, opened 2024-11-16
Use eval_type_backport on Python 3.9 if it's installed to resolve `int | None` etc.

### LINKED ISSUE #771: Backport type annotation syntax support for older python versions
Pydantic handles this via https://github.com/alexmojaki/eval_type_backport -- is there any appetite for doing the same in msgspec? Happy to PR it if that it'd be supported.

_Originally posted by @ashb in https://github.com/jcrist/msgspec/issues/737#issuecomment-2470956504_

ashb:
This uses the same module that pydantic does, and it allows people to use the
new pipe syntax if they have to support Python3.9 too -- very useful for
libraries.

(Also it works better with many type checkers which seem to mistakenly think
that with `from __future__ import annotations` means `int| None` will work,
but it doesn't out of the box.)

Fixes #771

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -78,7 +78,7 @@ jobs:
         os: [ubuntu-latest, macos-13, windows-latest]
 
     env:
-      CIBW_TEST_REQUIRES: "pytest msgpack pyyaml tomli tomli_w"
+      CIBW_TEST_EXTRAS: "test"
       CIBW_TEST_COMMAND: "pytest {project}/tests"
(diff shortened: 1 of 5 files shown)


ashb:
@jcrist PTAL. I've put in a place for making it installed or not, but didn't do anything in this PR to run it with or without this.

Let me know if you'd like the structure something else too. Or I'm more than happy for you to push changes to my fork too.

ashb:
(I thought about adding the test for this in test_common with a decoder, but this seemed to be more directly testing the code)

jcrist:
Thanks for the PR (and your patience here), this is in!

PR #780, opened 2024-12-05
Cover conversion to/from builtins in usage docs

### LINKED ISSUE #778: `asdict` should respect `omit_defaults`
### Description

(I'm not sure if this is a bug report or a feature request, so I opted for a feature request in case the current behaviour is intentional)

I am using `msgspec` to help communicate with a server where a boolean field can be set to `None` in the data model not because `null` is a valid value, but because it is permitted to omit the key entirely. The relevant struct has `omit_defaults=True` set accordingly.

However, `msgspec` isn't handling the full message serialisation (`httpx` is doing that), so I attempted to use `asdict` to generate the message spec portion of the message.

This failed, as `asdict` includes all fields, without paying attention to `omit_defaults`.

ncoghlan:
Also clarify differences between `to_builtins` and `asdict`.

Closes #778


ncoghlan:
Likely worth linking to https://github.com/jcrist/msgspec/blob/main/docs/source/converters.rst from the new subsection, since that goes into more detail on how `to_builtins` and `convert` relate to each other.

PR #785, opened 2024-12-19
chore(build): updated Github CI Actions

cofin:
@jcrist, in my recent testing with Python 3.13, msgspec, and Litestar, I found myself needing binary wheels to speed things up.

There were a few GH actions that have implemented breaking changes, and this PR has the minor updates required to publish releases.


jcrist (approved):
Thanks!

PR #796, opened 2025-01-07
Add forbid_unknown_fields as optional bool argument to all decoders.

eamartin:
When forbid_unknown_fields is True, then errors are raised for unknown
fields on all Structs, dataclasses, and TypedDicts.

When forbid_unknown_fields is False, then errors for unknown fields
are only raised on Structs defined with forbid_unknown_fields=True.


dhirschfeld:
This would be very useful!

I'm just now running into a situation where I want to ignore unknown fields when calling `msgspec.convert` to create a `Struct` which has been defined with `forbid_unknown_fields=True`.

return42:
@jcrist do you see a chance, we get this useful feature merged into one of the next releases?

update: sorry for the noise / to late I have seen, we already have [`forbid_unknown_fields=True`](https://jcristharif.com/msgspec/structs.html#forbidding-unknown-fields) in the `msgspec.Struct`.

PR #810, opened 2025-02-04
Python 3.14: Call __annotate__ on type objects to get annotations

hroncok:
Fixes https://github.com/jcrist/msgspec/issues/795

### DIFF
--- a/msgspec/_utils.py
+++ b/msgspec/_utils.py
@@ -149,7 +149,12 @@ def get_class_annotations(obj):
         cls_locals = dict(vars(cls))
         cls_globals = getattr(sys.modules.get(cls.__module__, None), "__dict__", {})
 
-        ann = cls.__dict__.get("__annotations__", {})
+        if "__annotations__" in cls.__dict__:
+            ann = cls.__dict__["__annotations__"]
+        elif "__annotate__" in cls.__dict__:
+            ann = cls.__dict__["__annotate__"](1)  # annotationlib.Format.VALUE
+        else:
+            ann = {}
         for name, value in ann.items():
             if name in hints:
                 continue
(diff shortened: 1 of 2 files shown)


edgarrmondragon:
This fix works on 3.14.0a6 👍

edgarrmondragon:
> This fix works on 3.14.0a6 👍

But it doesn't on 3.14.0b1 😅

hroncok:
I was thinking that maybe the solution is to reverse the order (aka look into `__annotate__` first), but it makes no difference.

hroncok:
@JelleZijlstra sorry for bothering you directly, but do you by chance know how to fix this code? I mostly contributed this by trial and error and it no longer works with 3.14.0b1. Thanks.

msgspec/_utils.py
```diff
@@ -149,7 +149,12 @@ def get_class_annotations(obj):
         cls_locals = dict(vars(cls))
         cls_globals = getattr(sys.modules.get(cls.__module__, None), "__dict__", {})
 
-        ann = cls.__dict__.get("__annotations__", {})
```
JelleZijlstra, line 152: This should just use `inspect.get_annotations(cls)` which will work on all versions.

You may want to use the FORWARDREF format on 3.14 to support classes with annotations that cannot be fully evaluated. I don't know if that's a design goal for this function.

JelleZijlstra:
Retrieving `__annotations__` and `__annotate__` directly from a class namespace is not supported; use https://docs.python.org/3.14/library/annotationlib.html#annotationlib.get_annotate_from_class_namespace instead.

I realize the code where this happens is in C so what I say might not be straightforward though. Would it be OK for the C code to import `annotationlib`?

PR #813, opened 2025-02-12
Add `__struct_encode_fields__` to type stub

ncoghlan:
Static type checkers currently complain if querying the encoded field names instead of the attribute names.


ncoghlan:
Should this also be added to the typing test cases in https://github.com/jcrist/msgspec/blob/dd965dce22e5278d4935bea923441ecde31b5325/tests/basic_typing_examples.py#L330 ?

sobolevn:
`__struct_encode_fields__` was already present, I added the test for it.
@ncoghlan welcome to the contributors of `msgspec`, nice to see you! 👋

sobolevn (approved):

PR #816, opened 2025-02-17
Add bool dunder method for type def of UnsetType

mZbZ:
Currently instances of UnsetType are falsy as would be expected but the type definition doesn't reflect this.

Tiny change so that I and others could stop seeing red squiggly lines


jcrist:
Superseded by #975. Closing.

PR #818, opened 2025-02-18
Allow Fields to be subclassable

Vizonex:
I bumped into a problem in my project in [SQLTable](https://github.com/Vizonex/SQLTable) which was meant to use this library to help sterilize SQLAlchemy objects to JSON with the fastest performance possible as well as the ability of having a faster init function but then I faced a problem where the Fields could not be subclassed allowing for the use of mapped-columns to be utilized, everything else that I have tried to force it to work turned up null. The SQLAlchemy Team Told me their thoughts and opinions on what my project should change in order to be [better than SQLModel](https://github.com/sqlalchemy/sqlalchemy/discussions/12278#discussioncomment-11972815) which I suggest looking into if this concept interests any of you. 

I've tried other alternative ways such as creating a metaclass similar to this library's version but in Cython to make the code smaller but it seems that [Cython is not ready for that yet](https://github.com/cython/cython/issues/6672) so this was my backup plan of mine to make SQLAlchemy run faster. 

This Idea of subclassing fields would finally let me bind mapped_column variables to msgspec fields allowing me to achieve my goals of having a faster SQL-Library than SQLModel which suffers from not being maintained all that often and with relationship values not sterilizing with the other data frustrated me. 

Incase to test and make sure that my idea works I added another test to the tests folder to make sure subclassing works correctly. I do see an issue however with Field-Inspections being a problem in the future for subclassing however I don't think it's something that we should worry about at the moment and believe me when I say that I've tried everything else before pulling this final straw together.

I mainly have interests in binding my designs to aiohttp but I could see projects such as litestar start taking interest to my brand new approach.


Vizonex:
You know what this is not needed anymore so I will drop this pr and refork what I have since I bumped into a new problem that requires fixing...

PR #824, opened 2025-03-15
Support custom enum metaclasses

### LINKED ISSUE #822: Struct that contains an Enum property with a metaclass cant be decoded
### Description

I don't know if this is a bug, or just something that can't be supported. I've reviewed some of the other issues surrounding metaclasses and it seems there isn't a workaround.

This case is a little different though, my classes which inherit from Struct do not define a metaclass - but instead they have a property which is a subclass of `Enum` with a metaclass.

I put together a minimal reproducible example. Is it possible to support this use case?

```py
from enum import Enum, EnumMeta

import msgspec


class BaseEnumSuccess(Enum):
    ...


class BaseEnumMeta(EnumMeta):
    ...


class BaseEnumFail(Enum, metaclass=BaseEnumMeta):
    ...


class EnumSuccess(BaseEnumSuccess):
    Success = "success"


class EnumFail(BaseEnumFail):
    Success = "success"


class TesterSuccess(msgspec.Struct):
    variant: EnumSuccess


class TesterFail(msgspec.Struct):
    variant: EnumFail


decoder_success = msgspec.json.Decoder(TesterSuccess)
decoded_success = decoder_success.decode(b'{"variant": "success"}')
print(decoded_success)

decoder_fail = msgspec.json.Decoder(TesterFail)
decoded_fail = decoder_fail.decode(b'{"variant": "success"}')
print(decoded_fail)
```

Output:

```
TesterSuccess(variant=<EnumSuccess.Success: 'success'>)
  File "/home/jonx/projects/py/example.py", line 39, in <module>
    decoded_fail = decoder_fail.decode(b'{"variant": "success"}')
msgspec.ValidationError: Expected `EnumFail`, got `str` - at `$.variant`
```

Thanks in advance for your time, love the library - it massively improved performance for me.

UnknownPlatypus:
Fixes #822 (cc @Jonxslays)

This will have the side effect of easing adoption for Django projects because they tend to use [TextChoices / IntChoices](https://docs.djangoproject.com/en/5.1/ref/models/fields/#enumeration-types) when they need enumerations and this makes them work.

This was my last blocker to adopt msgspec officially at work!

### DIFF
--- a/msgspec/_core.c
+++ b/msgspec/_core.c
@@ -4867,7 +4867,7 @@ typenode_collect_type(TypeNodeCollectState *state, PyObject *obj) {
     ) {
         out = typenode_collect_struct(state, t);
     }
-    else if (Py_TYPE(t) == state->mod->EnumMetaType) {
+    else if (PyType_IsSubtype(Py_TYPE(t), state->mod->EnumMetaType)) {
         out = typenode_collect_enum(state, t);
     }
     else if (origin == (PyObject*)(&PyDict_Type)) {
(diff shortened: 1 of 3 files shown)


TheRealSeber:
hey, @jcrist @ofek any updates on this?

PR #842, opened 2025-04-29
Use PyObject_HEAD_INIT for _Unset_Object and _NoDefault_Object on >=3.13

alexmalyshev:
This makes the immortal refcount an implementation detail and it accounts for GIL vs NoGIL builds of Python (ob_refcnt vs ob_ref_local).


alexmalyshev:
Whoops, this is wrong. The macro only immortalizes when Py_BUILD_CORE is defined.

Issue #841, opened 2025-04-28
Inheritable `kw_only` for `Struct` (e.g., `kw_always=True`)


nightsailer:
### Description

Currently, `kw_only=True` on `msgspec.Struct` does not inherit to subclasses.  
I'm building a `pydantic`-like wrapper over `msgspec`, and I want users to define models without repeating `kw_only=True`:

```python
class BaseModel(Struct, kw_only=True):
    ...

class User(BaseModel):
    age: int = Field(...)
    name: str

Today, users must write:

class User(BaseModel, kw_only=True):
    age: int = Field(...)
    name: str

```

which hurts usability.

⸻

### Proposal

Add an option like kw_always=True that enables kw_only automatically for subclasses, without needing to re-specify it.

⸻

Why it’s needed
	•	Easier migration for pydantic users
	•	Cleaner API
	•	Avoids complex workarounds (like dynamically generating inner structs)

⸻

Current workaround

I tried subclassing StructMeta but failed.
My only working solution is generating a dynamic InnerStruct at runtime, which complicates type handling, nested structures, and serialization.

⸻

Impact

Supporting kw_always=True would significantly improve framework and library development on top of msgspec.

Thanks for considering!

kent-foodini:
I'd love that option as well. Great library - thanks for the great project.

ofek:
Resolved by https://github.com/jcrist/msgspec/pull/890

Issue #837, opened 2025-04-23
Expose metaclass for `msgspec.Struct`


apoorvkh:
### Description

Hey! I was wondering if it would be possible to expose the metaclass for `Struct` (`msgspec._core.StructMeta`) in Python?

I think that would in general permit adding metaclasses to classes that inherit `Struct`. And could possibly be used to enable support for #806?

Thanks!

nightsailer:
check msgspec-x:

https://github.com/nightsailer/msgspec-x

ofek:
Resolved by https://github.com/jcrist/msgspec/pull/890

Issue #820, opened 2025-03-03
Msgspec Doesn't Decode Json Correctly when SQLAlchemy.Mapped is used.


Vizonex:
### Description

I am having a problem where when I want to write a Json Decoder form SQLAlchemy from a base class I bump into an issue where msgspec doesn't know how to handle these types.

```python
from sqlalchemy.orm import Mapped, MappedAsDataclass, DeclarativeBaseNoMeta, mapped_column, declared_attr
from msgspec.json import Decoder, Encoder


class SQLTable(MappedAsDataclass, DeclarativeBaseNoMeta):
    pass


class User(SQLTable):
    __tablename__ = "user"
    # Mapped is used to register columns in the sql table without 
    # Having to write mapped_column out many times.
    name:Mapped[str]
    id:Mapped[int] = mapped_column(primary_key=True, default=None)

encoder = Encoder()
# Encoder works just fine...
print(encoder.encode(User("Vizonex", 10)))


# But the decoder on the other-hand is broken and I expect to decode the User json 
# To a User Object...
# msgspec.ValidationError: Expected `Mapped`, got `str` - at `$.name`
user = Decoder(User).decode(b'{"name":"Vizonex", "id": 10}')
```

I am completely stumped on how to fix this problem and I have gone through all hoops and hurtles to try and make the decoder for my upcoming python library `SQLTable` Decode json correctly. 

## Other Things I've Tried
I have tried hacking the `Mapped[]` Object itself into an annotated object but then SQLAlchemy fails to find it...
```python
from sqlalchemy.orm import Mapped, MappedAsDataclass, DeclarativeBaseNoMeta, mapped_column, declared_attr
from typing import Annotated, TypeVar
from msgspec.json import Decoder, Encoder

T = TypeVar("T")

# this was an idea I came up with to allow msgspec to act friendly with 
# Sqlalchemy Mapped variables but sqlalchmey doesn't act friend with 
# finding Mapped attributes...
HackedMap = Annotated[T, Mapped[T]]

class SQLTable(MappedAsDataclass, DeclarativeBaseNoMeta):
    pass

class User(SQLTable):
    __tablename__ = "user"
    # We can try to hack it to how we please so that msgspec's decoder works correctly 
    # but then a new problem arises     
    name:HackedMap[str]
    id:HackedMap[int] = mapped_column(primary_key=True, default=None)



# It only shows {'id': <ColumnProperty at 0x20d09b14e40; id>} Because we set it with mapped_column
[...]

PR #843, opened 2025-05-03
Expose and Enable Inheritance of StructMeta Metaclass

nightsailer:
# Expose and Enable Inheritance of StructMeta Metaclass

## Changes
- Exposed `StructMeta` metaclass to Python for direct use
- Added `Py_TPFLAGS_BASETYPE` flag to enable Python inheritance of `StructMeta`
- Created tests to verify functionality

## Why
This allows users to:
- Use `StructMeta` directly without subclassing `Struct`
- Create custom metaclasses that extend `StructMeta` behavior
- Implement reusable patterns across multiple struct classes

This PR resolves:
- Issue #841 
- Issue  #837 

## Example

```python
from msgspec import StructMeta

# Custom metaclass with kw_only_default parameter
class KwOnlyMeta(StructMeta):
    _kw_only_default_settings = {}
    
    def __new__(mcls, name, bases, namespace, **kwargs):
        # Store and inherit kw_only_default
        kw_only_default = kwargs.pop('kw_only_default', None)
        if kw_only_default is None:
            for base in bases:
                if base.__name__ in mcls._kw_only_default_settings:
                    kw_only_default = mcls._kw_only_default_settings[base.__name__]
                    break
        else:
            mcls._kw_only_default_settings[name] = kw_only_default
            
        # Apply kw_only_default if kw_only not specified
        if 'kw_only' not in kwargs and kw_only_default is not None:
            kwargs['kw_only'] = kw_only_default
            
        return super().__new__(mcls, name, bases, namespace, **kwargs)

# Base model requiring keyword arguments
class BaseModel(metaclass=KwOnlyMeta, kw_only_default=True):
    pass

# Child class inherits kw_only=True
class User(BaseModel):
    name: str
    age: int

# Must use keyword arguments
user = User(name="John", age=30)
```

## Technical Details
Added StructMeta to module exports in _core.c
Added Py_TPFLAGS_BASETYPE flag to StructMetaType
Updated __init__.py to import and expose StructMeta
Added type hints in __init__.pyi

## Compatibility
These changes are backward compatible and only add new capabilities without modifying existing behavior.

## Related Issues

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -1,6 +1,6 @@
 # How to Contribute
 
-Thank you for taking the time to contribute to `msgspec`!
+Thank you for taking the time to contribute to `msgspec-x`!
 
 Here we document some contribution guidelines to help you ensure that your
 contribution is at its best.
@@ -17,17 +17,17 @@ Once you have those installed, you're ready to:
 
 - Clone the repository
 - Install all development dependencies
-- Build a development version of `msgspec`
+- Build a development version of `msgspec-x`
 - Install the `pre-commit` hooks
 
 ```bash
 # Clone the repository
-git clone https://github.com/jcrist/msgspec.git
+git clone https://github.com/nightsailer/msgspec-x.git
 
 # cd into the repo root directory
-cd msgspec/
+cd msgspec-x/
 
-# Build and install msgspec & all dev dependencies
+# Build and install msgspec-x & all dev dependencies
 pip install -e ".[dev]"
 
 # Install the pre-commit hooks
@@ -36,9 +36,9 @@ pre-commit install
 
 ## Editing and Rebuilding
 
-You now have a "development" build of `msgspec` installed. This means that you
+You now have a "development" build of `msgspec-x` installed. This means that you
 can make changes to the `.py` files and test them without requiring a rebuild
-of msgspec's C extension. Edit away!
+of msgspec-x's C extension. Edit away!
 
 If you do make changes to a `.c` file, you'll need to recompile. You can do
 this by running
@@ -47,7 +47,7 @@ this by running
 pip install -e .
 ```
 
-By default `msgspec` is built in release mode, with optimizations enabled. To
+By default `msgspec-x` is built in release mode, with optimizations enabled. To
 build a debug build instead (for use with e.g. `gdb` or `lldb`) define the
 `MSGSPEC_DEBUG` environment variable before building.
 
@@ -112,6 +112,6 @@ and fix any issues that come up.
 
 ## Code of Conduct
 
-``msgspec`` has a code of conduct that must be followed by all contributors to
+``msgspec-x`` has a code of conduct that must be followed by all contributors to
 the project. You may read the code of conduct
-[here](https://github.com/jcrist/msgspec/blob/main/CODE_OF_CONDUCT.md).
+[here](https://github.com/nightsailer/msgspec-x/blob/main/CODE_OF_CONDUCT.md).
--- a/.github/ISSUE_TEMPLATE/bug.yml
+++ b/.github/ISSUE_TEMPLATE/bug.yml
@@ -1,5 +1,5 @@
 name: 🪲 Bug Report
-description: Report a bug or unexpected behavior in msgspec
+description: Report a bug or unexpected behavior in msgspec-x
 body:
   - type: markdown
     attributes:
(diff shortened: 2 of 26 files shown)


Vizonex:
@nightsailer I sure hope this passes you don't know how annoying it has been to try and turn Msgspec into an Object Relational Mapping Library. If this passes it would likely remedy my recent problems https://github.com/jcrist/msgspec/issues/820 and possibly allow StructMeta and any SQLAlchemy Type Metaclass object to be theoretically mergeable similar to how SQLModel works.

Vizonex:
Also I ran your test on my computer and it is working correctly.

apoorvkh:
Ah, hey @nightsailer, seems like we were working in parallel! See my PR #844. [You were first, I did not re-check for open PRs :)]

Let's collaborate a bit? Our changes are quite similar, except that
- you also updated the equalities `(Py_TYPE(type) == &StructMetaType)` to `(PyType_IsSubtype(Py_TYPE(type), &StructMetaType))`
- You added `metaclass=StructMeta` to the `Struct` class in `__init__.pyi`
- I added a stub for `StructMeta` in `__init__.pyi`

In my version, I wasn't able to get the following test working:
```python
from msgspec import Struct, StructMeta

class FrozenStructMeta(StructMeta):
    def __new__(mcls, *args, **kwargs):
        kwargs["frozen"] = True
        return super().__new__(mcls, *args, **kwargs)

class ParentStruct(Struct, metaclass=FrozenStructMeta):
    a: int

class ChildStruct(ParentStruct):
    b: int

print(ParentStruct(a=1))
print(ChildStruct(a=1, b=2))
```

Yes, the metaclass inheritance works so these classes are indeed both immutable. However, `ChildStruct` only expects the argument `b: int` and does not inherit arguments from the base classes. That is different behavior compared to `msgspec.Struct`.

```
print(ChildStruct(a=1, b=2))
>> TypeError: Unexpected keyword argument 'a'
```

Does this work in your version?

apoorvkh:
Just checked, the above test does work in your fork! The causal factor is the `PyType_IsSubtype` equality check which you've included.

May I make one commit to your PR to update the stub for `StructMeta` to [this one](https://github.com/jcrist/msgspec/pull/844/files#diff-5f5f76b188c455c8647945f8371bdda146c75f62c4db83bc596fe1ec44a9a5ecR76-R112)?

apoorvkh:
Ok that makes sense, thanks for clarifying!

Please merge this PR for the change we discussed: https://github.com/nightsailer/msgspec/pull/1

makarr:
> @nightsailer I sure hope this passes you don't know how annoying it has been to try and turn Msgspec into an Object Relational Mapping Library. If this passes it would likely remedy my recent problems #820 and possibly allow StructMeta and any SQLAlchemy Type Metaclass object to be theoretically mergeable similar to how SQLModel works.

Seconded. I am working on an ORM and want to set msgspec.UNSET to everything by default in the metaclass, so that I can have quick and dirty DTOs on the cheap. But currently it is verbose and annoying having to manually annotate every model with `... | UnsetType = UNSET`. Class decorators don't work either, since __struct_defaults__ is readonly.

apoorvkh:
@jcrist I think there's a [large amount of interest](https://github.com/jcrist/msgspec/pulls?q=is%3Apr+sort%3Areactions-%2B1-desc+) in this PR! It would be really wonderful if you could take a look and consider merging. Thanks so much.

Vizonex:
> @jcrist I think there's a [large amount of interest](https://github.com/jcrist/msgspec/pulls?q=is%3Apr+sort%3Areactions-%2B1-desc+) in this PR! It would be really wonderful if you could take a look and consider merging. Thanks so much.

I'm interested because it means I would be able to add msgspec to a few libraries such as a few that I currently maintain such as aiocallback and a few others.

nightsailer:
I have fork a new package: msgspec-x,  this pr was  conflict, close now.

apoorvkh:
Hey @nightsailer, while you've made a new package with this feature, I feel like the existing package has many pre-existing users and also deserves to have this feature. Would you consider re-opening this PR? Thanks.

jacopoabramo:
@nightsailer given that there seems to be a transition to new mantainers, could you open a new PR with the original changes so that these can be merged into msgspec?

ofek:
I'd be eager to merge this if you or someone else wouldn't mind opening another PR 😄

apoorvkh:
That would be great, thanks a lot for taking over maintenance for this project @ofek!

Since I was involved in this effort (per #837, #843), I'm happy to re-open the PR.

Please see #890 for that! This corresponds to the current PR, prior to the `msgspec-x` rebranding.

PR #846, opened 2025-05-12
Simplify JSON schema of Optionals

natehardison:
I am using msgspec to generate JSON schemas for MCP tools, and MCP clients don't recognize the descriptions of optional fields when they are embedded in an `anyOf`. By special-casing the `Optional[...]` case of `Union` when generating a JSON schema to not use `anyOf`, MCP clients are able to recognize the "description" as well as the optionality of the field.


rafalkrupinski:
With this change, can I still serialize `None` as `null`?

CharString:
:-1: This PR generates invalid schemas. If downstream tools need special care, transform valid schemas to their need in post processing.

PR #854, opened 2025-05-26
Relax all getrefcount tests to allow lower numbers

hroncok:
This is related to https://docs.python.org/3.14/whatsnew/3.14.html#whatsnew314-refcount

Created via:

    sed -Ei 's/sys\.getrefcount\(([^\)]+)\) == ([0-9]+)/sys.getrefcount(\1) <= \2/' tests/test_*.py

Perhaps this is too strong a hammer, but it makes the tests pass on 3.14.


ofek (approved):
I can't think of a much better way than this, thanks!

PR #865, opened 2025-08-07
chore(docs): add faststream in used by

Sehat1137:
He, we in FastStream have added support for msgspec and would be very grateful if you could merge this small PR to include a reference in your documentation.

This will increase user awareness about of our libraries.
Thank you for considering this! Let us know if any changes are needed.

The guide is available here: https://faststream.ag2.ai/0.6/getting-started/subscription/msgspec/


Lancetnik:
Hi @ofek ! I think we can do it. Don't see any doubts. I introduced msgspec support in the latest release, so had no chance to update the doc accordingly

Lancetnik:
@ofek added

PR #877, opened 2025-10-14
add free-threading support 

kumaraditya303:
This PR fixes the thread safety issues in msgspec implementation and add free-threading support as well as making it compatible with 3.14. Python 3.13t support is omitted as it is experimental and has many thread safety issues which won't be fixed. 

For fixing compatibility with 3.14, many of the reference counting tests needed be adjusted to account for interpreter now using borrowed reference. 

This PR includes https://github.com/jcrist/msgspec/pull/852/ should that should be merged first.

edit: resolves https://github.com/jcrist/msgspec/issues/878

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -81,7 +81,7 @@ jobs:
     env:
       CIBW_TEST_EXTRAS: "test"
       CIBW_TEST_COMMAND: "pytest {project}/tests"
-      CIBW_BUILD: "cp39-* cp310-* cp311-* cp312-* cp313-* cp314-*"
+      CIBW_BUILD: "cp39-* cp310-* cp311-* cp312-* cp313-* cp314-* cp314t-*"
       CIBW_SKIP: "*-win32 *_i686 *_s390x *_ppc64le"
       CIBW_ARCHS_MACOS: "x86_64 arm64"
       CIBW_ARCHS_LINUX: "x86_64 aarch64"
@@ -100,7 +100,7 @@ jobs:
       - name: Set up Environment
         if: github.event_name != 'release'
         run: |
-            echo "CIBW_SKIP=${CIBW_SKIP} *-musllinux_* cp39-*_aarch64 cp311-*_aarch64 cp312-*_aarch64 cp313-*_aarch64 cp314-*_aarch64" >> $GITHUB_ENV
+            echo "CIBW_SKIP=${CIBW_SKIP} *-musllinux_* cp39-*_aarch64 cp311-*_aarch64 cp312-*_aarch64 cp313-*_aarch64 cp314-*_aarch64 cp314t-*_aarch64" >> $GITHUB_ENV
 
       - name: Build & Test Wheels
         uses: pypa/cibuildwheel@v3.2.1


kumaraditya303:
@ofek Should I add 3.14t CI in this PR or would you add it in https://github.com/jcrist/msgspec/pull/888?

ofek:
I just merged that PR, feel free to rebase and add support for 3.14t!

kumaraditya303:
> I just merged that PR, feel free to rebase and add support for 3.14t!
 
done!

kumaraditya303:
All green now with 3.14t

msgspec/_core.c
```diff
@@ -13742,9 +13881,13 @@ json_encode_sequence(EncoderState *self, Py_ssize_t size, PyObject **arr)
 static MS_NOINLINE int
 json_encode_list(EncoderState *self, PyObject *obj)
 {
-    return json_encode_sequence(
+    int ret;
+    Py_BEGIN_CRITICAL_SECTION(obj);
```
ofek: For some of these sections, what do you think about the following advice from AI? This first part came from just having the diff, then I provided more context.

---

You hold `Py_BEGIN_CRITICAL_SECTION(...)` while encoding values (e.g., `json_encode_object`, `mpack_encode_object`, `to_builtins_object`, `*_dict`, `*_list`). Those encodes recurse, allocate, and can re-enter the same container or other containers, producing deadlocks in a GIL-less runtime. Critical sections must be short, non-allocating, and never call back into Python.

**Fix pattern (snapshot, then encode):**

```c
// dict
Py_ssize_t n = PyDict_Size(d);
PyObject **keys = PyMem_Malloc(n * sizeof(*keys));
PyObject **vals = PyMem_Malloc(n * sizeof(*vals));
if (!keys || !vals) { PyErr_NoMemory(); goto error; }

Py_ssize_t pos = 0, i = 0;
PyObject *k, *v;
Py_BEGIN_CRITICAL_SECTION(d);
while (PyDict_Next(d, &pos, &k, &v)) {
    Py_INCREF(k); Py_INCREF(v);
    keys[i] = k;  vals[i] = v;
    i++;
}
Py_END_CRITICAL_SECTION();

// Now encode outside the lock
for (Py_ssize_t j = 0; j < i; j++) {
    /* encode keys[j], vals[j] */
    Py_DECREF(keys[j]); Py_DECREF(vals[j]);
}
PyMem_Free(keys); PyMem_Free(vals);

```

```c
// list
Py_ssize_t n = PyList_GET_SIZE(lst);
PyObject **items = PyMem_Malloc(n * sizeof(*items));
if (!items) { PyErr_NoMemory(); goto error; }

Py_BEGIN_CRITICAL_SECTION(lst);
for (Py_ssize_t i = 0; i < n; i++) {
    PyObject *it = PyList_GET_ITEM(lst, i);
    Py_INCREF(it);
    items[i] = it;
}
Py_END_CRITICAL_SECTION();

// Encode outside the lock...
```

Apply this across:

- `mpack_encode_list`, `mpack_encode_dict`, `mpack_encode_object`
- `json_encode_list`, `json_encode_dict`, `json_encode_object`
- `AssocList_FromDict`, `AssocList_FromObject`
- `to_builtins_dict`, `to_builtins_object`
- any `convert_*` that currently wraps the whole function in a critical section

---

I then provided the contents of the entire `msgspec/_core.c` and asked for specifics.

---

<details>

Pattern: hold the critical section only long enough to copy pointers and INCREF; release; then do conversion/encoding.

1. Lists → JSON

    Current holds the list while encoding every element.

    Change to:

```c
static MS_NOINLINE int
json_encode_list(EncoderState *self, PyObject *obj) {
    Py_ssize_t n;
    PyObject **items = NULL;
    int st = -1;

    Py_BEGIN_CRITICAL_SECTION(obj);
    n = PyList_GET_SIZE(obj);
    items = PyMem_Malloc(sizeof(PyObject*) * n);
    if (!items) { Py_END_CRITICAL_SECTION(); return -1; }
    for (Py_ssize_t i = 0; i < n; i++) {
        items[i] = PyList_GET_ITEM(obj, i);
        Py_INCREF(items[i]);
    }
    Py_END_CRITICAL_SECTION();

    st = json_encode_sequence(self, n, items);

    for (Py_ssize_t i = 0; i < n; i++) Py_DECREF(items[i]);
    PyMem_Free(items);
    return st;
}
```

2. Dicts → MessagePack

    Current encodes keys/values while holding the dict.

    Change to:

```c
static MS_NOINLINE int
mpack_encode_dict(EncoderState *self, PyObject *obj) {
    Py_ssize_t len = PyDict_GET_SIZE(obj);
    if (len == 0) { char h[1] = {MP_FIXMAP}; return ms_write(self, h, 1); }
    if (self->order != ORDER_DEFAULT) return mpack_encode_and_free_assoclist(self, AssocList_FromDict(obj));

    int st = -1;
    PyObject **keys = NULL, **vals = NULL;
    Py_ssize_t pos = 0; PyObject *k, *v; Py_ssize_t i = 0;

    keys = PyMem_Malloc(sizeof(PyObject*) * len);
    vals = PyMem_Malloc(sizeof(PyObject*) * len);
    if (!keys || !vals) goto done;

    Py_BEGIN_CRITICAL_SECTION(obj);
    while (PyDict_Next(obj, &pos, &k, &v)) {
        Py_INCREF(k); Py_INCREF(v);
        keys[i] = k; vals[i] = v; i++;
    }
    Py_END_CRITICAL_SECTION();

    if (mpack_encode_map_header(self, i, "dicts") < 0) goto done;
    if (Py_EnterRecursiveCall(" while serializing an object")) goto done;
    for (Py_ssize_t j = 0; j < i; j++) {
        if (mpack_encode_dict_key_inline(self, keys[j]) < 0) goto leave;
        if (mpack_encode_inline(self, vals[j]) < 0) goto leave;
    }
    st = 0;
leave:
    Py_LeaveRecursiveCall();
done:
    if (keys) { for (Py_ssize_t j = 0; j < i; j++) Py_DECREF(keys[j]); PyMem_Free(keys); }
    if (vals) { for (Py_ssize_t j = 0; j < i; j++) Py_DECREF(vals[j]); PyMem_Free(vals); }
    return st;
}
```

3. Dicts → JSON

    Same idea as (2). Current code holds the dict while calling `json_encode_*`.

    Apply the same snapshot of `keys[]/vals[]`, then:

```c
if (ms_write(self, "{", 1) < 0) goto done;
if (Py_EnterRecursiveCall(" while serializing an object")) goto done;
for (Py_ssize_t j = 0; j < i; j++) {
    if (json_encode_dict_key(self, keys[j]) < 0) goto leave;
    if (ms_write(self, ":", 1) < 0) goto leave;
    if (json_encode(self, vals[j]) < 0) goto leave;
    if (ms_write(self, ",", 1) < 0) goto leave;
}
*(self->output_buffer_raw + self->output_len - 1) = '}';
```

4. Objects (__dict__ + slots) → MessagePack

    Current encodes under object and dict critical sections.

    Change to: take two snapshots while locked, then encode unlocked.

```c
// before: after header
Py_ssize_t cap = max_size, n = 0;
PyObject **karr = PyMem_Malloc(sizeof(PyObject*) * cap);
PyObject **varr = PyMem_Malloc(sizeof(PyObject*) * cap);
if (!karr || !varr) goto cleanup;

Py_BEGIN_CRITICAL_SECTION(obj);
// __dict__
if (dict) {
    PyObject *k, *v; Py_ssize_t pos = 0;
    while (PyDict_Next(dict, &pos, &k, &v)) {
        if (!PyUnicode_CheckExact(k) || v == UNSET) continue;
        Py_INCREF(k); Py_INCREF(v);
        karr[n] = k; varr[n] = v; n++;
    }
}
// slots
PyTypeObject *type = Py_TYPE(obj);
while (type) {
    PyMemberDef *mp = MS_PyHeapType_GET_MEMBERS((PyHeapTypeObject*)type);
    for (Py_ssize_t i = 0; i < Py_SIZE(type); i++, mp++) {
        if (mp->type == T_OBJECT_EX && !(mp->flags & READONLY)) {
            PyObject *v = *(PyObject **)((char*)obj + mp->offset);
            if (v && v != UNSET && *mp->name != '_') {
                PyObject *k = PyUnicode_FromString(mp->name);
                if (!k) { /* leave locked? */ ; }
                Py_INCREF(v);
                karr[n] = k; varr[n] = v; n++;
            }
        }
    }
    type = type->tp_base;
}
Py_END_CRITICAL_SECTION();

// now encode unlocked
for (Py_ssize_t i = 0; i < n; i++) {
    Py_ssize_t key_len; const char *kb = unicode_str_and_size(karr[i], &key_len);
    if (!kb) goto cleanup2;
    if (mpack_encode_cstr(self, kb, key_len) < 0) goto cleanup2;
    if (mpack_encode(self, varr[i]) < 0) goto cleanup2;
    size++;
}
```

5. `to_builtins_dict`

    Current calls `to_builtins` while holding the dict.

    Take a snapshot first:

```c
PyObject **keys = NULL, **vals = NULL; Py_ssize_t i = 0, len = PyDict_GET_SIZE(obj);
keys = PyMem_Malloc(sizeof(PyObject*) * len);
vals = PyMem_Malloc(sizeof(PyObject*) * len);

Py_BEGIN_CRITICAL_SECTION(obj);
Py_ssize_t pos = 0; PyObject *k, *v;
while (PyDict_Next(obj, &pos, &k, &v)) {
    Py_INCREF(k); Py_INCREF(v);
    keys[i] = k; vals[i] = v; i++;
}
Py_END_CRITICAL_SECTION();

for (Py_ssize_t j = 0; j < i; j++) {
    PyObject *new_key = to_builtins(self, keys[j], true);
    PyObject *new_val = to_builtins(self, vals[j], false);
    // ... existing key normalization and PyDict_SetItem as today ...
}
```

6. `convert_dict`

    Current holds the input dict for the whole conversion.

    Change to snapshot then branch:

```c
// snapshot under lock
PyObject **keys = ..., **vals = ...; Py_ssize_t i = 0;
Py_BEGIN_CRITICAL_SECTION(obj);
Py_ssize_t pos = 0; PyObject *k, *v;
while (PyDict_Next(obj, &pos, &k, &v)) { Py_INCREF(k); Py_INCREF(v); keys[i]=k; vals[i]=v; i++; }
Py_END_CRITICAL_SECTION();

// then call convert_* on snapshot entries without holding the lock
```

7. Optional: `AssocList_FromObject`

    Today it builds and fills `AssocList` while locked and calls `unicode_str_and_size`.

    Safer: snapshot `(key,val)` pairs under lock, then outside lock do `unicode_str_and_size` and `AssocList_Append`. That keeps the critical section free of allocations and user code.

Why this helps

- Eliminates user-code execution under locks (encoding hooks, `__str__`, conversions).
- Keeps critical sections short and consistent.
- Preserves current ordering behavior when `order == ORDER_DEFAULT` and still supports sorted order via existing `AssocList_*` helpers.

These edits fit the existing structure where encode helpers are already separated from data access.

</details>

msgspec/_core.c
```diff
@@ -2949,7 +3007,18 @@ static PyObject* NamedTupleInfo_Convert(PyObject*);
...
-    return type->details[0].pointer;
+    StructInfo *info = type->details[0].pointer;
+#ifdef Py_GIL_DISABLED
+    if (atomic_load(&info->initialized)) {
+        return info;
+    }
+    Py_BEGIN_ALLOW_THREADS
+    /* wait for the StructInfo to be fully initialized by other thread */
+    while (!atomic_load(&info->initialized)) {
+    }
```
ofek: ---

`TypeNode_get_struct_info` spin-loops, which can burn CPU and still races if the pointer becomes visible before initialization.

Fix: Publish the pointer only after the object is fully initialized, then use a single atomic load on the pointer with acquire semantics. No flag, no wait loop:

```c
// writer
StructInfo *tmp = allocate_and_fill(...);            // fully init
atomic_store_explicit(&type->details[0].pointer, tmp, memory_order_release);

// reader
StructInfo *info = atomic_load_explicit(&type->details[0].pointer, memory_order_acquire);
```

If you cannot reorder publication, then use `_Py_yield()` or equivalent inside the loop at minimum, but prefer pointer publication after full init.

msgspec/_core.c
```diff
@@ -57,11 +58,78 @@ ms_popcount(uint64_t i) {                            \
...
 #endif
 
+#if defined(Py_GIL_DISABLED) && !PY314_PLUS
+#error "Py_GIL_DISABLED is only supported in Python 3.14+"
+#endif
+
 #if PY314_PLUS
-#define MS_IMMORTAL_INITIAL_REFCNT _Py_IMMORTAL_INITIAL_REFCNT
+#ifdef Py_GIL_DISABLED
+#define _PyObject_HEAD_INIT(type)    \
```
ofek: Last comment that might be worth transcribing.

---

You override a private CPython macro with your own initializer. That is brittle across micro-releases and platforms.

Fix: Do not redefine `_PyObject_HEAD_INIT`. Options:

- Keep the existing static sentinel approach but guarded by CPython’s own macros without shadowing names, or
- Initialize sentinels at module init with public API (allocate, set type, optionally mark immortal if a stable API exists; otherwise accept non-immortal + adjusted tests). Given you already relaxed `getrefcount` asserts, the runtime init path is acceptable and far more stable.

msgspec/_core.c
kumaraditya303: > TypeNode_get_struct_info spin-loops, which can burn CPU and still races if the pointer becomes visible before initialization.

I don't understand, the load uses seq-cst memory order so there is no data race possible here.
I have tested and verified it using TSAN as well on 3.14 and no data races were there.

Also, this is really an edge case when two threads tries to serialize an object whose typeNode is not fully constructed yet, the performance doesn't matter here as once the node is constructed the fastpath would be taken.

PR #895, opened 2025-10-26
Update codespell in pre-commit

Hasnep:


Hasnep:
> Thanks! Was this not working before? I'm curious why that typo wasn't caught.

Codespell checks words against a list of common typos rather than using a dictionary :)

ofek:
Ah that's an interesting approach, thanks for explaining 🙇‍♂️

Hasnep:
A dictionary approach has far too many false positives to be used in CI or as a pre-commit check, I've tried 😅

PR #896, opened 2025-10-26
Enable import sorting Ruff rule

Hasnep:
The option `tool.ruff.lint.isort.combine-as-imports` is enabled in the `pyproject.toml` file, so I assume that import sorting is intended to be on. If this is not wanted, I can open a new PR to remove the config option from `pyproject.toml`.


Hasnep:
I've pushed an updated version of this that should pass the CI checks :)

PR #897, opened 2025-10-26
Check links in docs

ofek:
As a follow-up to #893, this adds [Lychee](https://github.com/lycheeverse/lychee) to CI to prevent such issues in future.

I removed all references to EdgeDB as the [rebrand](https://www.geldata.com/blog/edgedb-is-now-gel-and-postgres-is-the-future) broke all links and the client library has changed (if not in functionality then definitely in name and imports). Anybody is free to add it back properly and I'll get that merged ASAP.


ofek:
Scratch the part about adding the example back post-rebrand as the company shut down/was acquired. https://www.geldata.com/blog/gel-joins-vercel

PR #899, opened 2025-10-26
Fix refcount leak to `re.Pattern`  when using `Meta(pattern=...)`

provinzkraut:
When using `msgspec.Meta(pattern="something")`, msgspec creates a leaky reference to an `re.Pattern`, by increasing its reference count too many times.

This is caused by the fact that it creates the `Pattern` object with a call to `PyObject_CallOneArg`, which returns a new strong reference. Msgspec then add this to the `Meta` object, during which the Pattern's refcount is increased again. Due to this, its refcount can never fall to 0, keeping it alive indefinitely. 

We can show this with a simple script:

```python
import gc
import re
import secrets
import msgspec

string_pattern = secrets.token_hex()
msgspec.Meta(pattern=string_pattern)
re.purge()

print(
    any(o for o in gc.get_objects() if isinstance(o, re.Pattern) and o.pattern == string_pattern)
)
```
this should print `False`, since there should be no such `re.Pattern` objects arounds anymore, but this is not the case.

<hr>

I have fixed this by foregoing the `SET_FIELD` macro, and instead simply assigning the field directly, without calling `Py_XINCREF` on its value.

### DIFF
--- a/src/msgspec/_core.c
+++ b/src/msgspec/_core.c
@@ -1907,16 +1907,22 @@ Meta_new(PyTypeObject *type, PyObject *args, PyObject *kwargs) {
     }
 
     Meta *out = (Meta *)Meta_Type.tp_alloc(&Meta_Type, 0);
-    if (out == NULL) return NULL;
+    if (out == NULL) {
+        Py_XDECREF(regex);
+        return NULL;
+    }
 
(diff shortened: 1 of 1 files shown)


ofek:
Just a heads up that in my last commit to your branch I fixed another leak I noticed that was already present!

Issue #874, opened 2025-10-07
__post_init__ not called from replace


kramar11:
### Description

When using msgspec.structs.replace to create a new object, the `__post_init__` method is not called (unlike with dataclass.replace). 

``` python
from dataclasses import dataclass, replace
import msgspec
                                                                                                                   
@dataclass
class D:                                                                                                           
    x: int
    def __post_init__(self):
        print("dataclass: post init called")
                                                                                                                   
                                                                                                                   
class M(msgspec.Struct):                                                                                           
    x: int                                                                                                         
    def __post_init__(self):                                                                                       
        print("msgspec: post init called")                                                                         
                                                                                                                   
                                                                                                                   
d = D(1)                                                                                                           
m = M(1)                                                                                                           
                                                                                                                   
d2 = replace(d, x=2)                                                                                               
m2 = msgspec.structs.replace(m, x=2) 

# Output:
# dataclass: post init called
# msgspec: post init called
# dataclass: post init called

```

Tested with python 3.13.5.

provinzkraut:
Looking at the PR that introduced this, it seems like `__post_init__` is intentionally not called when doing `copy` or `replace` operations. I am not entirely sure why that is, but it should probably at least be mentioned in the documentation, as this behaviour differs from e.g. dataclasses.

kramar11:
Thanks for looking into this. 
But I think msgspec should really be calling `__post_init__` to behave like dataclasses. Also when using `copy.replace` currently `__post_init__` is not called unlike dataclasses does. This is a very unexpected behaviour for devs using msgspec as a dataclass replacement and only mentioning this in the documentation is in my opinion not really the right thing to do.

provinzkraut:
Yeah, I guess it makes sense to keep the behaviour the same. I've created #933 with a fix and some open questions.

Siyet:
Ported to the community fork [msgspec-arise](https://github.com/Siyet/msgspec-arise): [issue #25](https://github.com/Siyet/msgspec-arise/issues/25), [PR #32](https://github.com/Siyet/msgspec-arise/pull/32)

PR #900, opened 2025-10-26
Improve `__replace__` typing

provinzkraut:
- Add `__replace__` method to stub files, so `copy.replace` doesn't report a type error
- ~Document `__post_init__` behaviour of `__replace__` (see #874)~

### DIFF
--- a/docs/structs.rst
+++ b/docs/structs.rst
@@ -161,14 +161,6 @@ at the end of the generated ``__init__`` method. It has the same semantics as th
 This method may be useful for adding additional logic to the init (such as
 custom validation).
 
-.. important::
-
-    ``__post_init__`` will *not* be called when using any of
-
-    - :func:`msgspec.structs.replace`,
-    - :func:`copy.replace`,
-    - :func:`copy.copy`,
-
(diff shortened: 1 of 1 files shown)


docs/source/structs.rst
```diff
@@ -161,6 +161,14 @@ at the end of the generated ``__init__`` method. It has the same semantics as th
...
 custom validation).
 
+.. important::
+
+    ``__post_init__`` will *not* be called when using any of
+
+    - :func:`msgspec.structs.replace`,
+    - :func:`copy.replace`,
+    - :func:`copy.copy`,
+
```
CharString: For a better flow of the text, I think this block should be moved after 

> In addition to in ``__init__``, the ``__post_init__`` hook is also called when:
> 
> - Decoding into a struct type (e.g. ``msgspec.json.decode(..., type=MyStruct)``)
> - Converting into a struct type (e.g. ``msgspec.convert(..., type=MyStruct)``)

CharString (changes requested):

jcrist (approved):
Thanks!

I'm going to fix the `__replace__` behavior in a quick follow-up, so I've dropped the docs addition here.

PR #911, opened 2025-10-29
Improve build hygiene in test job

ofek:
Changes:

1. Tests now run against an installation as a user would experience rather than an editable one.
    1. Moved to a `src`-layout project structure so that the current directory is never considered for Python imports.
2. Updated licensing package metadata now that we dropped the deprecated `python setup.py` call and use a newer, isolated version of `setuptools`.
3. Removed use of https://github.com/pre-commit/action since it's in maintenance mode.
    1. Added https://github.com/tox-dev/pre-commit-uv as a development dependency to make `pre-commit` faster since the maintainer is perpetually against a direct integration.


jacopoabramo:
@ofek there's an alternative to pre-commit, [`prek`](https://github.com/j178/prek), which while still under development should provide almost complete backward compatibility with the current pre-commit actions in `msgspec`. It's written in rust so it should hopefully be faster by default.

ofek:
Dude I had no idea, thanks a lot!

PR #914, opened 2025-11-05
Add Meltano to "Used By" in the docs

edgarrmondragon:


docs/index.rst
```diff
@@ -154,6 +154,11 @@ few:
 
         .. image:: _static/django-modern-rest.png
 
+    .. grid-item-card:: Meltano
+        :link: https://sdk.meltano.com/en/latest/guides/performance.html#use-a-different-message-writer-or-reader
```
ofek: I very much like linking the actual docs or source code showing proof of usage! I do the same for Hatch & Hatchling [here](https://hatch.pypa.io/latest/community/users/).

I would feel bad though if there wasn't a link directly to the project like the others. I'm going to push one more commit.

ofek (approved):
Thanks! I'm going to use this idea from my last commit for the others in a separate PR. Here is what it looks like:

Image

edgarrmondragon:
Thanks @ofek!

PR #929, opened 2025-11-14
Move documentation

ofek:
Changed:

1. Docs now reside under `docs/` directly rather than `docs/source/` and are now built in the `site/` directory rather than `docs/build/html`.
2. Added an exclusion to the link checker because of persistent flakiness.


ofek:
The failure is expected because the path to the file does not yet exist on `main` until after merging.

PR #930, opened 2025-11-14
Ensure that CFLAGS is passed to Linux containers

ofek:


ofek:
Superseded by https://github.com/jcrist/msgspec/pull/932

PR #942, opened 2025-11-20
Create CHANGELOG.md

ofek:
This was generated by using the `gh` CLI.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,343 @@
+# Changelog
+
+---
+
+## Unreleased
+
+## [0.19.0] - 2024-12-27
+
+- Improve JSON encoding performance by up to 40% (#647).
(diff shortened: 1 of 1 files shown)


jacopoabramo:
Just a silly question but wouldn't be simpler to redirect to the documentation changelog rather than copy-pasting it? It would reduce the amount of duplicated information.

EDIT: also, thank you for the titanic effort you're pushing into the project

ofek:
Thanks for your kind words and that's a good question! I didn't have time last night but I plan to have this file replace the existing one so that entries can be copied directly to the GitHub release rather than performing a manual translation every time.

ofek:
Regarding strategy, I chose to use a small script to construct the file based on the contents of each GitHub release rather than asking AI to translate the syntax. After I overwrite the existing file and look at the diff I will try to resolve any discrepancies between this file and each release. If it's a lot then I plan to just ask AI to translate the existing file and save time.

PR #945, opened 2025-11-21
fix inspection for structs that use the `StructMeta` metaclass (#941)

### LINKED ISSUE #941: Cannot do inspection when using StructMeta
### Description

It seems inspection is not possible when using `StructMeta`.

The following error is returned when calling `msgspec.structs.fields`:
```
TypeError: Must be called with a struct type or instance
```

```python
from msgspec import StructMeta, Struct

# Custom metaclass with kw_only_default parameter
class KwOnlyMeta(StructMeta):
    _kw_only_default_settings = {}
    
    def __new__(mcls, name, bases, namespace, **kwargs):
        # Store and inherit kw_only_default
        kw_only_default = kwargs.pop('kw_only_default', None)
        if kw_only_default is None:
            for base in bases:
                if base.__name__ in mcls._kw_only_default_settings:
                    kw_only_default = mcls._kw_only_default_settings[base.__name__]
                    break
        else:
            mcls._kw_only_default_settings[name] = kw_only_default
            
        # Apply kw_only_default if kw_only not specified
        if 'kw_only' not in kwargs and kw_only_default is not None:
            kwargs['kw_only'] = kw_only_default
            
        return super().__new__(mcls, name, bases, namespace, **kwargs)

# Base model requiring keyword arguments
class BaseModel(metaclass=KwOnlyMeta, kw_only_default=True):
    pass

# Child class inherits kw_only=True
class User(BaseModel):
    name: str
    age: int


msgspec.structs.fields(User)
```

If you construct the class using `msgspec.Struct`, this is working fine

provinzkraut:
Allow `msgspec.structs.field` on types that use `metaclass=StructMeta`.

Fixes #941.

### DIFF
--- a/docs/structs.rst
+++ b/docs/structs.rst
@@ -1020,7 +1020,7 @@ ABCs.
 .. _struct-gc:
 
 Disabling Garbage Collection
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 .. warning::
 


ofek:
Thanks! Have you confirmed that this also fixes the second comment in that issue? It appears to be a different error.

provinzkraut:
Ah, I didn't see that a second comment was added! I'll take a look at this as well and update this PR

ofek:
Do you have an estimate on when you might be able to check? I was planning on releasing in a few hours.

provinzkraut:
I've fixed `inspect.type_info`. `asdict` did work as expected already.

ofek:
I added another test to confirm the expected behavior!

provinzkraut:
@ofek seems like that test you added uncovered another case that was still buggy. I've pushed an additional fix for that. 

I've got the feeling though that there might still have some cases left elsewhere in the library where `StructMeta` isn't handled as expected :eyes:

ofek:
I pushed some code changes and a doc update. What we want to express in the type system currently is not possible, example discussion in https://github.com/python/typing/discussions/1489.

ofek (approved):
Thanks a lot for taking the initiative here!

PR #954, opened 2025-11-24
Drop support for Python 3.9

jacopoabramo:
Closes #904 

In detail:

- remove (almost) all references to `Union` and `Optional`
- remove `.pytest.ini` as noted by the TODO;
- use built-in `type`, `tuple`, `list`, and `dict` types...
  -  ... although I feel this is a mistake on my part, as type checking tests are failing;
  - @ofek any input in this is appreciated, should I revert back to use the `typing` imports rather than built-ins for specific occasions?
  - I'm hoping the runner gives some insight on the problems
- remove dependency on `eval-type-backport` (required for 3.9)
- make all code related to `PY310_PLUS` default as it should be now the minimal supported version

PR #956, opened 2025-11-25
Drop Python 3.9

### LINKED ISSUE #904: Drop support for 3.9?
### Question

I'm not exactly sure about `msgspec` deprecation policy (if there's such a thing), but given that 3.9 reaches EOL end of this month, wouldn't it be wise to drop support for 3.9 on the next release?

jacopoabramo:
Second attempt after #954

Closes #904 

- Remove `.pytest.ini`
- Remove references to `3.9` in CI
- Use hooks to autoformat imports in `kv.py`
- Remove dependency on `eval-type-backport` for `< 3.10`
- Remove `PY310_PLUS` check

### DIFF
--- a/.github/workflows/profile.yml
+++ b/.github/workflows/profile.yml
@@ -38,23 +38,23 @@ jobs:
     - name: Load environment file
       uses: ./.github/actions/load-env
 
-    - name: Install Python
-      uses: actions/setup-python@v6
-      with:
-        python-version-file: pyproject.toml
-
-    - name: Install uv
-      uses: astral-sh/setup-uv@v7
-
-    - name: Install command runner
-      run: uv tool install rust-just
-
     - name: Fetch latest supported Python
       if: inputs.python == ''
       id: latest-supported-python
       uses: ./.github/actions/latest-supported-python
       with:
         github-token: "${{ github.token }}"
+
+    - name: Install Python
+      uses: actions/setup-python@v6
+      with:
+        python-version: "${{ inputs.python || steps.latest-supported-python.outputs.version }}"
+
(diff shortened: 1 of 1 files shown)


jacopoabramo:
I don't think the falling test is due to this PR, right?

ofek:
Indeed, it looks like uv is downloading the wrong Python for that architecture. I'll look into that tonight.

ofek:
https://github.com/jcrist/msgspec/pull/959

jacopoabramo:
@ofek it is a bit annoying that PRs require approval for executing the CI/CD workflows. I can understand this is a safety measure (don't really know why but I can appreciate the need for harder security). Do you think that this could be changed at some point? It would be useful for PRs to automatically run the tests suite to check whether there's a fail in the tests or not, and would help you catch a glimpse of possible failures yourself (as in the case of #954 where I abruptly closed the PR - which now I regret but what can one do).

ofek:
I think CI runs automatically after someone's first PR is merged. I'm not an admin so I couldn't change the setting but I'll reach out to Jim about that.

jacopoabramo:
> I think CI runs automatically after someone's first PR is merged

Ah, if that's the case I can live with that!

ofek:
> as in the case of #954 where I abruptly closed the PR - which now I regret but what can one do

Feel free to reopen and get to the bottom of what's going on there, I'm quite curious as well!

jacopoabramo:
> Feel free to reopen and get to the bottom of what's going on there, I'm quite curious as well!

I will; I have to try and find some time where I can make a reasonable investigation but I'll get back to it at some point.

PR #962, opened 2025-11-29
Fix handling of type var syntax and `types.GenericAlias`

### LINKED ISSUE #957: Types inheriting from GenericAlias types (such as `Mapping`) cannot be decoded.
### Description

If you try to make a generic class that inherits from one of the `_SpecialGenericAlias` types (such as `Mapping`), the type inference fails to work.

A minimal example below (Python 3.13, msgspec 0.20.0):

```python
from collections.abc import Mapping
from dataclasses import dataclass

import msgspec


@dataclass
class Bar[T](Mapping[str, T]):
    data: dict[str, T]

    def __getitem__(self, x):
        return self.data[x]

    def __len__(self):
        return len(self.data)

    def __iter__(self):
        return iter(self.data)


x: Bar[int] = Bar({"x": 3})

msgspec.msgpack.decode(msgspec.msgpack.encode(x), type=Bar[int])
```

with traceback

```
  File "<censored>", line 31, in <module>
    msgspec.msgpack.decode(msgspec.msgpack.encode(x), type=Bar[int])
    ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/venv/lib/python3.13/site-packages/msgspec/_utils.py", line 252, in get_dataclass_info
    hints = get_class_annotations(obj)
  File "/opt/venv/lib/python3.13/site-packages/msgspec/_utils.py", line 149, in get_class_annotations
    mro, typevar_mappings = _get_class_mro_and_typevar_mappings(obj)
                            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^
  File "/opt/venv/lib/python3.13/site-packages/msgspec/_utils.py", line 126, in _get_class_mro_and_typevar_mappings
    inner(obj, {})
    ~~~~~^^^^^^^^^
  File "/opt/venv/lib/python3.13/site-packages/msgspec/_utils.py", line 124, in inner
    inner(b, new_scope)
    ~~~~~^^^^^^^^^^^^^^
  File "/opt/venv/lib/python3.13/site-packages/msgspec/_utils.py", line 116, in inner
    params = cls.__parameters__
             ^^^^^^^^^^^^^^^^^^
AttributeError: type object 'Mapping' has no attribute '__parameters__'
```

I believe this is caused by the logic in `inner(...)` in the utils. For this, we look up `__origin__` - however, in these specific cases, the types library has `typing.Mapping` as a `typing._SpecialGenericAlias`, which has different behaviour - it's `__origin__` is not `typing.Mapping`, it's `collections.abc.Mapping`. This has no generic type parameters so inference fails. 

This looks extremely tough to wade through so not sure what a good workaround is here - not a typing expert!

I'll note briefly that this behaviour goes away if it's not generic in inheritance, but it still fails if I use `msgspec.Struct` and a combined metaclass from `StructMeta` and `ABCMeta` so is not an artifact of dataclasses or anything, I think.

provinzkraut:
Fix #957.

This is a few fixes combined into one, since they were tightly coupled.

**Handling of "new style" generics during type resolution**

When subscribing a "new style" generic (such as `collections.abc.Mapping`), it produces a `types.GenericAlias` (vs. the "old style" `typing._GenericAlias`), which msgspec did not handle correctly during inspection

**Handling of type var syntax**

When dealing with builtin generics that resolve to `typing.TypeAlias`, msgspec did not account for type var syntax at correctly in all cases, so type information would get lost during the conversion process

**Type conversions on `types.GenericAlias`**

During type conversion, msgspec caches certain information on the type objects themselves, if the types are complex (i.e. `Struct`s or `dataclass`-like).

When decoding into a `Foo[int]`, msgspec will set an `__msgspec_cache__` attribute on the `Foo[int]` alias type.

For `typing._GenericAlias`, this work, since it has a `__dict__`, so you can just assign attributes to it. However, `types.GenericAlias` does *not* allow assigning arbitrary attributes to it.

I fix this by downtyping `types.GenericAlias` into a `typing._GenericAlias`, when encountering a generic `Struct` or `dataclass` type. This allows to keep the existing caching mechanism in place.

This seemed like the most reasonable fix to me, as the other alternatives would like incur some sort of performance penalty; By storing the typing info directly on the alias, msgspec can forego maintaining a dedicated cache, making lookups very fast. It also allows to not care about invalidating a cache, since it will just be gce'd when the alias isn't referenced anymore.

One thing to not here though is that in the future, `typing._GenericAlias` might just go away (at least from the stdlib), in which case we'll have to find another way to deal with this.


provinzkraut:
Okay, maybe this isn't a good solution. I've discovered a slight issue with the `typing._GenericAlias` implementation:

It does not cache itself when called directly (i.g. doing `_GenericAlias(<type>, <args>)`, but only when subscribed. This means that `List[int] is List[int]` holds true, but `_GenericAlias(List, int) is _GenericAlias(List, int)` does not. 

Since msgspec relies on that cache being preserved, the solution as currently proposed only maintains the cache within a decoder instance, meaning that calling `decode()` twice, even with the same type, does not maintain the cache. 

Note that this *only* affects generic types bound by a `types.GenericAlias`. Everything else is unaffected, including "regular" generic classes (e.g. a `class Foo[T](Struct)` will be cached just fine.

provinzkraut:
Alright, I think I managed to find a solution. Essentially, we're now reproducing a `typing._GenericAlias` that properly caches itself, from a `types.GenericAlias`.

It's basically doing
```python
typing._GenericAlias(
    alias.__origin__, 
    alias.__origin__.__parameters__
).__getitem__(*alias.__args__)
 ```
which is functionally the same as

```python
alias = typing._SpecialGenericAlias(list, 1)[int]
# this is the same as
alias = typing.List[int]
```

<hr>

I've also added some more test cases to ensure the caching works properly in all newly supported cases.

provinzkraut:
Another friendly ping to @ofek for a review :)
Hope you don't mind, I just didn't want these PRs to get buried, but I also don't want to continue to send PRs and have them pile up :sweat_smile: 
LMK if I can do something to make the review process easier for you.

ofek:
I've been really busy, sorry for the wait! Can you confirm locally that the minor performance regression [here](https://github.com/jcrist/msgspec/actions/runs/19798482440/job/56722363258?pr=962#step:12:28) compared to the [last run](https://github.com/jcrist/msgspec/actions/runs/19742095662/job/56569067225#step:12:28) on `main` is just noise?

I really appreciate all the work you've been doing 🙏

provinzkraut:
> Can you confirm locally that the minor performance regression [here](https://github.com/jcrist/msgspec/actions/runs/19798482440/job/56722363258?pr=962#step:12:28) compared to the [last run](https://github.com/jcrist/msgspec/actions/runs/19742095662/job/56569067225#step:12:28) on `main` is just noise?

I'll check it out!

provinzkraut:
> I've been really busy, sorry for the wait! Can you confirm locally that the minor performance regression [here](https://github.com/jcrist/msgspec/actions/runs/19798482440/job/56722363258?pr=962#step:12:28) compared to the [last run](https://github.com/jcrist/msgspec/actions/runs/19742095662/job/56569067225#step:12:28) on `main` is just noise?

Seems to be just noise, on my machine (Apple Silicon M4) at least.

src/msgspec/_utils.py
```diff
@@ -22,6 +23,8 @@ def get_type_hints(obj):
     return _get_type_hints(obj, include_extras=True)
 
 
+PY_31PLUS = sys.version_info >= (3, 12)
```
Siyet, line 26: Typo in variable name: `PY_31PLUS` reads as "Python 3.1+", but the comparison is `>= (3, 12)`. Should be `PY_312PLUS` (consistent with `PY312_PLUS` used in `_core.c`).

Siyet:
## Performance & correctness review

Tested on a dedicated VPS, pinned to a single core via `taskset -c 0`.

<details>
<summary>VPS specs</summary>

- **OS**: Ubuntu 24.04.3 LTS, kernel 6.8.0-107-generic
- **CPU**: AMD EPYC (with IBPB), 1 vCPU
- **RAM**: 1.9 GiB
- **Python**: 3.12.3
- **Method**: A/B/A (main -> PR -> main), 3x100 rounds per benchmark, trimmed mean (drop top/bottom 10%), `taskset -c 0`

</details>

### Benchmark results

| Benchmark | main_1 (ms) | PR (ms) | main_2 (ms) | PR vs avg(main) | stdev (m1/pr/m2) | verdict |
|---|---|---|---|---|---|---|
| point_encode | 0.0788 | 0.0810 | 0.0841 | -0.5% | 2.4/2.5/15.8% | noise |
| point_decode | 0.0837 | 0.0888 | 0.0871 | +4.0% | 2.6/1.4/4.4% | noise (main drifts) |
| generic_struct_decode | 0.0587 | 0.0603 | 0.0586 | +2.7% | 1.2/1.5/1.4% | borderline, see below |
| decoder_creation_simple | 0.0903 | 0.0925 | 0.0914 | +1.9% | 1.4/1.9/2.3% | noise |
| decoder_creation_generic | 0.1681 | 0.1696 | 0.1691 | +0.6% | 1.1/3.7/1.6% | noise |

`generic_struct_decode` shows +2.7%, but this benchmark uses a pre-created `Decoder(Box[int])` where `Box` is `Struct, Generic[T]` (a `typing._GenericAlias`, not a `types.GenericAlias`). The `convert_types_generic_alias` path is not hit here, so this is VPS noise despite the low stdev.

**Conclusion: no measurable performance regression.**

### Correctness verification

6/6 tests pass on the PR branch:

| Test | Status |
|---|---|
| dataclass inheriting `collections.abc.Mapping[str, T]` (original #957) | PASS |
| Struct + ABCMeta + `collections.abc.Mapping[str, T]` | PASS |
| `typing.Mapping[str, T]` variant | PASS |
| PEP 695 type var syntax (Python 3.12) | PASS |
| Refcount / cache across decoder instances | PASS |
| Cache persistence across 10 decode calls | PASS |

### Memory leak test

10,000 iterations of `Decoder(Foo[int])` where `Foo` inherits `collections.abc.Mapping` (exercising `convert_types_generic_alias` on every call):
- RSS growth: 60 KB (within noise)
- tracemalloc: 3.4 KB growth (from test scaffolding)

No memory leak detected.

<details>
<summary>Benchmark script</summary>

```python
import timeit
import statistics
import json
import typing

import msgspec
from msgspec import Struct

T = typing.TypeVar("T")

class Point(Struct):
    x: int
    y: int
    z: float

class Box(Struct, typing.Generic[T]):
    value: T

enc = msgspec.json.Encoder()
dec_point = msgspec.json.Decoder(Point)
dec_box_int = msgspec.json.Decoder(Box[int])
point_msg = enc.encode(Point(1, 2, 3.14))
box_msg = enc.encode(Box(value=42))

def bench_point_encode():
    p = Point(1, 2, 3.14)
    for _ in range(1000):
        enc.encode(p)

def bench_point_decode():
    for _ in range(1000):
        dec_point.decode(point_msg)

def bench_generic_struct_decode():
    for _ in range(1000):
        dec_box_int.decode(box_msg)

def bench_decoder_creation_simple():
    for _ in range(200):
        msgspec.json.Decoder(Point)

def bench_decoder_creation_generic():
    for _ in range(200):
        msgspec.json.Decoder(Box[int])

benchmarks = [
    ("point_encode", bench_point_encode),
    ("point_decode", bench_point_decode),
    ("generic_struct_decode", bench_generic_struct_decode),
    ("decoder_creation_simple", bench_decoder_creation_simple),
    ("decoder_creation_generic", bench_decoder_creation_generic),
]

# Warmup
for name, func in benchmarks:
    timeit.repeat(func, number=1, repeat=10)

# 3 full runs of 100 rounds each
all_results = {}
for run in range(3):
    for name, func in benchmarks:
        times = timeit.repeat(func, number=1, repeat=100)
        times_sorted = sorted(times)
        trimmed = times_sorted[10:-10]  # drop top/bottom 10%
        if name not in all_results:
            all_results[name] = []
        all_results[name].extend(trimmed)

results = {}
for name, times in all_results.items():
    results[name] = {
        "mean": statistics.mean(times),
        "stdev": statistics.stdev(times),
        "min": min(times),
        "p50": statistics.median(times),
    }

print(json.dumps(results, indent=2))
```

</details>

<details>
<summary>Verification script</summary>

```python
import sys
import typing
import collections.abc
import dataclasses
import gc

import msgspec
from msgspec import Struct


results = []

def test(name, func):
    try:
        func()
        results.append((name, "PASS", ""))
    except Exception as e:
        results.append((name, "FAIL", str(e)))


def test_dataclass_mapping():
    @dataclasses.dataclass
    class Bar(typing.Generic[typing.T], collections.abc.Mapping[str, typing.T]):
        data: dict[str, typing.T]
        def __getitem__(self, x): return self.data[x]
        def __len__(self): return len(self.data)
        def __iter__(self): return iter(self.data)

    x = Bar(data={"x": 3})
    encoded = msgspec.msgpack.encode(x)
    decoded = msgspec.msgpack.decode(encoded, type=Bar[int])
    assert decoded == x

test("dataclass_mapping_generic", test_dataclass_mapping)


def test_struct_mapping():
    import abc
    class CombinedMeta(msgspec.structs.StructMeta, abc.ABCMeta):
        pass

    T = typing.TypeVar("T")

    class Foo(collections.abc.Mapping[str, T], Struct, typing.Generic[T], metaclass=CombinedMeta):
        data: dict[str, T]
        def __getitem__(self, x): return self.data[x]
        def __len__(self): return len(self.data)
        def __iter__(self): return iter(self.data)

    encoded = msgspec.msgpack.encode(Foo(data={"x": 1}))
    decoded = msgspec.msgpack.decode(encoded, type=Foo[int])
    assert decoded.data == {"x": 1}

    try:
        msgspec.msgpack.decode(
            msgspec.msgpack.encode(Foo(data={"x": "foo"})), type=Foo[int]
        )
        assert False, "Should have raised ValidationError"
    except msgspec.ValidationError:
        pass

test("struct_mapping_generic", test_struct_mapping)


def test_typing_mapping():
    T = typing.TypeVar("T")

    @dataclasses.dataclass
    class Bar(typing.Generic[T], typing.Mapping[str, T]):
        data: dict[str, T]
        def __getitem__(self, x): return self.data[x]
        def __len__(self): return len(self.data)
        def __iter__(self): return iter(self.data)

    x = Bar(data={"x": 3})
    encoded = msgspec.msgpack.encode(x)
    decoded = msgspec.msgpack.decode(encoded, type=Bar[int])
    assert decoded == x

test("typing_mapping_generic", test_typing_mapping)


if sys.version_info >= (3, 12):
    def test_typevar_syntax():
        code = '''
from msgspec import Struct
class Ex[T](Struct):
    x: T
    y: list[T]
'''
        ns = {}
        exec(code, ns)
        Ex = ns["Ex"]
        msg = msgspec.json.encode(Ex(1, [1, 2]))
        res = msgspec.json.decode(msg, type=Ex[int])
        assert res.x == 1 and res.y == [1, 2]

    test("pep695_typevar_syntax", test_typevar_syntax)


def test_refcount_cache():
    T = typing.TypeVar("T")

    @dataclasses.dataclass
    class Foo(typing.Generic[T], collections.abc.Mapping[str, T]):
        data: dict[str, T]
        def __getitem__(self, x): return self.data[x]
        def __len__(self): return len(self.data)
        def __iter__(self): return iter(self.data)

    typ = Foo[int]
    dec1 = msgspec.json.Decoder(typ)
    dec2 = msgspec.json.Decoder(typ)
    msg = msgspec.json.encode(Foo(data={"a": 1}))
    r1 = dec1.decode(msg)
    r2 = dec2.decode(msg)
    assert r1 == r2
    del dec1, dec2
    gc.collect()

test("refcount_cache", test_refcount_cache)


def test_cache_persistence():
    T = typing.TypeVar("T")

    @dataclasses.dataclass
    class Foo(typing.Generic[T], collections.abc.Mapping[str, T]):
        data: dict[str, T]
        def __getitem__(self, x): return self.data[x]
        def __len__(self): return len(self.data)
        def __iter__(self): return iter(self.data)

    typ = Foo[int]
    msg = msgspec.msgpack.encode(Foo(data={"a": 1}))
    for i in range(10):
        dec = msgspec.msgpack.Decoder(typ)
        result = dec.decode(msg)
        assert result.data == {"a": 1}

test("cache_persistence", test_cache_persistence)

for name, status, detail in results:
    marker = "OK" if status == "PASS" else "FAIL"
    print(f"[{marker}] {name}" + (f" - {detail}" if detail else ""))
print(f"\n{len(results)} tests, {sum(1 for _, s, _ in results if s == 'FAIL')} failed")
```

</details>

<details>
<summary>Memory leak test script</summary>

```python
import gc
import sys
import typing
import collections.abc
import dataclasses
import tracemalloc

import msgspec
from msgspec import Struct

T = typing.TypeVar("T")

@dataclasses.dataclass
class Foo(typing.Generic[T], collections.abc.Mapping[str, T]):
    data: dict[str, T]
    def __getitem__(self, x): return self.data[x]
    def __len__(self): return len(self.data)
    def __iter__(self): return iter(self.data)

tracemalloc.start()
gc.collect()

with open("/proc/self/status") as f:
    rss_before = int([l for l in f if l.startswith("VmRSS:")][0].split()[1])

snap1 = tracemalloc.take_snapshot()

for i in range(10000):
    dec = msgspec.json.Decoder(Foo[int])
    msg = msgspec.json.encode(Foo(data={"a": 1}))
    dec.decode(msg)
    del dec

gc.collect()

with open("/proc/self/status") as f:
    rss_after = int([l for l in f if l.startswith("VmRSS:")][0].split()[1])

snap2 = tracemalloc.take_snapshot()

print(f"RSS diff: {rss_after - rss_before} KB")
for stat in snap2.compare_to(snap1, "lineno")[:5]:
    print(f"  {stat}")
```

</details>

src/msgspec/_utils.py
provinzkraut: fixed

provinzkraut:
@Siyet @ofek if you got the time :)
Would be nice to get this one in!

provinzkraut:
@sobolevn would be nice if you could take a look as well :)

src/msgspec/_core.c
```diff
@@ -4984,6 +4985,49 @@ is_dataclass_or_attrs_class(TypeNodeCollectState *state, PyObject *t) {
...
+        // false.
+        // to achieve the same behaviour when re-creating a typing._GenericAlias from a
+        // types.GenericAlias, we first construct a temporary *unbound*
+        // typing._GenericAlias, on which we then call __getattr__. effectively doing
+        // typing._GenericAlias(list, T)[int], for which
+        // 'typing._GenericAlias(list, T)[int] is typing._GenericAlias(list, T)[int]'
+        // holds true
+        PyObject *params = PyObject_GetAttrString(origin, "__parameters__");
+        if (params == NULL) {
+            Py_DECREF(origin);
```
sobolevn: why do we change the rc of `origin` here? We don't own it. why do we only decref it here and not in other branches? It is always cleared in `done: Py_XDECREF(origin);`

tests/unit/test_common.py
```diff
@@ -50,6 +50,13 @@
 T = TypeVar("T")
 
 
+def make_new_alias(alias):
+    """Replicate the logic to produce a types._GenericAlias from a typing.GenericAlias"""
+    return typing._GenericAlias(
```
sobolevn: Let's please test other subclasses like `typing._CallableGenericAlias`, `_LiteralGenericAlias`, etc. They can also be affected.

src/msgspec/_core.c
```diff
@@ -4984,6 +4985,49 @@ is_dataclass_or_attrs_class(TypeNodeCollectState *state, PyObject *t) {
...
 
+static MS_INLINE PyObject*
+convert_types_generic_alias(TypeNodeCollectState *state, PyObject *obj, PyObject *origin, PyObject *args) {
+    // if 'obj' is a 'types.GenericAlias', convert it into a 'typing._GenericAlias', so
+    // we can cache type info on it. 'types.GenericAlias' has __slots__, so caching on
+    // it directly does not work.
+    // it's unlikely to hit this case, as it will mostly occur when subclassing a
+    // built-in container generic, such as 'collections.abc.Mapping'
+
+    if (MS_UNLIKELY(Py_TYPE(obj) == (PyTypeObject *)state->mod->types_generic_alias)) {
```
sobolevn: You don't need to store it in a state, `Py_GenericAliasType` was added in Python 3.9

src/msgspec/_core.c
```diff
@@ -4984,6 +4985,49 @@ is_dataclass_or_attrs_class(TypeNodeCollectState *state, PyObject *t) {
...
+        // original types.GenericAlias.
+        // given a Mapping[str, int], this would produce a _GenericAlias(Mapping, (~K, ~V))
+        PyObject *new_alias = PyObject_CallFunctionObjArgs(state->mod->typing_generic_alias, origin, params, NULL);
+        if (new_alias == NULL) {
+            return NULL;
+        }
+
+        // bind it to the concrete types.
+        // given a _GenericAlias(Mapping, (~K, ~V)), produce a Mapping[str, int] again
+        PyObject *result = PyObject_CallMethod(new_alias, "__getitem__", "O", args);
```
sobolevn: Maybe instead of this C code, we can store some Python builder for this? Because I fear that it might hit quite a lot of corner cases. It would be hard to fix them in C. Especially given that `typing._GenericAlias` is not even a public API.

Since we call Python code here anyway, why not call our own wrapper?

tests/unit/test_common.py
```diff
@@ -1482,6 +1489,35 @@ class Ex(Struct, Generic[T], array_like=array_like):
         with pytest.raises(ValidationError, match="Expected `str`, got `int`"):
             proto.decode(msg, type=Ex[str])
 
+    @py312_plus
+    def test_generic_with_typevar_syntax(self, proto):
+        source = """
+        from msgspec import Struct
+        from typing import List
```
sobolevn: if deprecated `List` form is important here, please add a comment, so we won't remove it in the future.

tests/unit/test_common.py
```diff
@@ -1482,6 +1489,35 @@ class Ex(Struct, Generic[T], array_like=array_like):
...
+            sol = mod.Ex(1, [1, 2])
+            msg = proto.encode(sol)
+
+            res = proto.decode(msg, type=mod.Ex)
+            assert res == sol
+
+            res = proto.decode(msg, type=mod.Ex[int])
+            assert res == sol
+
+            res = proto.decode(msg, type=mod.Ex[Union[int, str]])
```
sobolevn, line 1556: Please, test `|` as well, seems related :)

tests/unit/test_common.py
```diff
@@ -1776,6 +1977,129 @@ def test_unbound_typevars_with_constraints_unsupported(self, proto):
...
+            def __iter__(self):
+                return iter(self.data)
+        """
+
+        with temp_module(source) as mod:
+            typ = mod.Foo[int]
+            dec = msgspec.json.Decoder(typ)
+            info = make_new_alias(typ).__msgspec_cache__
+            assert info is not None
+            assert sys.getrefcount(info) <= 4  # info + attr + decoder + func call
```
sobolevn, line 2450: `getrefcount`'s result is an implementation detail, so it can change at any moment. Recent change was to introduce "immortal" objects. This assert can be removed. But, I don't have a strong opinion here.

But, if want to keep this, let's call `gc.collect()` before checking the RC values :)

src/msgspec/_core.c
provinzkraut: > Since we call Python code here anyway, why not call our own wrapper?

You mean just this section, or the entire function? 

I guess we could move this function into `_utils.py`, yeah. We got half of the type machinery in there anyway

sobolevn: Whatever you feel would be best :) In my initial idea - it was only this section for type conversion :)

Siyet (dismissed):
Reviewed this carefully (built the branch on 3.12). A few things on top of what's already in the thread.

**1. Silent validation bypass on multi-level generic hierarchies.**

A subclass-of-a-subclass inheriting a builtin generic loses its type binding, so element validation is skipped. Built on this branch (3.12):

```python
from collections.abc import Mapping
from dataclasses import dataclass
import msgspec

@dataclass
class Base[T](Mapping[str, T]):
    data: dict[str, T]
    def __getitem__(self, x): return self.data[x]
    def __len__(self): return len(self.data)
    def __iter__(self): return iter(self.data)

@dataclass
class Child[U](Base[U]):
    pass

msgspec.json.decode(b'{"data": {"x": "not_an_int"}}', type=Child[int])
# -> Child(data={'x': 'not_an_int'})   # accepted, should raise ValidationError
```

Single-level `Base[int]` rejects it correctly; only the multi-level case breaks. Root cause is the new branch in `_utils.py` `inner()`: `dict(zip(c.__parameters__, typing.get_args(c)))` zips the alias' *free* typevars against *all* args (for `Mapping[str, T]` that is `zip((T,), (str, T)) -> {T: str}`), and it skips the `_apply_params(a, scope)` recursion the else-branch does, so the outer binding (`U -> int`) never propagates.

**2. More ownership issues in `convert_types_generic_alias` (beyond the `origin` decref you already flagged).**

- `params` (new ref from `GetAttrString`) is never `Py_DECREF`'d on any path.
- On the error paths the function returns NULL, which is passed straight into `typenode_collect_struct`/`_dataclass` with no NULL check -> `Py_INCREF(NULL)` -> crash.
- The return contract is mixed (borrowed `obj` vs new `result`), and the new `result` is not released by the caller. On the convert path I measured ~2 extra retained objects per conversion vs a non-Mapping control; a `just debug=1` refcount run would pin it down.

**3. Test bug.** The `future` parametrize passes `pytest.param(False, ...)` for both ids, so the `from __future__ import annotations` path is never actually exercised.

All built and reproduced on 3.12.

src/msgspec/_core.c
provinzkraut: Moved the alias construction to Python

tests/unit/test_common.py
provinzkraut: Can they? Iiuc, `Literal` always produces a `_LiteralGenericAlias`, or is there a way to get it to produce a `types.GenericAlias` as well?

tests/unit/test_common.py
provinzkraut: Not important

tests/unit/test_common.py
provinzkraut: It's just a sanity check. Do you have a better idea how to reliably prove this?

tests/unit/test_common.py
sobolevn: This is a rare case, but technically it is possible:

```python
>>> import typing
>>> import types
>>> x = types.GenericAlias(typing.Literal, (1,))
>>> x
typing.Literal[1]
>>> type(x)
<class 'types.GenericAlias'>
```

It might happen in C code, where people use `Py_GenericAlias` on some passed type, it can happen in some other complex substitutions, etc. (again, this is not ordinary and probably is not 100% correct)

provinzkraut: Interesting. Let me see how can can handle those.

provinzkraut:
@Siyet I fixed 1 and 3. 2 was already fixed as a side effect of moving the construction into Python

tests/unit/test_common.py
provinzkraut: Alright, I think I got it down. That was quite the rabbit hole actually and I uncovered a few edge cases previously not handled.

This PR has an interesting ratio of code:tests+comments now, but it should be worth it :)

provinzkraut:
@sobolevn I want to complain! :sweat_smile: 
`isinstance(list[int], type)` is `False` on 3.11+, but `True` in earlier versions, even though both return `types.GenericAlias` :/ 

Will have to work in a workaround for that.

sobolevn (dismissed):

sobolevn:
@provinzkraut yes, sadly typing runtime contains lots of bugs :(
Especially with `from __future__ import annotations` on 🌚️️

PR #965, opened 2025-12-16
Remove Union and Optional Typehints and replace with pipe annotations

Vizonex:
I would assume that __3.9 is dropped__ at this point in time so I made some pretty significant changes to how annotations are written out in the main library using Ruff with [UP045](https://docs.astral.sh/ruff/rules/non-pep604-annotation-union/) and [UP007](https://docs.astral.sh/ruff/rules/non-pep604-annotation-optional/) to help remove `typing.Optional` and `typing.Union` since these items are now treated as deprecated after __3.9__ if we did want backwards compatability for 3.9 we could add in `from __future__ import annotations` but I would assume that 3.9 is dropped for good, hence this pull request being made by me.


ofek:
Perhaps you can answer this question https://github.com/jcrist/msgspec/pull/954#issuecomment-3582583147

provinzkraut:
> since these items are now treated as deprecated after 3.9 if we did want backwards compatability for 3.9 we could add in `from __future__ import annotations` but I would assume that 3.9 is dropped for good, hence this pull request being made by me.

IMO we should support these feature as long as they're in a supported Python version. Otherwise we're not *really* compatible with e.g. Python 3.9, if you can't use certain 3.9 features just because they're deprecated

PR #969, opened 2026-01-07
Fix logo in README for dark and light modes

luwqz1:


Siyet:
Pushed a tiny follow-up commit dropping the `width="35%"` attribute from the two `<source>` tags. Per HTML spec, `width` on `<source>` inside `<picture>` only accepts integer pixel values, not percentages, and the actual sizing was already coming from the fallback `` anyway. Otherwise the change is exactly what we want for dark/light README rendering. Thanks for the fix!

PR #978, opened 2026-02-11
Add callable support for `decimal_format` option

### LINKED ISSUE #848: Decimal should be handled as a custom type during demarshaling.
Hey!

I have a problem when I use msgspec and Decimal. One popular solution to using Decimal is to not round the data as part of the algorithm execution and business logic. The rounding happens during demarshaling at the API endpoints. Decimal is the native type, so it's not possible to create an Encoder that can handle this.

Is it possible to change the behaviour for Decimal and allow interception in enc_hook?

Example:
```python
import msgspec
from typing import Any, Type


def enc_hook(obj: Any) -> Any:
    if isinstance(obj, Decimal):
        return str(value.quantize(Decimal("0.0001"), rounding=ROUND_HALF_UP))
    else:
        raise NotImplementedError(f"Objects of type {type(obj)} are not supported")


class Price(msgspec.Struct):
    primecost: Decimal

enc = msgspec.json.Encoder(enc_hook=enc_hook)

price = Price(primecost=Decimal("1.1234444"))

buf = enc.encode(price)
```

Creating a custom Decimal `MyDecimal` looks like a workaround.

NyanFisher:
Hello!

## Description of the problem solved by this PR

The msgspec library has many useful features, but the current version lacks the ability to correctly quantize Decimal values during encoding. In applications related to finance and precise calculations, it is critical to take into account maximum accuracy and return values rounded to a specified precision. Without implementing this functionality, a complete transition from pydantic to msgspec is not possible.

### Core Functionality
- Added DECIMAL_FORMAT_CALLABLE enum value to support callable decimal_format
- Added decimal_callable field to EncoderState and Encoder structs
- Added `in_decimal_callable` field to `EncoderState` struct
- Added recursion protection in `mpack_encode_decimal()` and `json_encode_decimal()`
- Implemented callable invocation for both JSON and MessagePack encoders

### Validation & Safety
- Added runtime check preventing callable from returning Decimal (avoids infinite recursion)
  - Direct `Decimal` returns: `lambda d: d`
  - Nested structures: `lambda d: [d]`, `lambda d: {"key": d}`

### Type Hints
- Updated json.pyi and msgpack.pyi stubs to include callable type hints:
```python
decimal_format: Union[
    Literal["string", "number"],
    Callable[[decimal.Decimal], Any],
]
```

#### Rounding to 2 Decimal Places
```python
import msgspec
import decimal

enc = msgspec.json.Encoder(
    decimal_format=lambda d: str(d.quantize(decimal.Decimal("0.01")))
)

value = decimal.Decimal("123.456789")
print(enc.encode(value))  # b'"123.46"'
```
#### MessagePack with Rounding
```python
import msgspec
import decimal

# MessagePack with custom rounding
enc = msgspec.msgpack.Encoder(
    decimal_format=lambda d: float(d.quantize(decimal.Decimal("0.001")))
)

value = decimal.Decimal("3.14159265")
msg = enc.encode(value)
print(msgspec.msgpack.decode(msg))  # 3.142
```
#### Error: Returning Decimal from Callable
```python
import msgspec
import decimal

# INVALID: callable must not return Decimal
enc = msgspec.json.Encoder(
    decimal_format=lambda d: d.quantize(decimal.Decimal("0.01"))  # Error!
)

try:
    enc.encode(decimal.Decimal("1.234"))
except TypeError as e:
    print(e)  # callable returned a value containing a Decimal
```

I would appreciate any comments on improving or restructuring the code, as I don't often write in C.

Fix my issue - Closes https://github.com/jcrist/msgspec/issues/848


NyanFisher:
CI failures are unrelated to this change:

- "Profile Windows (ARM64)" - runner infra issue ("Install command runner" step)
- "build" (documentation) - link checker fails on a pre-existing broken link (https://github.com/jcrist/msgspec/issues/976)

All build, test, and wheel jobs pass across all platforms.

Siyet:
Code looks solid and CI is green across the matrix - nice work, especially the test coverage in `test_common.py`.

One API design question I'd like to raise before this moves forward: the current shape places `decimal_quantize` / `decimal_rounding` on the encoder itself, which means every `Decimal` field in every struct passing through that encoder gets the same scale and rounding mode. In financial code it's common to have heterogeneous Decimal fields in the same payload (e.g. `price` at scale 4, `quantity` at scale 0, `tax_rate` at scale 6) - with an encoder-level setting you'd need separate encoders per shape, which defeats most of the ergonomic win.

An alternative would be to attach quantization to the type via `Annotated[Decimal, Meta(...)]`, e.g.

```python
Price = Annotated[Decimal, Meta(decimal_quantize="0.0001", decimal_rounding="ROUND_HALF_EVEN")]
```

That composes naturally with per-field configuration, lives next to the type where the constraint is logically defined, and matches how `gt`/`ge`/`pattern` etc. already work today. The downside is more plumbing through `TypeNode` instead of one encoder kwarg.

Did you consider the `Meta`-based approach? If so, what made you land on encoder-level? Both have trade-offs and I'd rather get the API right before merge.

cc @jcrist @ofek — this expands the encoder API surface, so I'd like your read on whether the encoder-kwarg shape is the one we want, or whether `Meta`-based quantization is preferable.

jcrist:
Instead of two new options for quantization, how about adding a single `decimal_format` option to `Encoder`? This would take either a string to pass to `quantize` (something like `decimal.quantize(Decimal(decimal_format))`), or a callable that takes in the decimal and returns a new value to encode. A few examples:

```python
# Uses default rounding
enc = Encoder(decimal_format="0.0001")

# Custom rounding
enc = Encoder(decimal_format=lambda d: d.quanitize(decimal.Decimal("0.001"), "ROUND_DOWN"))
```

I like this since it's more flexible, and also only adds a single new option. Otherwise I'd worry about other users needing further customization, resulting in a number of `decimal_*` kwargs.

I wouldn't expect a callable here to have a perf cost - calling into python here is negligible, most of the time will be in the `quantize` call itself.

> Did you consider the Meta-based approach? If so, what made you land on encoder-level? Both have trade-offs and I'd rather get the API right before merge.

In `msgspec`, (currently) encoding doesn't have any type-level information, it only has the values. This means customization for encoding cannot rely on information in annotations, it has to rely on the actual object instances themselves. This is admittedly less flexible in cases where you might want to encode different values differently, but keeps the encoder simple and supports values that exist outside of containers with attached annotations (e.g. `encode(decimal_object)` wouldn't have annotations, but `encode(struct_with_a_decimal_field)` would).

For now a single setting on an `Encoder` is both straightforward to implement, and matches the current conventions.

NyanFisher:
@Siyet @jcrist Hello! Thanks for the review!

@Siyet
> Did you consider the Meta-based approach? If so, what made you land on encoder-level? Both have trade-offs and I'd rather get the API right before merge.

I hadn't considered using `Meta`, but I think that approach would result in a large number of `TypeNode`. I work at a bank and know that a single `Price` isn't enough, since it's too general a concept. But it's a good idea for future 😃 

@jcrist 
> Instead of two new options for quantization, how about adding a single decimal_format option to Encoder? 

I like this idea, but the `decimal_format` [parameter already exists](https://jcristharif.com/msgspec/api.html#msgspec.json.Encoder). If you plan to extend the interface with additional types, I don’t think this is the best solution, as it will confuse users. I suggest using a separate additional parameter called `decimal_quantize` with the types `Decimal | Callable[[Decimal], Decimal]`, which would be responsible exclusively for quantization.
This way, we’ll retain the ability to convert `Decimal` to `“string”`/`“number”`, add quantization, and maintain backward compatibility.

Siyet:
After thinking it through I'm coming around to @jcrist's single-kwarg shape. One slot for everything is, in my view, the right call here.

```python
Encoder(decimal_format=lambda d: d.quantize(Decimal("0.001"), ROUND_DOWN))  # custom
Encoder(decimal_format="string")                                            # existing
Encoder(decimal_format="number")                                            # existing
```

We could split it along `dataclasses.field(default=..., default_factory=...)` lines (value in one kwarg, callable in another), but that split exists specifically to disambiguate "the value *is* a callable" from "call this to produce the value", and neither `"string"` nor `"number"` is callable. Introducing a separate `decimal_hook` just to satisfy a pattern we do not actually need feels like overcomplicating the interface.

There is also the naming angle: `decimal_format` reads as a verb just as naturally as it reads as a noun ("how to format the decimal"), which makes "pass a callable that does the formatting" fit the name rather than fight it.

@NyanFisher regarding your concern about overloading an existing kwarg: the three shapes (`"string"` / `"number"` / callable) dispatch unambiguously on type (string vs. callable), so the dispatch logic in C stays simple and the user-facing docs just enumerate the three accepted shapes in one place.

NyanFisher:
@Siyet

Please review this PR when you have a moment 🙂 I changed the implementation.

src/msgspec/_core.c
```diff
@@ -9567,12 +9570,12 @@ Encoder_init(Encoder *self, PyObject *args, PyObject *kwds)
...
     /* Process decimal format */
+    self->decimal_callable = NULL;
     if (decimal_format == NULL) {
         self->decimal_format = DECIMAL_FORMAT_STRING;
     }
-    else {
+    else if (PyUnicode_CheckExact(decimal_format)) {
         bool ok = false;
-        if (PyUnicode_CheckExact(decimal_format)) {
             if (PyUnicode_CompareWithASCIIString(decimal_format, "string") == 0) {
```
Siyet, line 9735: Indentation broke after flattening the nesting. The inner `if/elif` here are at 12 spaces instead of 8. Same below: bodies of `else if (PyCallable_Check)` (lines 9596-9600) and `else` (lines 9601-9608) are at 8 spaces instead of 4.

src/msgspec/json.pyi
```diff
@@ -23,15 +24,21 @@ schema_hook_sig = Optional[Callable[[type], dict[str, Any]]]
 
 class Encoder:
     enc_hook: enc_hook_sig
```
Siyet: `Callable[[Decimal], Union[str, float]]` is narrower than what the runtime actually accepts. After the callable returns, the value goes through the regular encode path and accepts anything encodable (`int`, `bool`, `dict`, `Struct`, `bytes`, etc.). A realistic case: scaled integer for cents (`lambda d: int(d * 100)`), which mypy rejects with the current stub but works at runtime. Should be `Callable[[Decimal], Any]`. Same for `msgpack.pyi`.

src/msgspec/_core.c
```diff
@@ -13311,11 +13338,27 @@ mpack_encode_decimal(EncoderState *self, PyObject *obj)
...
     }
+    else {
+        temp = PyObject_CallFunctionObjArgs(self->decimal_callable, obj, NULL);
+        if (temp == NULL) return -1;
+
+        PyTypeObject *type = Py_TYPE(temp);
+        if (type == (PyTypeObject *)(self->mod->DecimalType)) {
+            PyErr_SetString(
+                PyExc_TypeError,
+                "decimal_format callable must not return a Decimal"
```
Siyet: The guard only catches a direct `Decimal` return, not a nested one. Verified locally:

```
lambda d: [Decimal("0.5")]              -> RecursionError
lambda d: {"v": Decimal("0.5")}         -> RecursionError
lambda d: Struct(inner=Decimal("0.5"))  -> RecursionError
```

Worse: with `sys.setrecursionlimit(10**6)` (common in projects with deep graphs) the Python-level safety net is gone and the encoder hits SIGSEGV with a core dump. Cleanest fix: add an `in_decimal_callable` flag to `EncoderState`, on re-entry into `*_encode_decimal` raise `TypeError: callable returned a value containing a Decimal`. ~5 lines of C, covers all the nested cases. Same applies to `_core.c:14028` (`json_encode_decimal`).

Siyet (commented):
Re-checked after the rework. CI is green across the matrix, design matches what we landed on. Ran some scenarios beyond the existing tests locally (WSL Ubuntu 22.04, Python 3.10, build at 48102732), three blockers inline.

Plus docs: `docs/supported-types.rst:595-606` only mentions `'string'`/`'number'`, would be good to add a callable example covering the use case from #848.

Nit: `test_encoder_decimal_callable_raise_error_if_fn_return_decimal` should use `match="must not return a Decimal"` to pin the message rather than any `TypeError`.

src/msgspec/_core.c
NyanFisher: Thanks! Fixed it.

src/msgspec/json.pyi
NyanFisher: Thanks! Fixed it.

src/msgspec/_core.c
NyanFisher: Thanks! Fixed it.

NyanFisher:
@Siyet Hello!

I have corrected the comments. Please re-check this PR when you have a moment.

NyanFisher:
Hi @Siyet ! Just gently bumping this pull request. Let me know if there’s anything I can fix or rebase to help get it merged! 🙂

Siyet:
Looks good. A few non-blocking things before merge:

- Needs a rebase (currently conflicting with `main`).
- Test gap: `test_encoder_decimal_callable_convert_value_to_string` and `..._into_different_types` take the `proto` fixture but hardcode `msgspec.json.Encoder`, so the msgpack callable encode path is not actually exercised. Switching them to `proto.Encoder` would cover it.
- Minor: the `Encoder`/`JSONEncoder` docstrings only say "should return a supported object"; worth a line that the result must not contain a `Decimal` (it is in supported-types.rst but not the docstrings). The `decimal_format=1` `ValueError` -> `TypeError` change is also worth a CHANGELOG note.

NyanFisher:
@Siyet, I have corrected the comments. 

> The decimal_format=1 ValueError -> TypeError

I was thinking, maybe we should keep `ValueError` for backward compatibility? I can swap the conditions.

```cpp
if (decimal_format == NULL) {
    ...
}
else if (PyCallable_Check(decimal_format)) {
    ...
}
else {
    bool ok = false;
    if (PyUnicode_CheckExact(decimal_format)) {
        if (PyUnicode_CompareWithASCIIString(decimal_format, "string") == 0) {
            self->decimal_format = DECIMAL_FORMAT_STRING;
            ok = true;
        }
        else if (PyUnicode_CompareWithASCIIString(decimal_format, "number") == 0) {
            self->decimal_format = DECIMAL_FORMAT_NUMBER;
            ok = true;
        }
        if (!ok) {
            PyErr_Format(
                PyExc_ValueError,
                "`decimal_format` must be 'string', 'number', or a callable, got %R",
                decimal_format
            );
            return -1;
        }
    }
}
```

What do you think?

NyanFisher:
@sobolevn, hello! Could you please join the review of this PR? I want to finish it. 🙂

src/msgspec/_core.c
```diff
@@ -9617,30 +9621,42 @@ Encoder_init(Encoder *self, PyObject *args, PyObject *kwds)
...
+        }
+        else if (PyUnicode_CompareWithASCIIString(decimal_format, "number") == 0) {
+            self->decimal_format = DECIMAL_FORMAT_NUMBER;
+            ok = true;
         }
         if (!ok) {
             PyErr_Format(
                 PyExc_ValueError,
-                "`decimal_format` must be 'string' or 'number', got %R",
+                "`decimal_format` must be 'string', 'number', or a callable, got %R",
```
sobolevn, line 9747: Let's not repeat this error twice :)

src/msgspec/_core.c
```diff
@@ -13391,11 +13424,25 @@ mpack_encode_decimal(EncoderState *self, PyObject *obj)
...
+    else if (self->decimal_format == DECIMAL_FORMAT_NUMBER) {
         temp = PyNumber_Float(obj);
         if (temp == NULL) return -1;
         out = mpack_encode_float(self, temp);
     }
+    else {
+        if (self->in_decimal_callable) {
+            PyErr_SetString(
+                PyExc_TypeError,
+                "callable returned a value containing a Decimal"
```
sobolevn: Can we show the value here? It might be helpful for users.

src/msgspec/_core.c
```diff
@@ -14033,27 +14082,45 @@ json_encode_uuid(EncoderState *self, PyObject *obj)
...
 json_encode_decimal(EncoderState *self, PyObject *obj)
 {
-    PyObject *temp = PyObject_Str(obj);
-    if (temp == NULL) return -1;
+    PyObject *temp;
 
-    Py_ssize_t size;
-    const char* buf = unicode_str_and_size_nocheck(temp, &size);
-    bool decimal_as_string = (self->decimal_format == DECIMAL_FORMAT_STRING);
+    if (MS_LIKELY(self->decimal_format != DECIMAL_FORMAT_CALLABLE)) {
```
sobolevn: Let's flip the compare sign. First compare `if (self->decimal_format == DECIMAL_FORMAT_CALLABLE)`, then try to preserve the history of the original code without indenting it. We can use `return` inside `if`, so we can keep existing code untouched.

src/msgspec/msgpack.pyi
```diff
@@ -55,14 +56,15 @@ class Decoder(Generic[_T]):
...
-    decimal_format: Literal["string", "number"]
+    decimal_format: Literal["string", "number"] | Callable[[decimal.Decimal], Any]
     uuid_format: Literal["canonical", "hex", "bytes"]
     order: Literal["deterministic", "sorted"] | None
     def __init__(
         self,
         *,
         enc_hook: _EncHookSig = None,
-        decimal_format: Literal["string", "number"] = "string",
+        decimal_format: Literal["string", "number"]
```
sobolevn:
```suggestion
        decimal_format: (
            Literal["string", "number"]
```

style nit

src/msgspec/msgpack.pyi
```diff
@@ -55,14 +56,15 @@ class Decoder(Generic[_T]):
 @final
 class Encoder:
     enc_hook: _EncHookSig
-    decimal_format: Literal["string", "number"]
+    decimal_format: Literal["string", "number"] | Callable[[decimal.Decimal], Any]
```
sobolevn: You can make a type alias out of this :)

tests/unit/test_common.py
```diff
@@ -3974,13 +3976,111 @@ def test_encoder_decimal_format(self, proto):
...
+        proto,
+        fn: Callable[[decimal.Decimal], Any],
+    ):
+        enc = proto.Encoder(decimal_format=fn)
+        with pytest.raises(
+            TypeError, match="callable returned a value containing a Decimal"
+        ):
+            enc.encode(decimal.Decimal("1"))
+
+    def test_encoder_decimal_callable_raise_error(self, proto):
```
sobolevn, line 4863: Question (maybe for another PR): do we explicitly test `_pydecimal` types? where `decimal.c` was not available / compiled?

sobolevn (commented):
Thanks! Looks really promising!

src/msgspec/_core.c
NyanFisher: Good idea! Thx! Fixed it

src/msgspec/msgpack.pyi
NyanFisher: A new type `_DecimalFormatSig` is used.

src/msgspec/msgpack.pyi
NyanFisher: I created a new type `_DecimalFormatSig`.

src/msgspec/_core.c
NyanFisher: Removed the duplication. 

I have changed the terms to maintain backward compatibility. I wrote about it in a [comment](https://github.com/msgspec/msgspec/pull/978#issuecomment-4722599785).

src/msgspec/_core.c
NyanFisher: I don't see the point. I think that specifying the object will not give any additional information and will only confuse.

tests/unit/test_common.py
NyanFisher: Do you mean such tests?

```python
@pytest.fixture
def force_pydecimal(monkeypatch):
    monkeypatch.delitem(sys.modules, "_decimal", raising=False)
    monkeypatch.delitem(sys.modules, "decimal", raising=False)
    monkeypatch.setitem(sys.modules, "_decimal", None)

    if "decimal" in sys.modules:
        importlib.reload(sys.modules["decimal"])

    yield

    monkeypatch.delitem(sys.modules, "_decimal", raising=False)
    if "decimal" in sys.modules:
        importlib.reload(sys.modules["decimal"])

def test_pydecimal(force_pydecimal):
    import decimal
```

I have never written such tests. 😃

sobolevn: We can do that in a separate PR.

src/msgspec/_core.c
sobolevn: Yes, we should not change the error type here. Otherwise, we can break something for users.

src/msgspec/_core.c
```diff
@@ -14086,7 +14129,27 @@ json_encode_uuid(EncoderState *self, PyObject *obj)
...
-    PyObject *temp = PyObject_Str(obj);
+    PyObject *temp;
+
+    if (self->decimal_format == DECIMAL_FORMAT_CALLABLE) {
+        if (self->in_decimal_callable) {
+            PyErr_SetString(
+                PyExc_TypeError,
+                "callable returned a value containing a Decimal"
+            );
+            return -1;
```
```suggestion
            return ms_decimal_format_error();
```

this can also be an inline function.

sobolevn (commented):
Awesome! Just one comment from me :)

src/msgspec/_core.c
NyanFisher: Thx! Fixed it.

sobolevn (approved):
Great feature! Thank you! Let's keep it open for a while to potentially get more feedback from others.

Siyet (approved):
Second review after @sobolevn's approval, and it's a solid one - LGTM.

What I checked on 8c52acf (deep pass, including an empirical run):

- The recursion guard survived the post-review refactors: a callable returning a `Decimal` (direct or nested arbitrarily deep in dict/list/tuple, json and msgpack alike) raises the clean `TypeError`, never a `RecursionError` or crash. Exceptions from the callable propagate as-is, `KeyboardInterrupt` included.
- Refcounting and GC are clean: `Py_VISIT`/`Py_CLEAR` on the callable are wired into both encoder types, encoder-in-a-cycle-with-its-callable collects, and 600k encodes / 300k error-path iterations / 100k encoder create-destroy cycles all show zero RSS growth.
- All four `EncoderState` creation sites (encode, encode_into, encode_lines, module-level) initialize the new fields; `string`/`number` output is byte-identical to 0.21.1 for both protocols.
- Stubs (`_DecimalFormatSig`) match runtime validation, getter returns the callable by identity, `enc_hook` is not consulted for `Decimal` in any mode.
- All review threads are addressed in code; the one deferred item for the record is `_pydecimal` test coverage, agreed to land as a separate PR.

Two tiny wording nits, fine to fix in a follow-up push or leave as is:

- The msgpack `Encoder` docstring and the `supported-types.rst` note both say the callable "must return a JSON-serializable value" - for msgpack that reads oddly; "a value encodable by the encoder" would cover both.
- The docs warning says returning a `Decimal` "would cause infinite recursion" and then that a `TypeError` is raised - the guard means there's no recursion at all, the second half is what actually happens.

Since this is a feature, the final call is @ofek's per our merge policy - from my side it's ready.

Siyet:
Correcting myself from the review above: I keep forgetting our maintainer pool has grown. With @sobolevn's approval and mine this PR has two maintainer reviews, and the API shape itself was originally suggested by @jcrist - that's plenty for a feature. No need to gate on one specific person, so merging with a clear conscience.

Thanks @NyanFisher for seeing this through!

Issue #30, opened 2021-06-11
Consideration for decode array to `tuple` by default option.


goodboy:
`msgpack-python` [has an option](https://github.com/msgpack/msgpack-python#one-shot-pack--unpack): `use_list=False` to its unpacker to allow for decoding to `tuple` by default.

I noticed [in the docs](https://jcristharif.com/msgspec/#id1) that tuples are only used for array types when used as hashable keys.

Is there a reason there isn't a way to either offer the `tuple`-as-default by a manual flag or, just by default decode to the same type considering they're [ostensibly more performant in python](https://stackoverflow.com/questions/3340539/why-is-tuple-faster-than-list-in-python) then `list`?

jcrist:
Apologies for the delay here. Tuples aren't more performant than lists to create or use. If you read through the answers in the link above, you'd see that only _constant_ tuples (e.g. `(1, 2, 3)`) are "faster" since they're built only once by the compiler. Both lists and tuples have similar representations in cpython, and take equivalent time to construct dynamically. A quick benchmark using msgspec:

```python
In [5]: data = list(range(1000))

In [6]: dec_list = msgspec.Decoder(list)

In [7]: dec_tuple = msgspec.Decoder(tuple)

In [8]: buf = msgspec.encode(data)

In [9]: %timeit dec_list.decode(buf)
12.9 µs ± 20.8 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

In [10]: %timeit dec_tuple.decode(buf)
12.9 µs ± 17.3 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```

I'm not enthused about adding a `use_list`-like option. Lists are the natural default type for MessagePack's array type. If you want to use a different type for arrays then you likely have a schema you're following and I'd direct you to use msgspec's support for typed serialization.

goodboy:
@jcrist learn something new every time I report something here 🏄🏼 

You'd think i would have double checked the `tuple` create speed claim 🙄
Is it possible the `.encode()` step here is faster though? 

Honestly, keeping it as is works for me as simpler is always better imo.
I can close this is no one else is going to have quiffs.

> and I'd direct you to use msgspec's support for typed serialization.

Yeah i think focusing on a struct schema is really the right way to designing things anyway 👍🏼

jcrist:
No problem, happy to help.

> Is it possible the .encode() step here is faster though?

[...]

PR #980, opened 2026-02-22
Fix NamedTuple hash returning 0 on Python 3.14+

### LINKED ISSUE #967: NamedTuple hash changes during round trip msgspec deserialisation in python 3.14
### Description

Hey, 
I saw this behaviour in msgspec 0.20.0 and Python 3.14.1.
NamedTuples have hash changes after being serialised, and deserialised with msgpack- the hash becomes 0.
This means for msgspec structs where NamedTuples are used as dictionary keys, there is unexpected KeyErrors when accessing data 

Here is a minimal examples of the issue-  the following code throws no AssertionErrors in python 3.11-3.13

```
from typing import NamedTuple
import msgspec

class _TestKey(NamedTuple):
    """Test NamedTuple for demonstrating the msgspec Python 3.14 bug."""

    value: int


class _TestStruct(msgspec.Struct, array_like=True):
    """Test Struct that contains a dict with NamedTuple keys."""

    data: dict[_TestKey, str]


def test_msgspec_namedtuple_hash_bug():
    original = _TestKey(value=42)
    original_hash = hash(original)

    # Round-trip through msgspec - hash is broken in Python 3.14
    encoded = msgspec.msgpack.encode(original)
    decoded = msgspec.msgpack.decode(encoded, type=_TestKey)

    # These should be equal
    assert original == decoded, "NamedTuples should be equal after round-trip"

    # This is the bug: hash should be equal but isn't in Python 3.14
    decoded_hash = hash(decoded)
    assert decoded_hash == original_hash, (
        f"Hash mismatch after msgspec round-trip: original={original_hash}, decoded={decoded_hash}. "
        f"In Python 3.14, msgspec-created NamedTuples have hash=0."
    )

    # Dictionary lookup fails due to hash mismatch
    test_dict = {original: "value"}
    assert decoded in test_dict, "Decoded key should be found in dictionary (requires matching hash)"


def test_msgspec_struct_with_namedtuple_dict_key_bug():
    """
    Demonstrates the msgspec Python 3.14 bug when a NamedTuple is used as a dictionary
    key inside a msgspec.Struct - this mirrors the real-world usage in CalculationMetadata
    where SourceMarketKey is used as a key in base_inputs.

    After msgpack deserialization:
    - The Struct is correctly deserialized
    - The dict appears to contain the expected key
    - But dict equality fails because the deserialized keys have hash=0
    - Looking up the key also fails due to hash mismatch
    """
    key = _TestKey(value=42)
    original = _TestStruct(data={key: "test_value"})

    # Round-trip through msgspec
    encoded = msgspec.msgpack.encode(original)
    decoded = msgspec.msgpack.decode(encoded, type=_TestStruct)

    # Get the decoded key for hash comparison
    decoded_key = next(iter(decoded.data.keys()))
    original_key_hash = hash(key)
    decoded_key_hash = hash(decoded_key)

    # The struct should be equal - this is the assertion that fails
    assert original == decoded, (
        f"Struct with NamedTuple dict keys should be equal after round-trip. "
        f"This fails because dict comparison fails due to hash mismatch on keys. "
        f"original_key_hash={original_key_hash}, decoded_key_hash={decoded_key_hash}"
    )

    # Verify the key can be looked up in the decoded struct's dict
    assert key in decoded.data, (
        f"Original key should be found in decoded dict. "
        f"original_key_hash={original_key_hash}, decoded_key_hash={decoded_key_hash}"
    )
    assert decoded.data[key] == "test_value", "Value should be retrievable with original key"

```

worksbyfriday:
## Summary

Fixes #967.

In Python 3.14, CPython added hash caching to tuples ([gh-131525](https://github.com/python/cpython/issues/131525), [PR #131529](https://github.com/python/cpython/pull/131529)). `PyTupleObject` gained an `ob_hash` field initialized to `-1` ("not yet computed"). When `tuple_hash()` sees `ob_hash != -1`, it returns the cached value immediately.

msgspec allocates NamedTuples via `tp_alloc`, which zero-initializes memory. This sets `ob_hash` to `0` — a valid cached hash value — so `hash()` returns `0` for all msgspec-deserialized NamedTuples without ever computing the real hash. This breaks dictionary lookups and set membership when deserialized NamedTuples are used as keys.

The fix adds `MS_TUPLE_RESET_HASH(op)` — a macro that sets `ob_hash = -1` on Python 3.14+ (no-op on earlier versions) — immediately after `tp_alloc` in all three NamedTuple construction sites:

- `mpack_decode_namedtuple` (msgpack decoder)
- `json_decode_namedtuple` (JSON decoder)
- `convert_seq_to_namedtuple` (convert)

## Test plan

- Added `test_namedtuple_hash_preserved_after_roundtrip` in `test_common.py` (covers msgpack + JSON)
- Added `test_namedtuple_hash_preserved` in `test_convert.py` (covers convert path)
- Both tests verify `hash(original) == hash(decoded)` and dict lookup with decoded keys
- All 32 existing NamedTuple tests pass on Python 3.12 (macro is no-op)
- The fix is effective on Python 3.14+ where `ob_hash` exists in `PyTupleObject`

### DIFF
--- a/tests/unit/test_common.py
+++ b/tests/unit/test_common.py
@@ -2492,7 +2492,6 @@ class TestNamedTuple:
             assert "`$[1][0]`" in str(rec.value)
             assert "Expected `int`, got `str`" in str(rec.value)
 
-
     def test_namedtuple_hash_preserved_after_roundtrip(self, proto):
         """Hash of deserialized NamedTuple matches hash of original.
 


Siyet:
Ported to the community fork [msgspec-arise](https://github.com/Siyet/msgspec-arise): [PR #30](https://github.com/Siyet/msgspec-arise/pull/30)

PR #984, opened 2026-03-03
Add `uniqueItems` to `set` and `frozenset` json schema

sobolevn:
While working on `django-modern-rest`, I've noticed that `uniqueItems` is missing for `set` and `frozenset` in json schema. 

Docs: https://json-schema.org/understanding-json-schema/reference/array#uniqueItems

I think that this is quite easy to add and has better semantics.

### DIFF
--- a/.github/workflows/profile.yml
+++ b/.github/workflows/profile.yml
@@ -26,8 +26,6 @@ jobs:
           archs: arm64
         - os: macos-15-intel
           archs: x86_64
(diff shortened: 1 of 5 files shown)


sobolevn:
These two failures do not seem related, they also happen on https://github.com/jcrist/msgspec/pull/983

sobolevn:
Now it passes, thanks!

PR #1003, opened 2026-04-08
Use safe_load in yaml

EhteshamSid:
`yaml.load()` in `src/msgspec/yaml.py` doesn't pass a safe Loader. This can deserialize arbitrary Python objects and is an RCE risk if the YAML comes from user input or the network. Switched to `yaml.safe_load()`.

### DIFF
--- a/src/msgspec/yaml.py
+++ b/src/msgspec/yaml.py
@@ -178,7 +178,7 @@ def decode(buf, *, type=Any, strict=True, dec_hook=None):
         # call `memoryview` first, since `bytes(1)` is actually valid
         buf = bytes(memoryview(buf))
     try:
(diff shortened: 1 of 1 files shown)


src/msgspec/yaml.py
```diff
@@ -178,7 +178,7 @@ def decode(buf, *, type=Any, strict=True, dec_hook=None):
         # call `memoryview` first, since `bytes(1)` is actually valid
         buf = bytes(memoryview(buf))
     try:
-        obj = yaml.load(buf, Loader)
+        obj = yaml.safe_load(buf, Loader)
```
jcrist, line 181: The `Loader` provided here is a configured safe loader. It's equivalent to `yaml.safe_load`.

PR #1004, opened 2026-04-09
Support Literal[True] and Literal[False] types

### LINKED ISSUE #859: Support Literal[bool]
### Description

Many APIs use something like `{"success": true, data: ...}`.

It will be really nice to be able to validate such boolean field against constant value when deserializing:

```python
from typing import Literal

import msgspec


class ResponseData(msgspec.Struct):
    pass


class Response(msgspec.Struct):
    success: Literal[True]
    data: ResponseData


response = msgspec.convert(
    {
        "success": True,
        "data": {},
    },
    Response,
)
```

At the moment this raises:
```
TypeError: Literal may only contain None/integers/strings - typing.Literal[True] is not supported
```

Thank you 👍

Siyet:
Reopening #996 (auto-closed when fork was deleted).

## Summary
- Add support for `Literal[True]`, `Literal[False]`, and `Literal[True, False]` in type definitions
- Introduce `MS_TYPE_BOOLLITERAL_TRUE` / `MS_TYPE_BOOLLITERAL_FALSE` flags to distinguish bool literals from plain `bool`
- Handle `PyBool_Type` before `PyLong_Type` in `typenode_collect_literal` (since `bool` is a subclass of `int`)
- Update decode paths (JSON, msgpack, convert) to validate against specific bool literal values
- Update `inspect.py` `LiteralType.values` type annotation to include `bool`

Closes #859

### DIFF
--- a/docs/supported-types.rst
+++ b/docs/supported-types.rst
@@ -1143,6 +1143,7 @@ purpose, but with a `typing.Literal` the decoded values are literal `int` or
 A literal can be composed of any of the following objects:
 
 - `None`
+- `bool` values (`True` and `False`)
 - `int` values
 - `str` values
 - Nested `typing.Literal` types
@@ -1169,6 +1170,12 @@ values, or doesn't match any of their component types.
     Traceback (most recent call last):
       File "<stdin>", line 1, in <module>
     msgspec.ValidationError: Expected `int`, got `str`
+
+    >>> msgspec.json.decode(b'true', type=Literal[True])
+    True
+
+    >>> msgspec.json.decode(b'false', type=Literal[True, False])
+    False
 
 ``NewType``
 -----------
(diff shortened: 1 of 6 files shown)


ofek (approved):
Nice, thanks!

Siyet:
Two follow-ups pushed:

- Docs (`docs/supported-types.rst`) - added bool to the list of supported Literal value types, plus a couple of examples.

- Error messages (`src/msgspec/_core.c`) - while testing the doc examples I noticed validation errors for `Literal[True]` / `Literal[False]` produced messages with an empty type name (Expected '', got bool). Fixed all three decode paths (`mpack_decode_bool`, `json_decode_true/false`, `convert_bool`) to emit `Invalid enum value <val>`, consistent with how int/str literals are reported. Test added in `test_literal_bool_error_message`.

Side note: the Invalid enum value wording is misleading for Literal types in general since they are not enums. I opened #1009 to track that and submitted #1010 (based on this branch) to fix it across all Literal kinds at once.

Siyet:

PR #1007, opened 2026-04-09
Add porting guide for users coming from orjson

### LINKED ISSUE #642: Porting guide for users coming from `orjson`
msgspec's performance is generally on par with that of `orjson`. For some schemas we're faster, for some `orjson` is faster. For common message schemas though things should be about the same.

There are some valid reasons a user might prefer to migrate from `orjson` to `msgspec.json` instead:

- `msgspec` also can handles schema validation using python type hints, similar to (but faster than) `pydantic`. Adding types also generally [makes parsing _faster_](https://jcristharif.com/msgspec/benchmarks.html#json-serialization), while providing stronger guarantees to the user.
- A wide set of supported [builtin types](https://jcristharif.com/msgspec/supported-types.html).
- Possibly more reliable. Some recent `orjson` releases have led to segfaults for some users (see e.g. https://github.com/ijl/orjson/issues/452). While in rust, `orjson` is mostly a shim around the `yyjson` C library, and makes use of `unsafe` in several places, which can let memory safety issues sneak through. `msgspec` is all in C so we're not necessarily better :grimacing:, but we do seem to have a lighter [history of segfault bug reports](https://github.com/ijl/orjson/issues?q=is%3Aissue+segfault+is%3Aclosed).

For most users that aren't passing additional config options to `orjson`, porting should be as straightforward as swapping calls to `orjson.loads` to `msgspec.json.decode` and `orjson.dumps` to `msgspec.json.encode`. For other options we generally have a corollary, but the spelling is different. A short porting/comparison guide might be useful here.

Siyet:
Reopening #995 (auto-closed when fork was deleted).

## Summary
- Add `docs/porting-from-orjson.rst` covering API mapping, option equivalents, typed decoding, error handling, and features without direct equivalent
- Add page to toctree in `docs/index.rst`

Closes #642

Note: @ofek suggested restructuring this into a dedicated "Porting" docs section to accommodate future guides (e.g. Pydantic). Happy to rework if that's the preferred approach.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -201,6 +201,7 @@ few:
     extending.rst
     inspect.rst
     perf-tips.rst
+    porting-from-orjson.rst
 
 .. toctree::
     :hidden:
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -201,7 +201,13 @@ few:
     extending.rst
     inspect.rst
     perf-tips.rst
-    porting-from-orjson.rst
+
+.. toctree::
+    :hidden:
+    :maxdepth: 2
+    :caption: Porting
+
+    porting/index.rst
 
 .. toctree::
     :hidden:
--- a/docs/porting-from-orjson.rst
+++ b/docs/porting-from-orjson.rst
@@ -1,0 +1,322 @@
+Porting from orjson
+===================
+
+If you're currently using orjson_ and considering migrating to ``msgspec``,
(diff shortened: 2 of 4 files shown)


ofek:
It looks like Pydantic restructured their site to have documentation in the following format, one for each of their libraries:

```
https://pydantic.dev/docs/(validation|ai|logfire)/
```

Siyet:

PR #1008, opened 2026-04-10
Update Pydantic doc links to new site structure

Siyet:
Pydantic restructured their docs site to `https://pydantic.dev/docs/(validation|ai|logfire)/`, breaking the existing links and causing the link checker job to fail on `main`.

This updates all references:
- `https://docs.pydantic.dev/latest/` → `https://pydantic.dev/docs/validation/latest/get-started/`
- `https://docs.pydantic.dev/latest/usage/dataclasses/` → `https://pydantic.dev/docs/validation/latest/concepts/dataclasses/`

Files updated: `README.md`, `docs/benchmarks.rst`, `docs/index.rst`, `docs/usage.rst`, `docs/supported-types.rst`, `docs/structs.rst`.

PR #1010, opened 2026-04-10
Use 'Invalid value' wording for Literal and Enum validation errors

### LINKED ISSUE #1009: Validation errors for Literal types incorrectly say 'enum value'
## Description

When validation fails for a `typing.Literal` type, msgspec raises a `ValidationError` with the message `Invalid enum value <val>`. This is misleading because the user is not using `enum.Enum`, they are using `typing.Literal`.

## Reproduction

```python
import msgspec
from typing import Literal

msgspec.json.decode(b'4', type=Literal[1, 2, 3])
# msgspec.ValidationError: Invalid enum value 4

msgspec.json.decode(b'"bad"', type=Literal["one", "two"])
# msgspec.ValidationError: Invalid enum value 'bad'

msgspec.json.decode(b'false', type=Literal[True])
# msgspec.ValidationError: Invalid enum value False
```

## Expected

A wording that does not reference `enum`, since the type is not an enum. For example:

```
Invalid value 4
Invalid value 'bad'
Invalid value False
```

Or, even more useful, a message that lists the allowed literals:

```
Invalid value 4, expected one of (1, 2, 3)
```

## Notes

- This affects all decode paths (JSON, MsgPack, `convert`).
- The same error path (`_Lookup_OnMissing`) is shared between `Enum` and `Literal`, so the fix needs to differentiate between the two cases (when `lookup->cls != NULL` it is an `Enum`, otherwise it is a `Literal`).
- Bool literals (added in #1004) use a separate code path but produce the same misleading wording.

Siyet:
Closes #1009.

Validation errors for `typing.Literal` types previously used the wording `Invalid enum value <val>`, which is misleading because the user is not using `enum.Enum` (see #1009).

Following the discussion below, this uses a single universal wording `Invalid value <val>` for both `Literal` and `Enum` (and the bool literals from #1004), rather than splitting the two by type. The target type is never named in the message, which resolves the "arbitrary distinction" concern and stays consistent with the existing union tag errors that already raise plain `Invalid value`.

Before:

```
Literal[1, 2, 3]   + 4       -> Invalid enum value 4
Fruit (Enum)       + "grape"  -> Invalid enum value 'grape'
```

After:

```
Literal[1, 2, 3]   + 4       -> Invalid value 4
Fruit (Enum)       + "grape"  -> Invalid value 'grape'
```

Tests and docs updated; full unit suite passes (6368 passed).

### DIFF
--- a/docs/supported-types.rst
+++ b/docs/supported-types.rst
@@ -1112,7 +1112,7 @@ doesn't match any valid member.
     >>> msgspec.json.decode(b'"grape"', type=Fruit)
     Traceback (most recent call last):
       File "<stdin>", line 1, in <module>
-    msgspec.ValidationError: Invalid enum value 'grape'
+    msgspec.ValidationError: Invalid value 'grape'
 
     >>> class JobState(enum.IntEnum):
     ...     CREATED = 0
@@ -1129,7 +1129,7 @@ doesn't match any valid member.
     >>> msgspec.json.decode(b'4', type=JobState)
     Traceback (most recent call last):
       File "<stdin>", line 1, in <module>
-    msgspec.ValidationError: Invalid enum value 4
+    msgspec.ValidationError: Invalid value 4
 
 If the enum type includes a ``_missing_`` method (`docs
 <https://docs.python.org/3/library/enum.html#enum.Enum._missing_>`__), this
@@ -1162,7 +1162,7 @@ this is supporting case-insensitive enums:
     >>> msgspec.json.decode(b'"grape"', type=Fruit)
     Traceback (most recent call last):
       File "<stdin>", line 1, in <module>
-    msgspec.ValidationError: Invalid enum value 'grape'
+    msgspec.ValidationError: Invalid value 'grape'
 
 ``Literal``
 -----------
@@ -1196,7 +1196,7 @@ values, or doesn't match any of their component types.
     >>> msgspec.json.decode(b'4', type=Literal[1, 2, 3])
     Traceback (most recent call last):
       File "<stdin>", line 1, in <module>
-    msgspec.ValidationError: Invalid enum value 4
+    msgspec.ValidationError: Invalid value 4
 
     >>> msgspec.json.decode(b'"bad"', type=Literal[1, 2, 3])
     Traceback (most recent call last):
--- a/src/msgspec/_core.c
+++ b/src/msgspec/_core.c
@@ -884,7 +884,10 @@ _Lookup_OnMissing(Lookup *lookup, PyObject *val, PathNode *path) {
     }
     PyObject *suffix = PathNode_ErrSuffix(path);
     if (suffix != NULL) {
-        PyErr_Format(mod->ValidationError, "Invalid enum value %R%U", val, suffix);
+        const char *format = (lookup->cls != NULL)
+            ? "Invalid enum value %R%U"
+            : "Invalid value %R%U";
+        PyErr_Format(mod->ValidationError, format, val, suffix);
         Py_DECREF(suffix);
     }
 
@@ -15366,7 +15369,7 @@ mpack_decode_bool(DecoderState *self, PyObject *val, TypeNode *type, PathNode *p
         return Py_False;
     }
     if (type->types & (MS_TYPE_BOOLLITERAL_TRUE | MS_TYPE_BOOLLITERAL_FALSE)) {
-        ms_raise_validation_error(path, "Invalid enum value %R%U", val);
+        ms_raise_validation_error(path, "Invalid value %R%U", val);
         return NULL;
     }
     return ms_validation_error("bool", type, path);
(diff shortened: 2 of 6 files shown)


Siyet:

jcrist:
The implementation seems fine, so happy to merge. However, I'm not _100%_ sure this is the right move.

`msgspec` throws the same errors everywhere, regardless of decoding tool (`msgspec.json.decode`, `msgspec.msgpack.decode`, `msgspec.convert`, ...). To make the validation errors as universal as possible, I decided early on to use syntax that would make sense to a consumer of a JSON API (e.g. refer to JSON types and schema requirements) over things that more reveal implementation details (e.g. python class names). This is of course tricky to do universally, but was a general guideline.

JSON schema refers to collections of possible values as "enums", hence the use of "enum" in the error here. In Python of course these can be represented as a `enum.Enum` or as a `Literal`, but to a JSON API consumer that's not relevant information.

I don't think dropping the "enum" in the error messages makes things more confusing (and still doesn't leak python implementation details), so if you think what you have here is clearer than I'm happy with it. Just sharing context for why things were the way they were before. Up to you.

Siyet:
Thanks for the context. The push toward "Invalid value" came down to one thing: `enum` is overloaded. JSON Schema readers parse it as the schema keyword (your intent), but most msgspec users read it as `enum.Enum` and get confused when their `Literal[...]` field reports `Invalid enum value`. That's how #1009 was filed.

It also gets ambiguous when both coexist: a struct with a `Color` (`Enum`) field and a `Literal["a", "b"]` field today produces identical wording for both. Splitting them ("enum value" for real `Enum`, "value" for `Literal`) disambiguates without leaking Python details, since `Literal` never appears in the message. Real `Enum` errors are unchanged, so JSON Schema consumers aren't affected.

provinzkraut:
> It also gets ambiguous when both coexist: a struct with a `Color` (`Enum`) field and a `Literal["a", "b"]` field today produces identical wording for both. Splitting them ("enum value" for real `Enum`, "value" for `Literal`) disambiguates without leaking Python details, since `Literal` never appears in the message. Real `Enum` errors are unchanged, so JSON Schema consumers aren't affected.

This would introduce a different ambiguity though: If it refers to an `enum.Enum`, it's explicitly named in the error, if it's for a `Literal`, it's vague. This does feel like an arbitrary distinction, so IMO either the target type shouldn't be mentioned in any case, or it should be mentioned in every case.

Siyet:
Fair point, the split is arbitrary. I went through the alternatives:

**Mention the type in every case.** The only candidate term for Literal is "literal", but that's Python-specific (`typing.Literal`). JSON Schema has no separate keyword for it: both Enum and Literal map to the `enum` keyword. Using "Invalid literal value" would leak implementation details, which goes against the error message design. No existing message mentions Python type names (no "dataclass", "TypedDict", "NamedTuple", etc.).

**Qualify as "JSON Schema enum".** Doesn't work either: msgspec errors are protocol-agnostic (same message for JSON, MsgPack, `convert`). Referencing "JSON Schema" in a MsgPack decode error would be more confusing, not less.

**Include the field name in the message body.** Already covered by the path suffix (`- at $.field`), so redundant.

That leaves **dropping "enum" from both**, using "Invalid value" universally. This is consistent (no type qualifier in either case), doesn't leak implementation details, and matches the existing union tag errors which already use plain "Invalid value".

Bigger change than the current PR since it touches Enum errors too, but it's the only option that resolves the inconsistency cleanly. I can update the branch if you and @jcrist are fine with this direction.

Siyet:
Updated the branch to use `Invalid value` universally for both `Enum` and `Literal`, dropping the earlier split.

@provinzkraut this addresses the arbitrary-distinction point: the target type is no longer named in either case, and it matches the existing union tag errors which already raise plain `Invalid value`.

@jcrist heads up that this also changes the real `Enum` messages (`Invalid enum value` -> `Invalid value`), so it does step on the JSON Schema "enum" wording you described above. My read is that dropping "enum" still doesn't leak Python details and isn't less clear for a JSON API consumer, while it buys consistency across Enum, Literal, and union tags. Since it touches wording you chose deliberately, wanted your explicit ok before merging.

Rebased on main, tests and docs updated, full unit suite green (6368 passed).

ofek:
Have we considered the word "choice"? I think that fits quite nicely and at the very least removes ambiguity.

Siyet:
"choice" does read more precisely than "value". My hesitation is scope: to get consistent wording everywhere we'd also have to change the union tag errors (they already say `Invalid value`) and any other "not in the allowed set" message, so more error strings shift, and each one is something a user could be matching on literally. ZeroVer lets us land that kind of break in a minor, but churning error strings that don't need to change is a poor trade for a marginal wording gain.

`Invalid value` keeps the change contained and already lines enum/literal up with the existing tag errors, so I lean that way unless we deliberately want the broader rename. Not attached to it if you and @jcrist would rather standardize on "choice" everywhere.

Siyet:
Merged current main in, no content change.

@ofek on "choice" against "value", with the actual numbers. `_core.c` has exactly two user-visible strings for this condition: `Invalid enum value <x>` on 5 call sites (Enum, Literal, and the bool literals) and `Invalid value <x>` on 5 call sites, all of them union tag lookups. "value" changes one of the two, because the tag errors already say `Invalid value`. "choice" changes both, so it doubles the user-visible breakage and moves about 40 more test assertions. That is the whole of my argument, it is not bigger than that, and I am fine either way: if you prefer "choice", I will do the full rename here.

@provinzkraut @sobolevn this changes wording @jcrist picked deliberately and he has not been reachable since June, so I would rather have one collaborator sign off than merge it on two months of silence.

PR #1011, opened 2026-04-10
Clarify wording of order='deterministic' in docstrings

### LINKED ISSUE #825: Ordering of dict keys when using `msgspec.json.encode`
### Question

When using `msgspec.json.encode` for a dictionary in Python 3.12, for example:

```python
msgspec.json.encode({"b": 1, "a": 2})
```

I see that the ordering of the keys is preserved (with the default `order=None`). The above example returns `b'{"b":1,"a":2}'`. Is this order guaranteed to be the same between runs?

Siyet:
Closes #825.

Dicts in Python 3.7+ are ordered by insertion order per the language spec, so describing them as "unordered collections" in the `order='deterministic'` docstring was technically incorrect and confusing (as raised in #825).

The actual reason `order='deterministic'` sorts dict keys is not "consistency between runs" (insertion order is already stable within a run), but to ensure that dicts which compare equal produce identical encoded output, regardless of how they were constructed:

```python
a = {"b": 1, "a": 2}
b = {"a": 2, "b": 1}
assert a == b  # True, but without sorting they'd encode differently
```

Reword the bullet across all encoder docstrings (`json.encode`, `msgpack.encode`, `yaml.encode`, `toml.encode`, `Encoder`, `JSONEncoder`, `convert`) to make this clearer. No behavior change.


jcrist (approved):

jcrist:
Thanks!

PR #1013, opened 2026-04-10
fix wrapping `ValidationError` and `DecodeError` in `ValidationError` in `dec_hook` #1012

### LINKED ISSUE #1012: Exceptions inheriting from `ValidationError` raised in `dec_hook` implicitly wrapped in another `ValidationError` post `0.21.0`
### Description

Pre `0.21.0`, this would raise `SomeError` directly, now it raises `msgspec.ValidationError` with `SomeError` as a `cause`:

```python
import msgspec

class SomeError(msgspec.ValidationError):
    pass

class Foo:
    pass

def dec_foo(tp, val):
    raise SomeError()

try:
    msgspec.convert("", type=Foo, dec_hook=dec_foo)
except Exception as exc:
    print(type(exc))
```

I believe this was an unintended side-effect of #790. 

I'm not sure if this is desirable or not. It is however a breaking change

provinzkraut:
Fix #1012.

Fix a regression introduced in #790 that would cause `ValidationError` and `DecodeError` raise in `dec_hook` to be wrapped in `ValidationError` again.

### DIFF
--- a/src/msgspec/_core.c
+++ b/src/msgspec/_core.c
@@ -5285,18 +5285,11 @@ ms_maybe_wrap_validation_error(PathNode *path) {
     MsgspecState *mod = msgspec_get_global_state();
 
     /* If it's a TypeError or ValueError, wrap it in a ValidationError.
-     * Otherwise we reraise the original error below
-     * Special case ValidationError and DecodeError which are also reraised, even though
-     * they inherit from ValueError. */
+     * Otherwise we reraise the original error below.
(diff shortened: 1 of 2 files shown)


jcrist:
Thanks for this! Going to push a quick fixup, but this is _almost_ good to go.

jcrist (approved):

provinzkraut:
Ah, good catch with the unnecessary check. I initially only checked for `ValidationError` and only added `DecodeError` later. Didn't backtrack to think if this still makes sense. Thanks!

PR #1016, opened 2026-04-12
fix `NULL` check in  `structmeta_get_module_ns`

provinzkraut:
Fix a bug in `structmeta_get_module_ns` that could theoretically lead to a segfault due to a missing `NULL` check, even though it's quite unlikely to ever hit that path


Siyet (approved):
Nice catch, thanks Janek!

PR #1019, opened 2026-04-12
Fix TypeError in type_info for mixed-type Literals

### LINKED ISSUE #1018: inspect.type_info crashes on mixed-type Literals
\`msgspec.inspect.type_info\` crashes with \`TypeError\` when called on a \`Literal\` type that mixes value types (e.g. \`Literal[1, None]\` or \`Literal[True, \"yes\"]\`).

## Reproduction

```python
import msgspec.inspect as mi
from typing import Literal

mi.type_info(Literal[1, None])
# TypeError: '<' not supported between instances of 'NoneType' and 'int'

mi.type_info(Literal[True, "yes"])
# TypeError: '<' not supported between instances of 'str' and 'bool'
```

## Cause

\`inspect.py:888\` does \`LiteralType(tuple(sorted(args)))\` which calls \`sorted()\` on the Literal args. Python 3 does not allow comparison between different types, so mixed-type tuples like \`(1, None)\` or \`(True, "yes")\` raise \`TypeError\`.

Additionally, \`LiteralType.values\` type annotation (\`Union[Tuple[bool, ...], Tuple[str, ...], Tuple[int, ...]]\`) does not cover mixed-type tuples.

## Suggested fix

1. Replace \`sorted(args)\` with a type-aware sort (e.g. \`sorted(args, key=lambda x: (type(x).__name__, x))\`) or skip sorting for mixed types.
2. Widen \`LiteralType.values\` to \`Tuple[Union[bool, str, int, None], ...]\` or similar.
3. Add tests for \`type_info(Literal[True])\`, \`type_info(Literal[True, "yes"])\`, \`type_info(Literal[1, None])\`.

This is a pre-existing issue (not introduced by #1004), but #1004 makes it more likely to encounter since it enables \`Literal[True, ...]\` combinations.

lawrence3699:
Closes #1018

`inspect.type_info()` crashes with `TypeError` when called on `Literal` types that mix value types:

```python
import msgspec.inspect as mi
from typing import Literal

mi.type_info(Literal[1, None])
# TypeError: '<' not supported between instances of 'NoneType' and 'int'
```

**Cause:** `sorted(args)` on line 888 fails in Python 3 when args contain values of different types (e.g. `int` and `NoneType`).

**Fix:** Use a type-aware sort key `("", 0) if x is None else (type(x).__name__, x)` that groups values by type name, with `None` always sorting first. Also updates the `LiteralType.values` annotation to `Tuple[Union[None, int, str], ...]` since mixed-type tuples are valid.

**Before:** `type_info(Literal[1, None])` raises `TypeError`
**After:** `type_info(Literal[1, None])` returns `LiteralType(values=(None, 1))`

Existing behavior for same-type Literals is unchanged (e.g. `Literal[3, 1, 2]` still produces `(1, 2, 3)`).

**Validation:** all 168 inspect tests pass (including 4 new regression tests), ruff lint + format clean.

### DIFF
--- a/src/msgspec/inspect.py
+++ b/src/msgspec/inspect.py
@@ -315,11 +315,11 @@ class LiteralType(Type):
     Parameters
     ----------
     values: tuple
-        A tuple of possible values for this literal instance. Only `str` or
-        `int` literals are supported.
+        A tuple of possible values for this literal instance. Values may be
+        ``None``, ``int``, or ``str`` literals.
     """
 
-    values: Union[Tuple[str, ...], Tuple[int, ...]]
+    values: Tuple[Union[None, int, str], ...]
 
 
 class CustomType(Type):
@@ -885,7 +885,14 @@ class _Translator:
             args = tuple(self.translate(a) for a in args if a is not _UnsetType)
             return args[0] if len(args) == 1 else UnionType(args)
         elif t is Literal:
-            return LiteralType(tuple(sorted(args)))
+            return LiteralType(
+                tuple(
+                    sorted(
+                        args,
+                        key=lambda x: ("", 0) if x is None else (type(x).__name__, x),
+                    )
+                )
+            )
         elif _is_enum(t):
             return EnumType(t)
         elif is_struct_type(t):
--- a/tests/unit/test_inspect.py
+++ b/tests/unit/test_inspect.py
@@ -350,6 +350,25 @@ def test_str_literal():
     assert mi.type_info(Literal["c", "a", "b"]) == mi.LiteralType(("a", "b", "c"))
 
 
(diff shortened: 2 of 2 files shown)


Siyet (commented):
Tested on a Linux VPS (Python 3.12, Ubuntu): all 168 inspect tests green, including the 4 new ones. Also ran 12 extra edge-case scenarios (negative ints, single-element Literals, multi-type mixes), all pass.

Two things to address:

1. +1 to Copilot's inline: confirmed on VPS that `msgspec.json.schema(Literal[1, None])` crashes with `TypeError`.
2. Sort order (None first, then by type name, then by value) is not documented. A comment on the lambda would help.

PR #1021, opened 2026-04-13
fix leak in `typenode_collect_literal`

provinzkraut:
Fix leak caused by a missing `Py_DECREF` on an error path in `typenode_collect_literal`. Only possible on Python <3.12, as empty tuples are immortalized from 3.12 onwards.


jcrist (approved):
Thanks!

Just noting severity for anyone following along - this leak would only ever leak references to an empty tuple (an immortal singleton on all python versions, effectively leaking nothing). It also only shows up in the case of a `Literal` with no arguments, which is already a user-error.

Not saying we shouldn't fix it! Some people just associate leaks like this with major issues, and this one is very very minor.

provinzkraut:
Yeah, it's virtually impossible to hit, and completely irrelevant in 3.12+

PR #1023, opened 2026-04-13
fix reference leak in `ms_decode_bigint`

provinzkraut:
Fix a reference leak in `ms_decode_bigint`, caused by improper handling of referenced obtained via `PyErr_Fetch`

### DIFF
--- a/src/msgspec/_core.c
+++ b/src/msgspec/_core.c
@@ -10222,7 +10222,8 @@ ms_decode_bigint(const char *buf, Py_ssize_t size, TypeNode *type, PathNode *pat
             goto out_of_range;
         }
         else {
-            /* Restore the exception state */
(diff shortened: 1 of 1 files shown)


provinzkraut:
Last one of this series for now. I'll go look at something else :)

src/msgspec/_core.c
jcrist, line 10227: Can you also modify the comment here noting that `PyErr_Restore` steals a reference to all arguments? Had to look this up just to make sure this path didn't need a decref too.

jcrist (commented):
Thanks!

src/msgspec/_core.c
provinzkraut: Done

jcrist (approved):
Thanks!

PR #1024, opened 2026-04-18
implement multi-phase init

### LINKED ISSUE #563: Use multi-phase init for c-extension initialization
### Description

`msgspec` currently uses the old style single-phase init for module import. We should transition to use the newer multi-phase init process. Besides code quality, this is one of the steps required for us to make use of the new [per-interpreter GIL subinterpreters](https://peps.python.org/pep-0684/) feature in CPython 3.12.

Some relevant resources:
- https://peps.python.org/pep-0489/
- https://docs.python.org/3.12/howto/isolating-extensions.html
- https://peps.python.org/pep-0573/

provinzkraut:
Closes #563.

Implement [multi-phase init](https://docs.python.org/3.13/c-api/module.html#multi-phase-initialization).

Relatively straightforward, the only slight hiccough is `msgspec_get_global_state`, which relied on `PyState_FindModule` / `PyState_AddModule`, which in turn don't work (well) with multi-phase init. I've replaced them with a global `static MsgspecState`, set during module init; This is *not* safe for sub-interpreters, but handling it *properly* would be a large refactor that I think is best kept for separate PRs.

<hr>

**Road to subinterpreter support**

This is the first step for msgspec to support subinterpreters, but it's still a long way to go. I'll open a separate tracking issue for that once this gets merged, but it'll likely involve some larger refactors, especially around moving things to [heap types](https://docs.python.org/3/c-api/typeobj.html#heap-types) where absolutely necessary, and ensuring this does not impact performance negatively.


tests/unit/test_module.py
provinzkraut, line 1: since we're changing how the subinterpreter support is flagged, i wanted some tests to show it's still working

src/msgspec/_core.c
```diff
@@ -22399,195 +22405,180 @@ msgspec_traverse(PyObject *m, visitproc visit, void *arg)
     return 0;
 }
 
-static struct PyModuleDef msgspecmodule = {
```
provinzkraut, line 22402: This was simply moved down to make the flow easier, as it references `_core_exec`

src/msgspec/_core.c
```diff
@@ -539,13 +539,18 @@ msgspec_get_state(PyObject *module)
     return (MsgspecState *)PyModule_GetState(module);
 }
 
-/* Find the module instance imported in the currently running sub-interpreter
-   and get its state. */
+/*
+with multi-phase init PyState_FindModule is not usable.
+since we declare MULTIPLE_INTERPRETERS_NOT_SUPPORTED, we are guaranteed to have
+at most one live module instance per process, so we can cache its state here.
+state is populated populate in _core_exec and cleared by m_clear / m_free.
```
Siyet: A few things to clean up in this block:

- `populated populate` - duplicated word.
- `cleared by m_clear / m_free` - only `m_free` clears `_core_state`. `m_clear` only Py_CLEARs the per-module state fields, the global pointer stays put.
- `we declare MULTIPLE_INTERPRETERS_NOT_SUPPORTED` - reads like an explicit declaration, but we don't set `Py_mod_multiple_interpreters` at all. The single-interpreter behavior comes from the default for that slot. Worth rephrasing, e.g. "we don't opt into `Py_mod_multiple_interpreters`, so the module defaults to single-interpreter only".
- Sentence casing inside the block is inconsistent with other block comments in this file.

Siyet (commented):
Leaving a note for the history / backlink: this obsoletes the `PyState_AddModule` workaround jcrist added in #561, where the body said "in the long run we should move to using multiphase module init which should avoid this problem entirely." Good to have the thread connected.

goodboy:
> I'll open a separate tracking issue for that once this gets merged, 

please do ping if/when that happens 🙏🏼

src/msgspec/_core.c
```diff
@@ -22399,195 +22405,176 @@ msgspec_traverse(PyObject *m, visitproc visit, void *arg)
...
-    .m_size = sizeof(MsgspecState),
-    .m_methods = msgspec_methods,
-    .m_traverse = msgspec_traverse,
-    .m_clear = msgspec_clear,
-    .m_free =(freefunc)msgspec_free
-};
 
-PyMODINIT_FUNC
-PyInit__core(void)
+int _core_exec(PyObject *m)
```
Siyet: Missing `static` - the other module-level helpers (`msgspec_clear`, `msgspec_free`, `msgspec_traverse`, `module_slots`, `msgspecmodule`) are all static, and `nm -D` confirms `_core_exec` currently leaks as an exported symbol from the .so.

tests/unit/test_module.py
```diff
@@ -0,0 +1,29 @@
...
+@pytest.mark.skipif(sys.version_info < (3, 12), reason="3.12+ only")
+class TestSubinterpreterGuard:
+    # sub-interpreters currently not supported. make sure we correctly report that
+    @pytest.mark.skipif(sys.version_info >= (3, 13), reason="3.12 only")
+    def test_subinterpreter_import_rejected_312(self):
+        interpreters = pytest.importorskip("test.support.interpreters")
+
+        interp = interpreters.create()
+        try:
+            with pytest.raises(match=".*does not support loading in subinterpreters"):
```
Siyet: `pytest.raises(match=...)` accepts any exception. The actual one here is `interpreters.RunFailedError`, so `pytest.raises(interpreters.RunFailedError, match=...)` would also catch a regression where some other exception is raised. Not a blocker.

src/msgspec/_core.c
Siyet: Missing `static` - the other module-level helpers (`msgspec_clear`, `msgspec_free`, `msgspec_traverse`, `module_slots`, `msgspecmodule`) are all static, and `nm -D` confirms `_core_exec` currently leaks as an exported symbol from the .so.

tests/unit/test_module.py
Siyet: `pytest.raises(match=...)` accepts any exception. The actual one here is `interpreters.RunFailedError`, so `pytest.raises(interpreters.RunFailedError, match=...)` would also catch a regression where some other exception is raised. Not a blocker.

Siyet:
@provinzkraut Happy to push these small fixups to your branch myself if you're short on time, just give me a thumbs up. Keen to get this in so the subinterp work in #1026 can start moving.

provinzkraut:
@Siyet feel free to push. Otherwise I could get it done some time later this week

src/msgspec/_core.c
provinzkraut: Fixed

src/msgspec/_core.c
provinzkraut: I've added `MULTIPLE_INTERPRETERS_NOT_SUPPORTED` to the module slots. I'm not sure why it worked before, I'll have to revisit the tests, since according to the docs, not setting it should default to `Py_MOD_MULTIPLE_INTERPRETERS_SUPPORTED`.

https://docs.python.org/3/c-api/module.html#c.Py_mod_multiple_interpreters

tests/unit/test_module.py
provinzkraut: Added

src/msgspec/_core.c
```diff
@@ -539,13 +539,19 @@ msgspec_get_state(PyObject *module)
...
-/* Find the module instance imported in the currently running sub-interpreter
-   and get its state. */
+/*
+With multi-phase init PyState_FindModule is not usable.
+since we declare MULTIPLE_INTERPRETERS_NOT_SUPPORTED, we are guaranteed to have
+at most one live module instance per process, so we can cache its state here.
+state is populated in _core_exec and cleared and freed by m_clear and m_free
+respectively.
+*/
+static MsgspecState *_core_state = NULL;
```
sobolevn, line 549: I would advokate against this pattern :)
It would be required to refactor it anyway for future FT / SI builds.

Maybe we can instead use something like

```c
static MsgspecState *
_get_current_module_state(void)
{
    PyObject *mod = _get_current_module();
    if (mod == NULL) {
        mod = PyImport_ImportModule("msgspec._core");
        if (mod == NULL) {
            return NULL;
        }
    }
    MsgspecState *state = get_module_state(mod);
    Py_DECREF(mod);
    return state;
}
```
?

src/msgspec/_core.c
```diff
@@ -22824,16 +22812,44 @@ PyInit__core(void)
...
+};
+
+static struct PyModuleDef msgspecmodule = {
+    PyModuleDef_HEAD_INIT,
+    .m_name = "msgspec._core",
+    .m_size = sizeof(MsgspecState),
+    .m_methods = msgspec_methods,
+    .m_traverse = msgspec_traverse,
+    .m_clear = msgspec_clear,
+    .m_free =(freefunc)msgspec_free,
```
```suggestion
    .m_free = (freefunc)msgspec_free,
```

style nit

src/msgspec/_core.c
```diff
@@ -22824,16 +22812,44 @@ PyInit__core(void)
...
     );
-    if (st->StructType == NULL) return NULL;
+    if (st->StructType == NULL) return -1;
     Py_INCREF(st->StructType);
-    if (PyModule_AddObject(m, "Struct", st->StructType) < 0) return NULL;
+    if (PyModule_AddObject(m, "Struct", st->StructType) < 0) return -1;
 #ifdef Py_GIL_DISABLED
-    PyUnstable_Module_SetGIL(m, Py_MOD_GIL_NOT_USED);
+    #if !PY313_PLUS
+        PyUnstable_Module_SetGIL(m, Py_MOD_GIL_NOT_USED);
```
sobolevn: This is not needed, we set `{Py_mod_gil, Py_MOD_GIL_NOT_USED},` module slot

src/msgspec/_core.c
```diff
@@ -22710,86 +22698,86 @@ PyInit__core(void)
...
     Py_DECREF(temp_module);
-    if (temp_obj == NULL) return NULL;
+    if (temp_obj == NULL) return -1;
     st->astimezone = PyObject_GetAttrString(temp_obj, "astimezone");
     Py_DECREF(temp_obj);
-    if (st->astimezone == NULL) return NULL;
+    if (st->astimezone == NULL) return -1;
 
     /* uuid module imports */
     temp_module = PyImport_ImportModule("uuid");
```
sobolevn, line 22759: There are missing `Py_DECREF(temp_module)` calls in many places.

provinzkraut: Yeah, I think this was discussed in a previous PR. I think we should clean those up. Should be a separate PR though IMO

src/msgspec/_core.c
sobolevn: Or `PyType_GetModuleByDef` / `PyType_GetModule` :)

provinzkraut: I've tried using `PyType_GetModule` at first, but there were some issues. Can't remember what it was though, it's been a while. Let me re-evaluate that again

provinzkraut: `PyType_GetModuleByDef` is only available on 3.11+. I'll see if `PyType_GetModule` can be made to work

provinzkraut: Best would probably to just pipe the mod through to every caller that needs it. A bit bigger API surface, but less mucking around with global state and such. Currently checking if that's viable

provinzkraut: @sobolevn looking at your suggested alternative, what would `_get_current_module` do? Is that still some internal global state? Where does it store the ref?

provinzkraut: @sobolevn So I played around with it a bit. I tried to reduce the calls to the global state by piping it through wherever possible, which does work. However, it reveals quite a lot of issues, specifically in places where that piping through is currently just not possible where we have static types, which means if we don't want to store process-global state anywhere, things get really complicated (or at least I haven't been able to find a good solution).

What we *could* do is reverse the implementation order here:

1. Make the move to heap types
2. Remove global state
3. Multi-phase init

provinzkraut: The solution I could find to address everything within this PR would be to use change `msgspec_get_global_state` to use [`PyInterpreterState_GetDict`](https://docs.python.org/3/c-api/subinterpreters.html#c.PyInterpreterState_GetDict), but its docs say 

> This is not a replacement for [PyModule_GetState()](https://docs.python.org/3/c-api/module.html#c.PyModule_GetState), which extensions should use to store interpreter-specific state information.


Soo.. I think unless we're fine with having this global state around until we migrate to heap types, I don't see a way how we can get this PR merged. Which is fine, I don't mind doing it the other way around

sobolevn: Oh, we don't have heap types yet 🤦‍♂️ 
You are right, this is the required first step for new APIs :)

provinzkraut: > You are right, this is the required first step for new APIs :)

Alright, let's try working towards this then. For that, we really ought to have some continuous benchmarking set up though. I'll take a look at https://codspeed.io

ofek: I'm not sure if you can see but I accepted the installation request. It looks like they [recommend](https://codspeed.io/docs/integrations/ci/github-actions/configuration#authentication) the OIDC method of authenticating in GitHub Actions and here is the badge:

```
[![CodSpeed](https://img.shields.io/endpoint?url=https://app.codspeed.io/badge.json)](https://app.codspeed.io/msgspec/msgspec)
```

Siyet:
Reviewed and tested this (built on 3.12, full unit suite green, the `test_module.py` subinterp check passes). The multi-phase mechanics look correct. A couple of non-blocking things on top of what's already in the thread:

- This overlaps PR #1059 (which converts the same `PyModule_AddObject` calls in `_core_exec` to `PyModule_AddObjectRef`). Whichever lands second needs a rebase, worth sequencing them deliberately.
- On 3.10/3.11 the `Py_mod_multiple_interpreters` slot doesn't exist, so subinterpreter rejection is structurally absent there (CPython limitation, not this PR). A one-line comment near `module_slots` noting that would help future readers.

Built and suite-verified on 3.12.

Siyet:
Pushed the small review nits we discussed:

- corrected the `_core_state` comment (`m_clear` does not touch the pointer; only `m_free` resets it; also dropped the overstated "guaranteed at most one instance per process" wording)
- dropped the dead `PyUnstable_Module_SetGIL` block, the `{Py_mod_gil, Py_MOD_GIL_NOT_USED}` slot already covers it (per @sobolevn)
- `.m_free` spacing

Left out of scope as agreed: the missing `Py_DECREF(temp_module)` cleanups (separate PR per the thread) and the global-state -> heap-types rework.

provinzkraut:
Currently blocked until we have moved to heap types. See https://github.com/msgspec/msgspec/pull/1024#discussion_r3404623107

I'll revert this to a draft

PR #1028, opened 2026-05-09
place `null` last in `anyOf` for optional unions in JSON schema

### LINKED ISSUE #1027: `field: Schema | None = None` should generate `anyOf` where `null` is the last item
Hi 👋 

I am using `msgspec` to generate JsonSchema for my schemas in https://github.com/wemake-services/django-modern-rest/issues/990

But, there's a big difference with how `msgspec` and `pydantic` generates schema for this type:

```python
class _TailsSchema(pydantic.BaseModel):
    id: int
    name: str

class _UserCreateModel(pydantic.BaseModel):
    tails: _TailsSchema | None = None
```

vs

```python
class _TailsSchema(msgspec.Struct):
    id: int
    name: str

class _UserCreateModel(msgspec.Struct):
    tails: _TailsSchema | None = None
```

## What schema is generated

`pydantic`:

```json
"_UserCreateModel": {
  "properties": {
    "tails": {
      "anyOf": [
        {
          "$ref": "#/components/schemas/_TailsSchema"
        },
        {
          "type": "null"
        }
      ]
    },
    // ...
}
```

`msgspec`:

```json
"_UserCreateModel": {
  "properties": {
    "tails": {
      "anyOf": [
        {
          "type": "null"
        },
        {
          "$ref": "#/components/schemas/_TailsSchema"
        }
      ]
    },
    // ...
}
```

The final result from `msgspec` makes Swagger to render

```json
{
  "email": "string",
  "tail": null,
  "tail_optional": null
}
```

as the example value, which is correct, but less ideal than the pydantic's version:

```json
{
  "email": "string",
  "tail": {
    "id": 0,
    "name": "string"
  },
  "tail_optional": {
    "id": 0,
    "name": "string"
  }
}
```

Looks like that Swagger just takes the first value from `anyOf`, it would be reasonable to change the order. It would also be more semantically similar to how `Schema | None` is used in Python, not `None | Schema`

SarthakB11:
Fixes #1027.

For `T | None`, `_json_schema.py` was emitting `{"type": "null"}` before
the struct/`$ref` entry in `anyOf`. Tools that pick the first `anyOf`
element for example generation (Swagger UI in the issue) ended up showing
`null`. Pydantic puts `null` last for the same reason.

Move `NoneType` out of the union iteration into a `has_none` flag and
append `{"type": "null"}` last in all three union branches (no-struct,
single-struct, tagged-struct). The tagged-struct branch's `if options:`
guard becomes `if options or has_none:` so `Tagged | None` keeps `null`
in its `anyOf` instead of dropping it.

Observable changes:

- `Struct | None`: was `[null, $ref]`, now `[$ref, null]`.
- `Union[X, None, Y]`: `null` moves from the middle to the end.
- `str | None`: already `[string, null]` on CPython, unchanged.

Tests added in `tests/unit/test_schema.py` for `Optional[Struct]`,
`Union[int, None, str]`, and `Union[Tagged, None]` (covers the third
branch that previously dropped `null` entirely).


Siyet:
Hi @SarthakB11, this currently fails `tests/unit/test_schema.py::test_struct_tagged_union_with_none` across all Python versions, the `null`-last reordering interacts with tagged-union schemas. Could you take a look when you get a chance? Happy to review once CI is green.

SarthakB11:
Thanks for the catch, @Siyet. The bug was in the `len(structs) >= 2` branch: I flipped `if options:` to `if options or has_none:`, which forced the wrapped shape even when there were no other members, hiding the discriminator one level down.

Fix: keep the flat `schema.update(struct_schema)` path when only None is added, and append `{"type": "null"}` to the resulting `anyOf` after the update. All 113 tests in `tests/unit/test_schema.py` pass locally. Pushed `139e407`.

Siyet (changes requested):
Thanks for the quick turnaround on the tagged-union fix! I did a deep pass on b5b078a, including an empirical schema diff against 0.21.1 across ~25 type shapes. The ordering change itself looks right and minimal for most shapes, but two behavior changes go beyond reordering and need fixing before this can merge:

**1. `Meta` metadata on a `None` union member is silently dropped**

```python
from typing import Annotated, Union
from msgspec import Meta, json

typ = Union[Annotated[None, Meta(description="explicit null means reset")], int]
json.schema(typ)
```

- 0.21.1: `{"anyOf": [{"description": "explicit null means reset", "type": "null"}, {"type": "integer"}]}`
- this PR: `{"anyOf": [{"type": "integer"}, {"type": "null"}]}` - the `description` (same for `title`/`examples`/`extra_json_schema`) is lost.

The unwrap loop detects `NoneType` through the `Metadata` wrapper but then discards the subtype and appends a hardcoded `{"type": "null"}`. Fix: remember the subtype (e.g. `none_member = subtype`) and append `self.to_schema(none_member)` instead of the literal.

**2. `Optional[tagged union]` changes schema shape, not just order**

For `Union[A, B, None]` where A/B are tagged structs and there are no other members:

- 0.21.1: `{"anyOf": [{"type": "null"}, {"anyOf": [ref A, ref B], "discriminator": ...}]}` - discriminator scoped to the struct members only.
- this PR: `{"discriminator": ..., "anyOf": [ref A, ref B, {"type": "null"}]}` - discriminator hoisted onto a composite that now contains a null branch.

Problems:

- OpenAPI 3.x expects every schema under a discriminator-bearing composite to carry the discriminator property; an inline `{"type": "null"}` can't, and it has no `mapping` entry. Strict tooling (openapi-generator, spectral lints) chokes on this.
- It's inconsistent within this same PR: `Union[A, B, int, None]` keeps the nested `{anyOf + discriminator}` member form, so the same tagged union renders in two different shapes depending on its neighbors.
- pydantic emits the nested form here, so this diverges exactly where #1027 asks for parity.

Fix: in the `len(structs) >= 2` branch with empty `options`, keep the nested form and emit `{"anyOf": [struct_schema, {"type": "null"}]}` instead of `schema.update(struct_schema)` plus a top-level append. `test_struct_tagged_union_with_none` currently pins the flat shape and would need updating along with this.

Smaller notes:

- Test gaps worth closing while you're here: `Union[A, B, int, None]` (the mixed branch isn't covered by any test) and the `Annotated[None, Meta(...)]` case from point 1.
- The PR description still mentions the `if options or has_none:` guard that's no longer in the diff - worth refreshing once the above lands.

Everything else checked out clean: no lost union members, no duplicated nulls, `$defs`/recursive refs intact, plain unions byte-identical to main, and all outputs pass JSON Schema Draft 2020-12 meta-validation.

SarthakB11:
Thanks for the review. Both points are addressed.

1. None-member metadata is now preserved. The union unwrap loop keeps the actual none_member subtype and emits self.to_schema(none_member) instead of a hardcoded {"type": "null"}, so Annotated[None, Meta(...)] carries its metadata into the schema.

2. Optional[tagged union] no longer flattens the shape. In the empty-options case it now emits {"anyOf": [struct_schema, <null>]}, keeping the nested discriminator form so only ordering changes, not the shape.

Tests: updated test_struct_tagged_union_with_none to the nested anyOf shape, and added test_struct_tagged_union_with_none_and_other (Union[Point, Point3D, int, None]) and test_optional_union_null_member_metadata_preserved (Annotated[None, Meta(...)]). Full tests/unit/test_schema.py passes (115 passed, 8 skipped).

One note: the PR description still references an `if options or has_none:` guard that is no longer in the diff. I will refresh the description to match the current code.

Siyet (approved):
Both change requests are resolved. Re-checked empirically against current `main` across the shapes from the last pass:

- None-member `Meta` is preserved now: `Union[Annotated[None, Meta(...)], int]` keeps `description` (and `title`/`examples`/`extra_json_schema`) on the null branch, just moved last.
- `Optional[tagged union]` keeps the nested form: `Union[A, B, None]` emits `{"anyOf": [{"anyOf": [refs], "discriminator": ...}, {"type": "null"}]}` with no top-level discriminator, and `Union[A, B, int, None]` renders the same nested member, so the two are consistent again.

No regressions: plain unions and tagged unions without `None` are byte-identical to `main`, null is only relocated (never duplicated or dropped), $defs/refs intact, and every output passes JSON Schema Draft 2020-12 meta-validation. `test_schema.py` is green (115 passed) and the branch merges cleanly on current `main`. Thanks for the thorough follow-up!

PR #1038, opened 2026-06-07
refactor: Use `Py_UNREACHABLE()` macro for unreachable code

sobolevn:
This macro is available since 3.7 and should be used in such cases :)


provinzkraut (approved):

provinzkraut:
Thanks for all the effort @sobolevn :)

PR #1040, opened 2026-06-07
Fix potential memory leak in `Encoder_init`

sobolevn:
There was a potential memory leak in `Encoder_init`.

Why?
1. Some valid callable is passed as `enc_hook`, it has +1 rc in `Py_INCREF(enc_hook);`
2. Then, invalid order is checked, we return `-1` from `Encoder_init`, but does not `Py_DECREF(enc_hook);`

We can just swap the check order. First check things that do not incref, lastly incref.

Repro:

```python
>>> from msgspec.json import Encoder
>>> import sys
>>> def test(*args, **kwargs): ...
... 
>>> sys.getrefcount(test)
2
>>> Encoder(enc_hook=test, order='wrong')
ValueError: `order` must be one of `{None, 'deterministic', 'sorted'}`, got 'wrong'
>>> sys.getrefcount(test)
3
```


provinzkraut (approved):
Good catch, thanks!

PR #1041, opened 2026-06-07
Update all code left-overs to 3.10+

sobolevn:
Things I used:
- `pyupgrade`
- A lot of manual labour 😅 

Considerations:
- I checked that both new and old type forms are covered with tests
- I double checked that no particular APIs require older type forms (`typing._GenericAlias` vs `types.GenericAlias` difference)


src/msgspec/_utils.py
```diff
@@ -5,7 +5,7 @@
 from typing import _AnnotatedAlias  # noqa: F401
 
 try:
-    from typing_extensions import get_type_hints as _get_type_hints
+    from typing import get_type_hints as _get_type_hints
 except Exception:
     from typing import get_type_hints as _get_type_hints
```
provinzkraut, line 10: That whole `try/except` should go now, since the branches are the same

tests/unit/test_common.py
provinzkraut, line 1: I'm a bit torn with these test upgrades. Even though we can use the builtin generics, we should still support the `typing.`-forms, and as such, these should also be tested. Not sure what the least amount of effort is, but ideally we'd have at least some parametrisation for the core usages, that show both forms work

sobolevn: Yes, there are many cases where I added both forms.

sobolevn:
@provinzkraut does this look better now? :)
I want to merge this, because my `frozendict` PR (https://github.com/msgspec/msgspec/pull/1052) would have lots of conflicts.

provinzkraut:
> @provinzkraut does this look better now? :) I want to merge this, because my `frozendict` PR would have lots of conflicts.

I'll check it out later today!

ofek:
I was going to take a look but noticed that this needs to be rebased.

sobolevn:
Oups, updated!

tests/unit/test_inspect.py
```diff
@@ -263,14 +263,14 @@ def test_sequence(kw, typ, info_type, has_item_type):
...
-@pytest.mark.parametrize("typ", [Tuple, tuple])
+@pytest.mark.parametrize("typ", [tuple, Tuple])
 def test_tuple(typ):
     assert mi.type_info(typ[()]) == mi.TupleType(())
     assert mi.type_info(typ[int]) == mi.TupleType((mi.IntType(),))
     assert mi.type_info(typ[int, float]) == mi.TupleType((mi.IntType(), mi.FloatType()))
 
 
-@pytest.mark.parametrize("typ", [Dict, dict])
+@pytest.mark.parametrize("typ", [dict, Dict])
```
provinzkraut: Just curious, is there a reason why these are flipped in order now?

src/msgspec/_utils.py
provinzkraut: Did something go wrong here? Seems like this change was completely reverted.

provinzkraut:
@sobolevn all good from my side, except one open question

tests/unit/test_inspect.py
sobolevn: Nope :)
Reverted!

src/msgspec/_utils.py
provinzkraut: @sobolevn in case you missed the comment on the resolved comment :sweat_smile:

sobolevn: Yes, this is correct. Thanks for the double check!
`get_type_hints` should be used from `typing_extensions`:
1. Because they are already. Changing this would possible introduce behavior changes
2. Because `typing_extensions.get_type_hints` are different to `typing.get_type_hints` on many python version in many specific cases

Changing it was a mistake :)

provinzkraut (approved):

amrebrahim682-spec (commented):
``

PR #1049, opened 2026-06-10
Improve PR docs previews

ofek:


ofek:
@msgspec/committers We now have PR previews of docs changes accessible via a "View deployment" button 😄

PR #1050, opened 2026-06-11
See changes to PR deployment metadata

ofek:


ofek:
@msgspec/committers This is now the final version, where the metadata is cleaned up (no more `deployed` linking to a Cloudflare account view) and the URLs are prefixed by the PR number rather than a hash e.g. `pr-1050.msgspec.pages.dev`.

ofek:
FYI, I just realized that the docs shouldn't have even been deployed since nothing changed. Can one of you please make an issue to track this? Either I messed up CI or our docs builds are not reproducible.

Issue #885, opened 2025-10-20
Add CI gate for non-trivial regressions in startup time


ofek:

provinzkraut:
We have this now, added in #1075 and #1087. Not extremely detailed, but significant regressions will be caught now

ofek:
Fantastic work, thank you very much!

Issue #884, opened 2025-10-20
Improve startup overhead


ofek:

ofek:
When it comes time to start robustly tracking this, I strongly recommend that we use `hyperfine` to measure the true overhead by computing the delta of the interpreter startup time.

```nu
❯ (
    uv run --python 3.14 --no-project --isolated --with msgspec
    -- hyperfine
        --export-json -
        --style none
        --shell none
        --min-runs 100
        --warmup 10
        --reference-name baseline
        --reference 'python -c pass'
        --command-name combined
        'python -c "import msgspec"'
    err> null-device
    | from json
    | get results
    | select command median
    | update median {|r| $r.median * 1000}
    | transpose --header-row --as-record
    | insert overhead {|r| $r.combined - $r.baseline}
)
╭──────────┬───────╮
│ baseline │ 47.02 │
│ combined │ 87.67 │
│ overhead │ 40.65 │
╰──────────┴───────╯
```

Issue #310, opened 2023-02-10
Set up continuous benchmarking


jcrist:
### Description

`msgspec` is a performance oriented library - would be useful to set up continuous benchmarking as part of our CI to catch any performance regressions. These might make use of airspeed velocity, or something else.

jcrist:
A few links that might be useful:

- https://labs.quansight.org/blog/2021/08/github-actions-benchmarks
- https://asv.readthedocs.io/en/stable/commands.html#asv-continuous
- https://github.com/conbench/conbench

jcrist:
I've seen some libraries are using CodSpeed (https://codspeed.io/) for this now. I'm *slightly* hesitant to rely on a new closed-source service.

From reading their Python client code, it looks like they're using callgrind to handle instrumentation. Depending on metrics used this may be a nice way to eliminate variation between runs in a public CI environment. Worth considering at least.

jcrist:
Another potentially useful reference: https://pythonspeed.com/articles/consistent-benchmarking-in-ci/

provinzkraut:
One thing to note about https://codspeed.io/ is that it runs the benchmark itself within the GitHub actions runner. While they claim that 

> Running benchmarks using a traditional statistical approaches is usually unreliable on CI environments. To measure the performance of your code in an accurate and reproducible way, CodSpeed uses various instruments when running your code. Those instruments isolate noisy neighbors and all the side effect they can have on when measuring the performance of our workloads. [^1]

I'm not super convinced this is viable for the sort of measurements we are taking; the required accuracy is quite high, and to get comparable results in the microsecond realm between entirely different runners sounds implausible. 

https://bencher.dev would be an alternative, where the benchmark is actually run in the cloud, on a bare metal machine, and they guarantee that the same benchmarks run on the same machine every time, which sounds good. It's also open source.

[^1]: https://codspeed.io/docs/what-is-codspeed

ofek:
Bencher looks way better, nice find!

UnknownPlatypus:
> https://bencher.dev/ would be an alternative, where the benchmark is actually run in the cloud, on a bare metal machine

fyi, codespeed also have [bare metal runners](https://codspeed.io/docs/features/macro-runners), might be similar to bencher

provinzkraut:
> fyi, codespeed also have [bare metal runners](https://codspeed.io/docs/features/macro-runners), might be similar to bencher

[...]

PR #1052, opened 2026-06-11
Add `frozendict` support for 3.15+

### LINKED ISSUE #1029: support for python 3.15 frozendict
### Description

Thank you for the lovely library!

Now that python 3.15 beta 1 has been released with the [newly added frozendict builtin type](https://peps.python.org/pep-0814/), I'm curious if it can be included as a supported field in `msgspec`. I'd find this exciting because it would allow a clean way to define a frozen `Struct` in which dict members are also immutable (and hashable to boot).

sobolevn:
Closes #1029


src/msgspec/_core.c
```diff
@@ -4011,7 +4016,7 @@ typenode_from_collect_state(TypeNodeCollectState *state) {
         Py_INCREF(state->c_str_regex);
         out->details[e_ind++].pointer = state->c_str_regex;
     }
-    if (state->dict_key_obj != NULL) {
+    if (state->dict_key_obj != NULL && state->dict_val_obj != NULL) {
```
provinzkraut, line 4045: Why do we now also check the val state before converting?

src/msgspec/_core.c
```diff
@@ -4160,7 +4165,7 @@ typenode_collect_check_invariants(TypeNodeCollectState *state) {
             MS_TYPE_TYPEDDICT | MS_TYPE_DATACLASS
         )
     );
-    if (state->types & MS_TYPE_DICT) {
+    if (state->types & (MS_TYPE_DICT | MS_TYPE_FROZENDICT)) {
```
provinzkraut: There are a lot of these around now, easy to miss. Maybe it would be good to introduce a generic `MS_ANY_DICT`, and then use that instead?

src/msgspec/_core.c
```diff
@@ -14688,6 +14711,11 @@ json_encode_inline(EncoderState *self, PyObject *obj)
     else if (PyDict_Check(obj)) {
         return json_encode_dict(self, obj);
     }
+#if PY315_PLUS
+    else if (PyFrozenDict_Check(obj)) {
+        return json_encode_dict(self, obj);
+    }
+#endif
```
provinzkraut, line 14799: Not sure if it's worth it, but on 3.15+ we could collapse these into [`PyAnyDict_Check`](https://docs.python.org/3.15/c-api/dict.html#c.PyAnyDict_Check)

src/msgspec/_core.c
```diff
@@ -16086,6 +16114,20 @@ mpack_decode_dict(
...
+static PyObject *
+mpack_decode_frozendict(
+    DecoderState *self, Py_ssize_t size, TypeNode *key_type,
+    TypeNode *val_type, PathNode *path
+) {
+    PyObject *dict = mpack_decode_dict(self, size, key_type, val_type, path);
+    if (dict == NULL) {
+        return NULL;
+    }
+    return PyFrozenDict_New(dict);
```
provinzkraut: According to [PEP 814, calling `PyFrozenDict_New` with a `dict` is `O(n)`](https://peps.python.org/pep-0814/#construction). Maybe it would be better to adjust `mpack_decode_dict` / `json_decode_dict` so they can also natively decode into a frozendict?

provinzkraut: Ah, never mind. They create the dicts by calling `PyDict_SetItem`, which obviously does not work with frozendicts.
Mhm. We should add a note in the docs about this possibly unexpected loss of performance when decoding `frozendict`s.

src/msgspec/_core.c
```diff
@@ -21535,6 +21630,19 @@ convert_dict_to_dict(
     return NULL;
 }
 
+#if PY315_PLUS
+static PyObject *
+convert_dict_to_frozendict(
+    ConvertState *self, PyObject *obj, TypeNode *type, PathNode *path
+) {
+    PyObject *dict = convert_dict_to_dict(self, obj, type, path);
```
provinzkraut, line 21705: Also `O(n*2)`, worth mentioning

src/msgspec/inspect.py
```diff
@@ -881,6 +908,13 @@ def _translate_inner(
                 min_length=min_length,
                 max_length=max_length,
             )
+        elif sys.version_info >= (3, 15) and t is frozendict:  # noqa: F821
```
provinzkraut: Probably worth to extract the version check into a constant

provinzkraut: Also use the `_future_builtins_` here?

tests/unit/test_convert.py
provinzkraut, line 1: Maybe parametrize `TestDict`?

tests/unit/test_to_builtins.py
```diff
@@ -309,6 +316,95 @@ def test_dict_sequence_keys(self):
...
 
+    def check_frozen(self, val, expected):
+        assert type(val) is type(expected)
+        assert val == expected
+
+    @pytest.mark.skipif(
+        sys.version_info < (3, 15), reason="frozendict was added in 3.15"
+    )
+    @pytest.mark.parametrize("subclass", [False, True])
+    def test_frozendict(self, subclass):
```
provinzkraut, line 325: Same here, maybe parametrize the dict tests? A bit easier to keep track of when modifying tests

provinzkraut (commented):
Great stuff @sobolevn! Couple of questions :)

src/msgspec/_core.c
sobolevn: It is a separate issue: this state was never checked. And `dict_val_obj` could be `NULL` in theory. Now, it is properly checked.

src/msgspec/_core.c
sobolevn: yes, this will be a nice clean up in the far future :)

src/msgspec/_core.c
sobolevn: I don't think that it will be this bad. Collections had been rebuilt before in the code.

In the meanwhile I proposed to add `PyFrozenDict_FromDictSteal()` C-API to CPython.

src/msgspec/inspect.py
sobolevn: nope, this is a runtime code, it might have this module. I used this hack in tests only, where we control the env.

tests/unit/test_convert.py
sobolevn: It would not be that simple, some tests are slighly diffent in two classes :(

src/msgspec/_core.c
provinzkraut: Nice! Did you make a pass through the rest of the code if there's other branches where we're missing that check?

src/msgspec/_core.c
provinzkraut: > I don't think that it will be this bad. Collections had been rebuilt before in the code.

Probably not, but we also have a section in the docs on how to fine-tune garbage collection, so I think it's worth mentioning as a micro-optimisation. I could also see it mattering for biiig dicts, which I've heard sometimes occur in some code :)

> In the meanwhile I proposed to add `PyFrozenDict_FromDictSteal()` C-API to CPython.

That would be super nice. You got my vote!

tests/unit/test_convert.py
provinzkraut: Sad

src/msgspec/_core.c
sobolevn: Yes, there were two places, I updated both of them :)

sobolevn:
Done! Very helpful feedback, thanks a lot!
I've added a new doc section.

docs/perf-tips.rst
```diff
@@ -288,6 +288,49 @@ average another ~2x speedup for decoding (and ~1.5x speedup for encoding).
...
+    ...     items: tuple[str, ...]
+
+    >>> msg = b'{"items": ["pen", "book"]}'
+
+    >>> msgspec.json.decode(msg, type=Example)
+    Example(items=("pen", "book"))
+
+We would first create a ``list`` object and then convert it to ``tuple``,
+using double the memory and ``O(n)`` time to do that.
+This is true for all conversion methods.
```
provinzkraut: Maybe something like

```suggestion
This is true for all conversions, that require an intermediate representation.
```
?

docs/perf-tips.rst
```diff
@@ -288,6 +288,49 @@ average another ~2x speedup for decoding (and ~1.5x speedup for encoding).
...
+
+To achieve the best performance,
+use native ``json`` types: ``list`` and ``dict``.
```
provinzkraut: Other collection types can be used as well, e.g. `set` does not have the overhead, as it can be constructed iteratively.

docs/perf-tips.rst
```diff
@@ -288,6 +288,49 @@ average another ~2x speedup for decoding (and ~1.5x speedup for encoding).
...
+
+  class Example(msgspec.Struct):
+      items: frozendict[str, int]  # requires Python3.15+ to be used
+
+  msg = b'{"items": {"pen": 1, "book": 2}}'
+  msgspec.json.decode(msg, type=Example)
+  # Example(items=frozendict({"pen": 1, "book": 2}))
+
+Will first parse ``items`` as a regular :class:`dict`
+and will then convert it to ``frozendict`` using ``O(n)`` complexity.
```
provinzkraut: A bit more precise and also explaining the reason *why*: (don't have to use this literal phrasing):

```suggestion
As `frozendict` does not allow adding items one at a time, 
msgspec will first parse ``items`` as a regular :class:`dict`,
and will then convert it to ``frozendict``, which results in the 
total decoding operation having ``O(n*2)`` time complexity.
```

provinzkraut (commented):
Just some thoughts on the docs, the rest looks good :)

docs/perf-tips.rst
```diff
@@ -323,12 +325,14 @@ The same can be said for :class:`tuple` vs :class:`list`:
...
-This is true for all conversion methods.
+We would first create a ``list`` object and then convert
+it to variable-sized ``tuple``,
+using double the memory and ``O(n*2)`` time to do that.
+This is true for all conversions, that require an intermediate representation.
 
 To achieve the best performance,
-use native ``json`` types: ``list`` and ``dict``.
+use collection types that can be constructed natively:
+``list``, ``set``, ``frozenset``, fixed-sized ``tuple``, and ``dict``.
```
provinzkraut, line 335: Just curious, do you know why `PySet_Add` can add to a `frozenset`, but `PyDict_SetItem` cannot be used on a `frozendict`?

Was there a reasoning behind it?

provinzkraut (dismissed):
🔥

sobolevn:
Let's keep this open for some time to collect more feedback 👍 
Thanks for your help!

ofek:
I don't think Siyet has done this yet so would either of you be interested in tackling my comment [here](https://github.com/msgspec/msgspec/issues/990#issuecomment-4208327334)?

> Specifically, as a first step I would want to improve the performance checks so that the results of each commit to main are persisted (perhaps on a dedicated branch) and can be used on PRs to detect regressions.

We need to do the following, in order, for both performance (#310) and startup time (#884):

1. Set up some sort of rudimentary measurement in CI (profiling was introduced in #923 and improved in subsequent PRs like #934)
2. Manifest the results in a comment on PRs that gets updated after every commit
3. Fail CI for regressions (e.g. #885)
4. Improve upon the measurements themselves and measure more things we care about (for example [Memray](https://github.com/bloomberg/memray) is [included](https://github.com/msgspec/msgspec/blob/38aeb6ab10e6ced849bd9aa55c0571a065bcbac2/pyproject.toml#L94) as a profiling dependency but not actually used anywhere because I didn't have time and it doesn't support Windows)

Pretty please try to prioritize this over other tasks and be sure to manually look at the drift of the existing profiling jobs before the next release 🙏

sobolevn:
I created a public discussion about new potential C-API to efficently convert `dict` into `frozendict` objects: https://discuss.python.org/t/provide-a-c-api-to-effectively-convert-dict-into-frozendict/107754

Siyet (changes requested):
Took a careful pass on this (including building the branch against 3.15.0a7) and found two issues to fix before merge. Nice feature otherwise, the type-system integration is clean.

**1. Reference leak in every frozendict construction path.**

`PyFrozenDict_New(dict)` does not steal its argument: for a plain dict it copies (via `PyTuple_Pack` + `frozendict_new`) and leaves the caller's reference intact (CPython `Objects/dictobject.c`, which is exactly why you proposed `PyFrozenDict_FromDictSteal` upstream). The wrappers do `return PyFrozenDict_New(dict);` without releasing the intermediate dict, so it leaks on every call: `mpack_decode_frozendict`, `json_decode_frozendict`, `to_builtins_frozendict`, `convert_dict_to_frozendict`, and the frozendict branch in `convert_other`.

Reproduced on 3.15.0a7, counting live `dict` objects after 5000 ops:

```
decode -> dict         : +0
decode -> frozendict   : +5000
convert -> frozendict  : +5000
to_builtins frozendict : +5000
```

Fix at each site:

```c
PyObject *res = PyFrozenDict_New(dict);
Py_DECREF(dict);
return res;
```

(this also handles the case where `New` returns NULL). Worth a refcount regression test, since the functional tests pass straight through the leak.

**2. `msgspec.yaml.encode` raises on frozendict.**

`to_builtins` yields a frozendict, and PyYAML's `SafeDumper` has no representer for it:

```python
msgspec.yaml.encode(frozendict({"a": 1}))
# RepresenterError: ('cannot represent an object', frozendict({'a': 1}))
```

Same for a struct or dict containing a frozendict. (stdlib `json.dumps(to_builtins(...))` happens to work, so it is specific to consumers that do not special-case frozendict, but `yaml.encode` is public API.) The yaml path likely needs to emit a plain dict for frozendict. A yaml roundtrip test would catch it.

Both reproduced on a local 3.15.0a7 build of this branch.

Siyet (dismissed):
Both blockers fixed, re-verified against 3.15.0a7: `_PyFrozenDict_NewSteal` removes the refcount leak (5k iters/path, +0 vs control), and `to_builtins` returning a plain dict by default fixes `yaml.encode` (both `builtin_types` modes work). No dict regression from the dispatch refactor. Clearing changes-requested.

sobolevn:
@provinzkraut and here we go with flukes :(
Since this PR only touches 3.15+ and the test is executed on 3.14, there should not be any real effect.

provinzkraut:
> @provinzkraut and here we go with flukes :( Since this PR only touches 3.15+ and the test is executed on 3.14, there should not be any real effect.

Yeah. 

It's this: 

> Some benchmarks with significant performance changes were compared across different runtime environments,
which may affect the accuracy of the results.

Idk if there's a way around it, other than using self-hosted runners, to ensure we always run on the same thing.

provinzkraut:
@sobolevn I've opened https://github.com/CodSpeedHQ/codspeed/issues/409 asking for some support. Maybe we're missing something.

provinzkraut:
@sobolevn I've implemented some fixes in #1082 that seem to have made the benchmarks more stable!

sobolevn:
Awesome, now it shows that no perf changes are made :) 
Can we merge this?

docs/api.rst
ofek, line 174: Let's puts all dictionary types under here instead.

docs/usage.rst
```diff
@@ -183,8 +183,8 @@ representation.
     types: `bytes` / `bytearray` / `memoryview` to base64 string,
     `datetime.datetime` / `datetime.date` / `datetime.time` /
     `datetime.timedelta` to ISO 8601 string, `uuid.UUID` and
-    `decimal.Decimal` to string, `set` / `frozenset` to `list`, `enum.Enum`
-    to its member value.
+    `decimal.Decimal` to string, `set` / `frozenset` to `list`,
+    ``frozendict`` to `dict`, `enum.Enum` to its member value.
```
ofek, line 187: We also need to document support on this page under `frozenset` https://pr-1052.msgspec.pages.dev/supported-types

src/msgspec/_core.c
```diff
@@ -20528,7 +20620,7 @@ PyDoc_STRVAR(msgspec_to_builtins__doc__,
 "builtin_types: Iterable[type], optional\n"
 "    An iterable of types to treat as additional builtin types. These types will\n"
 "    be passed through ``to_builtins`` unchanged. Currently supports `bytes`,\n"
-"    `bytearray`, `memoryview`, `datetime.datetime`, `datetime.time`,\n"
+"    `bytearray`, `memoryview`, `frozendict`, `datetime.datetime`, `datetime.time`,\n"
```
ofek, line 20686: Should `frozenset` be supported or would that not make sense?

src/msgspec/inspect.py
```diff
@@ -472,6 +476,31 @@ class DictType(Type):
...
+    value_type: Type
+        The value type.
+    min_length: int, optional
+        If set, an instance of this type must have length greater than or equal
+        to ``min_length``.
+    max_length: int, optional
+        If set, an instance of this type must have length less than or equal
+        to ``max_length``.
+
+    Can only be emitted on Python 3.15+.
```
ofek: The rendering is broken for some reason https://pr-1052.msgspec.pages.dev/api#msgspec.inspect.FrozenDictType

Image

ofek (changes requested):
Thanks a lot!

src/msgspec/_core.c
provinzkraut: `to_builtins` currently treats `set` the same as `frozenset`. I think if we add one, we ought to add the other as well. 

IMO it feels a bit odd to have `frozendict` in here but not `dict`. I think it would be best to leave `frozendict` out for now, and revisit extending the supported types here in a separate PR

sobolevn: Yes, I would provide `set` and `frozenset` support in later PRs. Thanks for the idea! Currently they would be just converted to `list` objects (which is the correct default):

```python
>>> import msgspec
>>> msgspec.to_builtins([{1, 2}, frozenset({3, 4})])
[[1, 2], [3, 4]]
```

sobolevn:
@provinzkraut @Siyet @ofek please re-review :)

sobolevn:
I will fix the unrelated failure and then would love to merge this 🙏

sobolevn:
CI Fix: https://github.com/msgspec/msgspec/pull/1099

provinzkraut:
> @provinzkraut @Siyet @ofek please re-review :)

I'll take a look next week. My brain is currently offline due to the heat wave over europe and the non-existence of AC here :upside_down_face:

sobolevn:
@provinzkraut no worries, take care! 👍

sobolevn:
> Refusing to check out fork pull request code from a 'pull_request_target' workflow. This workflow runs with the base repository's GITHUB_TOKEN, secrets, default-branch cache scope, and runner access. Fetching and executing a fork's code in that trusted context commonly leads to "pwn request" vulnerabilities. To opt in, review the risks at https://gh.io/securely-using-pull_request_target and set 'allow-unsafe-pr-checkout: true' on the actions/checkout step.

This is a bug in our CI definition :(
We can't run this workflow from forks.
I will send a fix 😢

provinzkraut (dismissed):

docs/api.rst
```diff
@@ -174,12 +174,13 @@ Inspect
...
 .. autoclass:: TupleType
-.. autoclass:: DictType
 .. autoclass:: Field
-.. autoclass:: TypedDictType
 .. autoclass:: NamedTupleType
 .. autoclass:: DataclassType
 .. autoclass:: StructType
+.. autoclass:: DictType
+.. autoclass:: TypedDictType
+.. autoclass:: FrozenDictType
```
ofek: I un-resolved https://github.com/msgspec/msgspec/pull/1052#discussion_r3428749581 as this isn't the place I was talking about.

src/msgspec/inspect.py
```diff
@@ -461,6 +465,31 @@ class DictType(Type):
     max_length: int | None = None
 
 
+class FrozenDictType(Type):
```
ofek, line 468: Can we only define this on 3.15+?

tests/unit/test_common.py
```diff
@@ -41,6 +41,13 @@
 import msgspec
 from msgspec import UNSET, Meta, Struct, UnsetType, ValidationError
 
+try:
+    # This is needed for `ruff` to recognize `frozendict` name
+    # and to not raise `F821`:
+    from _future_builtins_ import frozendict
+except ImportError:
+    pass
```
ofek: Can you try conditionally importing it based on the Python version? That would be much cleaner and is what I'm used to doing to satisfy type checkers.

tests/unit/test_common.py
```diff
@@ -1069,6 +1076,37 @@ def test_err_union_with_multiple_dict_like_types(self, typ, proto):
         assert "more than one dict-like type" in str(rec.value)
         assert repr(typ) in str(rec.value)
 
+    @pytest.mark.skipif(
+        sys.version_info < (3, 15), reason="frozendict was added in 3.15"
+    )
```
ofek: Can we turn this into a utility fixture that can be imported rather than repeating everywhere?

src/msgspec/_core.c
```diff
@@ -2874,7 +2899,7 @@ AssocList_Sort(AssocList* list) {
  * O | TYPEDDICT | DATACLASS |
  * O | NAMEDTUPLE |
  * O | STR_REGEX |
- * T | DICT [key, value] |
+ * T | DICT | FROZENDICT [key, value] |
```
ofek: I know this is just documentation but asking because it's the first instance: should we maintain the key/value pair on both?

src/msgspec/_core.c
```diff
@@ -5108,10 +5134,23 @@ typenode_collect_type(TypeNodeCollectState *state, PyObject *obj) {
         if (args != NULL && PyTuple_GET_SIZE(args) != 2) goto invalid;
         out = typenode_collect_dict(
             state,
+            MS_TYPE_DICT,
             (args == NULL) ? state->mod->typing_any : PyTuple_GET_ITEM(args, 0),
             (args == NULL) ? state->mod->typing_any : PyTuple_GET_ITEM(args, 1)
         );
     }
+#if PY315_PLUS
+    else if (origin == (PyObject*)(&PyFrozenDict_Type)) {
```
ofek, line 5152: For performance reasons, do you think we should move this down one? I think a list would be far more likely.

ofek (commented):
Thanks for working on this! I wish we had better benchmarks (or added back the old profiling test dependency) so that we could assess the impact on performance.

src/msgspec/inspect.py
sobolevn: I don't think that this is a good idea. Users might want to provide an API based on these types. It would require them to use conditional imports as well. I don't think that this extra complexity is needed. Saying that it won't be issued is enough IMO.

sobolevn:
Done! Thanks a lot for the review!

tests/unit/test_common.py
```diff
@@ -29,9 +29,19 @@
     Union,
 )
 
+if sys.version_info >= (3, 15):
```
ofek: Nit: is it possible to move these after the normal imports?

ofek (dismissed):

src/msgspec/inspect.py
ofek: For typing purposes, shouldn't we act as the responsible party here and enforce that? It seems to me that it's more correct to conditionally define this (just as our extension has conditional support) and far better to have static analysis catch errors rather than tests of runtime behavior.

sobolevn:
(I need one more approve now after the changes) 😅

tests/unit/test_common.py
```diff
@@ -4976,6 +4286,39 @@ def test_order_dict_unsortable(self, proto):
...
+        sys.version_info < (3, 15), reason="frozendict was added in 3.15"
+    )
+    @pytest.mark.parametrize("msg", [{}, {"y": 1, "x": 2, "z": 3}])
+    @pytest.mark.parametrize("order", [None, "deterministic", "sorted"])
+    @pytest.mark.parametrize("use_encoder", [False, True])
+    def test_order_frozendict(self, msg, order, use_encoder, proto):
+        if use_encoder:
+            res = proto.Encoder(order=order).encode(msg)
+        else:
+            res = proto.encode(msg, order=order)
```
vqko, line 4299: Shouldn't `msg` be converted to frozendict here?

sobolevn: Fixed in https://github.com/msgspec/msgspec/pull/1107

src/msgspec/inspect.py
provinzkraut: What runtime errors could arise from this?

Since this is inspection code, pre 3.15, you just never get back this type.
However, for consumers of this inspection, a conditional define would be awkward, e.g.

```python
if isinstance(o, (inspect.FrozenDictType, inspect.DictType)):
 ...
```
would have to become
```python
if isinstance(o, inspect.DictType) or (sys.version_info >= (3, 15) and isinstance(o, inspect.FrozenDictType)):
 ...
```

PR #1053, opened 2026-06-12
Fix `msgspec.msgpack.Ext` types

sobolevn:
Several typing improvments:
1. `msgspec.msgpack.Enc` now accepts all buffers, not just `bytes | memoryview | bytearray`. I've added a runtime test to proove that
2. Renamed `T` and other typing-only utils to be protected names: `T` and `_enc_hook_sig` can't be imported from `msgspec.msgpack` in runtime

### DIFF
--- a/src/msgspec/msgpack.pyi
+++ b/src/msgspec/msgpack.pyi
@@ -14,9 +14,9 @@ from typing_extensions import Buffer
 
 _T = TypeVar("_T")
 
-_enc_hook_sig: TypeAlias = Callable[[Any], Any] | None
-_ext_hook_sig: TypeAlias = Callable[[int, memoryview], Any] | None
-_dec_hook_sig: TypeAlias = Callable[[type, Any], Any] | None
+_EncHookSig: TypeAlias = Callable[[Any], Any] | None
+_ExtHookSig: TypeAlias = Callable[[int, memoryview], Any] | None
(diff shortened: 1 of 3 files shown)


provinzkraut:
Same things seem to apply to `json.pyi`. Do you wanna add that to this PR?

sobolevn:
I would prefer to analyze all modules one by one, there are multiple type errors there as well :)

src/msgspec/msgpack.pyi
```diff
@@ -12,66 +12,66 @@ from typing import (
 
 from typing_extensions import Buffer
 
-T = TypeVar("T")
+_T = TypeVar("_T")
 
-enc_hook_sig: TypeAlias = Callable[[Any], Any] | None
-ext_hook_sig: TypeAlias = Callable[[int, memoryview], Any] | None
-dec_hook_sig: TypeAlias = Callable[[type, Any], Any] | None
+_enc_hook_sig: TypeAlias = Callable[[Any], Any] | None
```
provinzkraut: Maybe we could also convert this to the more canonical camel case, i.e. `_EncHookSig`?

sobolevn: Oh, right 🤦‍♂️ 
How did I miss that!

provinzkraut (approved):

PR #1062, opened 2026-06-13
Add missing `__class_getitem__` methods to typing

sobolevn:
Runtime:

```python
>>> import msgspec
>>> msgspec.json.Decoder.__class_getitem__((int,))
msgspec.json.Decoder[int]
```

Mypy:

```
» mypy -c 'import msgspec; msgspec.json.Decoder.__class_getitem__'
<string>:1: error: "type[Decoder[_T]]" has no attribute "__class_getitem__"  [attr-defined]
Found 1 error in 1 file (checked 1 source file)
```

Now this symbol would be defined. 
I used the same definition that we use in `typeshed`: https://github.com/python/typeshed/blob/7e3b6cd8101ec0b80eedffa9a0ffd949f3ea88ab/stdlib/queue.pyi#L49

### DIFF
--- a/src/msgspec/json.pyi
+++ b/src/msgspec/json.pyi
@@ -1,4 +1,5 @@
 from collections.abc import Callable, Iterable
+from types import GenericAlias
 from typing import (
     Any,
     Generic,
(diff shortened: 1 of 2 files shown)


Siyet:
Thanks @sobolevn! Verified the stub matches runtime: both `msgspec.json.Decoder` and `msgspec.msgpack.Decoder` register `__class_getitem__` via `Py_GenericAlias` (`METH_CLASS`) in `_core.c`, and at runtime `Decoder[int]` returns a real `types.GenericAlias`. These two are the only `Generic` classes in the stubs (`Encoder` is not generic), so coverage is complete. Merging.

PR #1065, opened 2026-06-14
Fix `src/msgspec/structs.pyi` typings

sobolevn:


.pytest.toml
```diff
@@ -3,7 +3,6 @@
 # We default to all non-profiling test suites and we run unit tests first.
 testpaths = [
   "tests/unit",
-  "tests/typing",
```
provinzkraut, line 6: I think they should still run, to prove they're correct both at runtime and during type checking.

sobolevn: They don't actually run :(
Because they are inside `check_` functions, which are not tests. Moreover, `assert_type` does nothing in runtime. Moreover, `pytest` even does not collect the file itself, because it is named `basic_typing_examples.py` not `test_*.py`

```python
def assert_type(val, typ, /):
    return val
```

It used to have `test_mypy.py` and `test_pyright.py` test files in this package. Now they are gone and this is no longer needed.

provinzkraut (approved):

PR #1068, opened 2026-06-14
Add `.editorconfig` file

sobolevn:
This file is really useful for better DX. For example, it allows users to have TAB key work properly in different files and insert just the right amount of spaces instead. One of the few plugins for text editors that I actually use.

Docs: https://editorconfig.org/

I copied this file from CPython: https://github.com/python/cpython/blob/main/.editorconfig with modifications and current project's styles.


Siyet (dismissed):
Looks good, CI is green and the config is sane (4-space for py/c/h/rst, 2-space for yaml/toml/md, trailing-whitespace + final-newline trimming where it matters).

One optional nit: the `[*.{yml,yaml,...}]` section sets `indent_size` but not `indent_style = space`. YAML can't use tabs at all, so making the style explicit there avoids any editor defaulting to tabs. Not blocking.

PR #1070, opened 2026-06-15
Support `pyrefly` for typing tests

sobolevn:


tests/typing/basic_typing_examples.py
```diff
@@ -29,7 +29,8 @@ def check_unset() -> None:
 
 def check_unset_type_lowering(x: int | msgspec.UnsetType) -> None:
     if x is msgspec.UNSET:
-        assert_type(x, Literal[msgspec.UnsetType.UNSET])
+        # this is a bug in `pyrefly`:
```
ofek: Let's provide a link to an issue (opening one if necessary) so that we know when to remove this hack.

pyproject.toml
```diff
@@ -140,9 +141,17 @@ typeCheckingMode = "strict"
...
 reportUnusedClass = false
 reportPrivateUsage = false
+
+[tool.pyrefly]
+project-includes = ["tests/typing/"]
+# Strict mode:
+preset = 'strict'
+min-severity = 'warn'
+# Don't use '# type: ignore' to suppress with pyright:
+enabled-ignores = ["pyrefly"]
```
ofek, line 157: How does this interact with our upstream users type checking their own code bases?

sobolevn: it does not, this configuration is only for our own use in tests.

tests/typing/basic_typing_examples.py
sobolevn: I will open an issue, because I report back all findings in `pyrefly` :)
But, it will be autoremoved once the bug will be fixed, because all unused ignores will be reported :) 

See https://pyrefly.org/en/docs/configuration/#preset-strict It enables `unused-ignore` as an error.

sobolevn: Done: https://github.com/facebook/pyrefly/issues/3820

sobolevn: And https://github.com/facebook/pyrefly/issues/3821

ofek: I forgot about the unused ignore safety net; it's just a habit of mine to avoid stale comments. Thank you!

ofek (approved):

PR #1073, opened 2026-06-16
Limit decode nesting depth to prevent C stack overflow

### LINKED ISSUE #1033: Stack overflowed when testing test_JSONTestSuite.py on system with small stack size
### Description

System: Windows 1607
Python: 3.10.10
msgspec: 0.21.1

```
(star) PS E:\ProgramData\Pycharm\msgspec> python -m pytest tests/unit
====================================================================== test session starts ======================================================================
platform win32 -- Python 3.10.10, pytest-8.3.5, pluggy-1.5.0
rootdir: E:\ProgramData\Pycharm\msgspec
configfile: .pytest.ini
plugins: anyio-3.6.2
collected 6210 items                                                                                                                                             

tests\unit\test_JSONTestSuite.py ...............................................................................................................Windows fatal exce
ption: stack overflow

Current thread 0x0001b19c (most recent call first):
  File "E:\ProgramData\Pycharm\msgspec\tests\unit\test_JSONTestSuite.py", line 313 in test_invalid
  File "E:\ProgramFiles\Anaconda3\envs\star\lib\site-packages\_pytest\python.py", line 159 in pytest_pyfunc_call
  File "E:\ProgramFiles\Anaconda3\envs\star\lib\site-packages\pluggy\_callers.py", line 103 in _multicall
  File "E:\ProgramFiles\Anaconda3\envs\star\lib\site-packages\pluggy\_manager.py", line 120 in _hookexec
  File "E:\ProgramFiles\Anaconda3\envs\star\lib\site-packages\pluggy\_hooks.py", line 513 in __call__
  File "E:\ProgramFiles\Anaconda3\envs\star\lib\site-packages\_pytest\python.py", line 1627 in runtest
  File "E:\ProgramFiles\Anaconda3\envs\star\lib\site-packages\_pytest\runner.py", line 174 in pytest_runtest_call
  File "E:\ProgramFiles\Anaconda3\envs\star\lib\site-packages\pluggy\_callers.py", line 103 in _multicall
  File "E:\ProgramFiles\Anaconda3\envs\star\lib\site-packages\pluggy\_manager.py", line 120 in _hookexec
  File "E:\ProgramFiles\Anaconda3\envs\star\lib\site-packages\pluggy\_hooks.py", line 513 in __call__
  File "E:\ProgramFiles\Anaconda3\envs\star\lib\site-packages\_pytest\runner.py", line 242 in <lambda>
  File "E:\ProgramFiles\Anaconda3\envs\star\lib\site-packages\_pytest\runner.py", line 341 in from_call
  File "E:\ProgramFiles\Anaconda3\envs\star\lib\site-packages\_pytest\runner.py", line 241 in call_and_report
  File "E:\ProgramFiles\Anaconda3\envs\star\lib\site-packages\_pytest\runner.py", line 132 in runtestprotocol
  File "E:\ProgramFiles\Anaconda3\envs\star\lib\site-packages\_pytest\runner.py", line 113 in pytest_runtest_protocol
  File "E:\ProgramFiles\Anaconda3\envs\star\lib\site-packages\pluggy\_callers.py", line 103 in _multicall
  File "E:\ProgramFiles\Anaconda3\envs\star\lib\site-packages\pluggy\_manager.py", line 120 in _hookexec
  File "E:\ProgramFiles\Anaconda3\envs\star\lib\site-packages\pluggy\_hooks.py", line 513 in __call__
  File "E:\ProgramFiles\Anaconda3\envs\star\lib\site-packages\_pytest\main.py", line 362 in pytest_runtestloop
  File "E:\ProgramFiles\Anaconda3\envs\star\lib\site-packages\pluggy\_callers.py", line 103 in _multicall
  File "E:\ProgramFiles\Anaconda3\envs\star\lib\site-packages\pluggy\_manager.py", line 120 in _hookexec
  File "E:\ProgramFiles\Anaconda3\envs\star\lib\site-packages\pluggy\_hooks.py", line 513 in __call__
  File "E:\ProgramFiles\Anaconda3\envs\star\lib\site-packages\_pytest\main.py", line 337 in _main
  File "E:\ProgramFiles\Anaconda3\envs\star\lib\site-packages\_pytest\main.py", line 283 in wrap_session
  File "E:\ProgramFiles\Anaconda3\envs\star\lib\site-packages\_pytest\main.py", line 330 in pytest_cmdline_main
  File "E:\ProgramFiles\Anaconda3\envs\star\lib\site-packages\pluggy\_callers.py", line 103 in _multicall
  File "E:\ProgramFiles\Anaconda3\envs\star\lib\site-packages\pluggy\_manager.py", line 120 in _hookexec
  File "E:\ProgramFiles\Anaconda3\envs\star\lib\site-packages\pluggy\_hooks.py", line 513 in __call__
  File "E:\ProgramFiles\Anaconda3\envs\star\lib\site-packages\_pytest\config\__init__.py", line 175 in main
  File "E:\ProgramFiles\Anaconda3\envs\star\lib\site-packages\_pytest\config\__init__.py", line 201 in console_main
  File "E:\ProgramFiles\Anaconda3\envs\star\lib\site-packages\pytest\__main__.py", line 9 in <module>
  File "E:\ProgramFiles\Anaconda3\envs\star\lib\runpy.py", line 86 in _run_code
  File "E:\ProgramFiles\Anaconda3\envs\star\lib\runpy.py", line 196 in _run_module_as_main
(star) PS E:\ProgramData\Pycharm\msgspec>
```

### Reproduce

1. Run `python -m pytest tests/unit` on system with small stack size, e.g. old windows, embeded devices.
  On system with normal stack size, increase recursion limit at the head of test file before running tests.
```python
import sys
sys.setrecursionlimit(40000)
```
2. For each `[]`, msgspec enters multiple stacks:
```
json_decode_list -> json_decode -> json_decode_nocustom -> json_decode_array -> json_decode_list -> ...
```
3. Stack overflow occurs before reaching python's recursion limit.

This could lead to DoS security risk in some cases, as attackers can send a deeply recursive json to shutdown server

Siyet:
Closes #1033.

The JSON and MessagePack decoders recurse once per nesting level, using a few C stack frames each. The only guard is `Py_EnterRecursiveCall`, which on Python < 3.14 is tied to `sys.getrecursionlimit()` (or the C recursion limit) and does not track the actual remaining C stack. So deeply nested input can overflow the C stack and crash the process (SIGSEGV, no catchable exception) before the guard fires, either on a thread with a small stack or when the recursion limit has been raised. This is a DoS vector for code decoding untrusted input (same class as orjson's CVE-2024-27454).

This adds a hard nesting-depth cap (`MS_MAX_DECODE_DEPTH = 512`) independent of `sys.getrecursionlimit()`, raising a `DecodeError` instead of recursing further. It covers both decoders and the unknown-field skip paths.

Python 3.14 rewrote the recursion guard to check the actual remaining C stack, so it already prevents the overflow there. The explicit cap is therefore only applied on Python < 3.14; on 3.14+ the decoder defers to CPython's guard and the cap is compiled out.

On the cap value: 512 sits below the default Python recursion limit (~1000), so on the affected versions this cap, rather than a `RecursionError`, is what bounds nesting consistently. It's also low enough to keep a realistic small stack (e.g. a default 1 MB Windows main thread) safe, while staying far above any real-world nesting depth. orjson/simdjson use 1024, but that assumes a large server stack; the point of #1033 is that a recursion-limit-tied guard is insufficient on small stacks, so a lower fixed cap is used here. Happy to adjust the value.

Note: TOML/YAML decode goes through pure-Python `tomllib`/PyYAML, not this C path, so they're unaffected (and have their own limits).

### DIFF
--- a/src/msgspec/_core.c
+++ b/src/msgspec/_core.c
@@ -14876,16 +14876,21 @@ typedef struct DecoderState {
     Py_ssize_t recursion_depth;
 } DecoderState;
 
-/* Maximum nesting depth allowed when decoding. The recursive decoders use a few
- * C stack frames per level, and CPython's recursion guard is tied to
+/* Bound nesting depth when decoding. The recursive decoders use a few C stack
+ * frames per level, and CPython's recursion guard is tied to
  * `sys.getrecursionlimit()` (or the C recursion limit), neither of which tracks
  * the actual remaining C stack. On systems with a small stack (or when the
  * recursion limit is raised) deeply nested input can overflow the C stack and
  * crash the process before the guard fires. A hard cap independent of the
  * recursion limit prevents that. The value sits below the default Python
  * recursion limit (so this cap, rather than a `RecursionError`, is what bounds
- * nesting consistently across Python versions) while staying far above any
- * realistic nesting depth. */
+ * nesting consistently across versions) while staying far above any realistic
+ * nesting depth.
+ *
+ * Python 3.14+ rewrote that guard to check the actual remaining C stack, so it
+ * already prevents the overflow there; the explicit cap is only needed on older
+ * versions. */
+#if PY_VERSION_HEX < 0x030E0000
 #define MS_MAX_DECODE_DEPTH 512
 
 /* Enter a recursive decode level, returning -1 (with a DecodeError set) if the
@@ -14908,6 +14913,18 @@ static MS_INLINE void
 ms_leave_recursive(Py_ssize_t *depth) {
     (*depth)--;
 }
+#else
+static MS_INLINE int
+ms_enter_recursive(Py_ssize_t *depth) {
+    (void)depth;
+    return 0;
+}
+
+static MS_INLINE void
+ms_leave_recursive(Py_ssize_t *depth) {
+    (void)depth;
+}
+#endif
 
 typedef struct Decoder {
     PyObject_HEAD
--- a/src/msgspec/_core.c
+++ b/src/msgspec/_core.c
@@ -14882,8 +14882,11 @@ typedef struct DecoderState {
  * the actual remaining C stack. On systems with a small stack (or when the
  * recursion limit is raised) deeply nested input can overflow the C stack and
  * crash the process before the guard fires. A hard cap independent of the
- * recursion limit prevents that. Matches the limit used by orjson/simdjson. */
-#define MS_MAX_DECODE_DEPTH 1024
+ * recursion limit prevents that. The value sits below the default Python
+ * recursion limit (so this cap, rather than a `RecursionError`, is what bounds
+ * nesting consistently across Python versions) while staying far above any
+ * realistic nesting depth. */
+#define MS_MAX_DECODE_DEPTH 512
 
 /* Enter a recursive decode level, returning -1 (with a DecodeError set) if the
  * nesting limit would be exceeded, else 0. Each successful call must be paired
--- a/src/msgspec/_core.c
+++ b/src/msgspec/_core.c
@@ -14877,25 +14877,21 @@ typedef struct DecoderState {
 } DecoderState;
 
 /* Bound nesting depth when decoding. The recursive decoders use a few C stack
- * frames per level, and CPython's recursion guard is tied to
- * `sys.getrecursionlimit()` (or the C recursion limit), neither of which tracks
- * the actual remaining C stack. On systems with a small stack (or when the
- * recursion limit is raised) deeply nested input can overflow the C stack and
- * crash the process before the guard fires. A hard cap independent of the
- * recursion limit prevents that. The value sits below the default Python
- * recursion limit (so this cap, rather than a `RecursionError`, is what bounds
- * nesting consistently across versions) while staying far above any realistic
- * nesting depth.
+ * frames per level, and CPython's recursion guard (tied to the recursion limit)
+ * doesn't track the actual remaining C stack, so on a small stack (or a raised
+ * recursion limit) deeply nested input can overflow the C stack and crash the
+ * process. A hard cap independent of the recursion limit prevents that. 512
+ * sits below the default recursion limit (so the cap, not a `RecursionError`,
+ * bounds nesting consistently across versions) yet far above any realistic
+ * depth. Python 3.14+ rewrote that guard to track the real C stack, so the cap
+ * is only needed on older versions.
  *
- * Python 3.14+ rewrote that guard to check the actual remaining C stack, so it
- * already prevents the overflow there; the explicit cap is only needed on older
- * versions. */
+ * `MS_DEPTH_WRAP_OBJ`/`_INT` wrap a recursive decode function `f`: the body is
+ * defined as `f_inner` and entered only after bumping `self->recursion_depth`,
+ * which is restored on return. On 3.14+ they expand to a plain definition. */
 #if PY_VERSION_HEX < 0x030E0000
 #define MS_MAX_DECODE_DEPTH 512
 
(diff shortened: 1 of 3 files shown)


src/msgspec/_core.c
```diff
@@ -14873,8 +14873,55 @@ typedef struct DecoderState {
...
+ * defined as `f_inner` and entered only after bumping `self->recursion_depth`,
+ * which is restored on return. On 3.14+ they expand to a plain definition. */
+#if PY_VERSION_HEX < 0x030E0000
+#define MS_MAX_DECODE_DEPTH 512
+
+static MS_INLINE int
+ms_enter_recursive(Py_ssize_t *depth) {
+    if (MS_UNLIKELY(*depth >= MS_MAX_DECODE_DEPTH)) {
+        PyErr_SetString(
+            msgspec_get_global_state()->DecodeError,
```
sobolevn, line 14899: Maybe `RecursionError` would make more sence? It might be a braking change technically, but this is something we can think about.

src/msgspec/_core.c
```diff
@@ -15944,9 +15990,10 @@ mpack_decode_struct_array_union(
     return mpack_decode_struct_array_inner(self, size, true, info, path, is_key);
 }
 
-static PyObject *
-mpack_decode_array(
-    DecoderState *self, Py_ssize_t size, TypeNode *type, PathNode *path, bool is_key
+MS_DEPTH_WRAP_OBJ(
```
sobolevn, line 15993: What if instead of a custom function definition macros we can just use functions?

It would be

```c
#if PY313_OR_LESS
static PyObject *
check_c_recursion(DecoderState *self) {
     if (MS_UNLIKELY(ms_enter_recursive(&self->recursion_depth))) {
        PyErr_SetString(
            msgspec_get_global_state()->DecodeError,
            "Input data is too deeply nested"
        );
        return NULL;
    }
}
#endif

// we can inline it as well, if this would make sense?
static PyObject *
mpack_decode_array_recursion_checked(
    DecoderState *self, Py_ssize_t size, TypeNode *type, PathNode *path, bool is_key
) {
   // old code from `mpack_decode_array`
}

static PyObject *
mpack_decode_array(
    DecoderState *self, Py_ssize_t size, TypeNode *type, PathNode *path, bool is_key
) {
   
#if PY313_OR_LESS
    if (check_c_recursion(self)) return NULL;
#endif
    return mpack_decode_array_recursion_checked(...);
}
```

This way we just keep things simple enough. Functions are always better than macros if possible.

jcrist:
Why is the system recursion limit not sufficient here?

- The system limit takes into account the total call stack depth, not just the depth within `msgspec` itself.
- On typical systems, the system stack default is large enough to prevent stack overflows
- On systems with small stacks (like in #1033) I'd argue the user should configure a smaller recursion limit with `sys.setrecursionlimit` instead. This handles cases outside of `msgspec` as well, and remains configurable.

Hardcoding a limit in `msgspec` (and depth handling) adds more branching on our side, and only takes into account the stack usage within a single call to `msgspec`. As you said, in Python 3.14 the default recursion limit handling in CPython is improved (making this less of a problem); for earlier python versions I'd argue this is a problem outside the domain of `msgspec` to resolve.

provinzkraut:
Agree with @jcrist. This isn't really due to a bug or misbehaviour on msgspec's side, it's just Python behaviour you'd get in other cases as well.

ofek:
I agree with what has been said and think that we should not impose such a limit.

Siyet:
Consensus is clear, closing. Thanks everyone for weighing in.

I'll leave a summary in #1033 as well.

PR #1075, opened 2026-06-16
Set up continous benchmarking

### LINKED ISSUE #310: Set up continuous benchmarking

provinzkraut:
Set up continous benchmarking via codespee.io. Closes #310.
For reasoning and decision making process, refer to https://github.com/msgspec/msgspec/issues/310#issuecomment-4701926566

<hr>

Re-implement existing benchmarks in `/benchmark` as pytest style benchmarks in `/tests/prof/perf`, so they work with codspeed. I've kept around the existing benchmarks, for local use and to produce comparisons against other libraries. Over time we might chose to consolidate them, but I think we should evaluate codspeed for a bit before doing so.

- Performance benchmarks on Linux amd64 + arm using codspeeds [CPU instrumentation](https://codspeed.io/docs/instruments/cpu/index)
- Additional memory benchmarks on amd64

**Limitations**

Benchmarks only run on Linux, because codspeeds instrumentation requires a Linux toolchain. Wallclock benchmarking is available for macOS as well, however, much larger variance is expected there.


.github/workflows/ci.yml
```diff
@@ -18,6 +18,7 @@ on:
     - .github/workflows/prepare-build.yml
     - .github/workflows/build-wheels.yml
     - .github/workflows/profile.yml
+    - .github/workflows/codspeed.yml
```
sobolevn:
```suggestion
    - .github/workflows/*
```

Let's just use one rule for all files.

tests/prof/perf/test_structs.py
```diff
@@ -0,0 +1,87 @@
...
+
+    def fn():
+        return defstruct("SomeStruct", fields)
+
+    res = benchmark(fn)
+    assert issubclass(res, msgspec.Struct)
+
+
+@pytest.mark.memory
+def test_create(benchmark):
```
sobolevn, line 105: Let's also test `tag=` logic here, because it is quite complex / important.

tests/prof/perf/test_structs.py
```diff
@@ -0,0 +1,87 @@
...
+class C{n}(Struct, order=True):
+    a: int
+    b: int
+    c: int
+    d: int
+    e: int
+"""
+
+N = 1000
+RAMDOM = random.Random(b"12345")
```
sobolevn: Please, add a `TODO` to use `pytest-randomnly` instead

sobolevn (commented):
Awesome! I think that we also need a `import msgspec` timing test! 👍

tests/prof/perf/test_structs.py
provinzkraut: Added `pytest-randomly`

provinzkraut:
> Awesome! I think that we also need a `import msgspec` timing test! 👍

Added, with some caveats

tests/prof/perf/test_module.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+def test_import_time(benchmark):
+    # also measures interpreter startup etc., but is the best we can do now, since we
+    # don't have a fully isolated module, and some per-process global state, so
+    # re-importing purely within the current interpreter doesn't get reliable timings
+    def do_import():
+        subprocess.run([sys.executable, "-c", "import msgspec"], check=True)
+
+    do_import()  # execute once to reduce possibility of a suuper cold FS cache
```
sobolevn:
```suggestion
    do_import()  # execute once to reduce possibility of a super cold FS cache
```

sobolevn (dismissed):
Awesome! Thanks a lot for your work!

provinzkraut:
Mhm. Seems like codspeed is indeed a bit unstable: https://github.com/msgspec/msgspec/pull/962#issuecomment-4721378754, because checks are running in different envs. Re-ran these 3x now. When run in the same env, they do same stable, but getting the same env as the previous runs, with 3 jobs seems like quite a bit of a gamble. Could probably improve this with self-hosted runners.

ofek:
I tried to run the manual profiling job to check if the results are similar to Codspeed but it doesn't work now https://github.com/msgspec/msgspec/actions/runs/27883637649/job/82515828562

provinzkraut:
> I tried to run the manual profiling job to check if the results are similar to Codspeed but it doesn't work now https://github.com/msgspec/msgspec/actions/runs/27883637649/job/82515828562

Probably because `pytest-benchmark` had to be replaced with `pytest-codspeed`, which isn't a 100% drop in? 

Also, the results won't be comparable at all, since the codspeed tests run via the [CPU simulation](https://codspeed.io/docs/instruments/cpu)

provinzkraut:
So, I set up the threshold to 2%, down from the 10% default, and now pretty much every PR is failing again. Seems that there's still a lot of variance.

@ofek can we try setting up macro runners? I don't have the permissions for that: https://codspeed.io/docs/features/macro-runners#repository-access-configuration

ofek:
I just followed that guide so everything should be set up now. Yes, I don't have high confidence due to the current variance and was trying to run the old workflow to verify.

provinzkraut:
> Yes, I don't have high confidence due to the current variance and was trying to run the old workflow to verify.

I feel the same. Which is a shame because otherwise codspeed is really nice.

ofek:
Do you think using the same measurement framework as CPython would help us with our micro-benchmarks? https://github.com/psf/pyperf

ofek:
I'm not sure what we can do on GitHub-provided runners but what about some of the advice here? https://google.github.io/benchmark/reducing_variance.html

I found a nice quickstart setup gist https://gist.github.com/itzmeanjan/05dc3e946f635d00c5e0b21aae6203a7

provinzkraut:
> Do you think using the same measurement framework as CPython would help us with our micro-benchmarks? https://github.com/psf/pyperf

Unfortunately not, as on-machine variance isn't our issue. As long as we get the same runner hardware, the results are stable. What tools like these can help with is stabilising results across runs on the same hardware, which codspeed already does through similar means.

> I'm not sure what we can do on GitHub-provided runners but what about some of the advice here? https://google.github.io/benchmark/reducing_variance.html

I'll check those out!

PR #1076, opened 2026-06-16
Document that omit_defaults ignores custom default_factory

### LINKED ISSUE #645: Allow `omit_defaults` to exclude fields when encoded value is `{}` (empty dict)
### Description

The following code:

```python
import msgspec as ms


class Sub(ms.Struct, omit_defaults=True):
    f1: str = ""
    f2: int = 0


class Struct(ms.Struct, omit_defaults=True, tag=True):
    f1: int = 1
    sub: Sub = ms.field(default_factory=Sub)


print(ms.yaml.encode(Struct()).decode())
print(ms.json.encode(Struct()).decode())

y = """
type: Struct

"""
j = '{"type":"Struct"}'

print()
print(ms.yaml.decode(y, type=Struct))
print(ms.json.decode(j, type=Struct))

```

Will output:

```python
type: Struct
sub: {}

{"type":"Struct","sub":{}}

Struct(f1=1, sub=Sub(f1='', f2=0))
Struct(f1=1, sub=Sub(f1='', f2=0))
```

Although it is clear [why](https://jcristharif.com/msgspec/structs.html#omitting-default-values) the empty dict is being encoded, it is still desirable to avoid that, given that decoding absent fields does not cause any errors when the empty constructor (i.e. empty dict) is producing the expected default.

It would seem that recursive check with `matches_default` on subclasses of `msgspec.Struct` might do the job, but this would probably mean some hit on perfomance (?).

Perhaps there is a way to check the resulting `'{}'` value and omit it instead?

Thank you for considering this.

### LINKED ISSUE #1032: Default not omitted with custom default factory
### Description

In my project, I have structs that contains list fields. I would like to omit these fields when the list is empty to reduce the size of the saved files.

When using the `list` constructor as the default factory or a default empty list, it works fine. However, the former conflicts with the type linter, and I want to avoid the latter (shared across all instances).

When I use a custom factory function, the field always appear in the output.

It may be the correct behavior, but I did not find any mention of this in the docs.

### Reproduction

I ran this using Python 3.14 and msgspec 0.21.1.

```py

import msgspec


class Fruit(msgspec.Struct):
    name: str = msgspec.field(name="name")


def fruits_factory() -> list[Fruit]:
    return []


class Basket(msgspec.Struct, omit_defaults=True):
    size: int = msgspec.field(name="size")
    material: str = msgspec.field(name="material", default="wicker")
    fruits: list[Fruit] = msgspec.field(name="fruits", default_factory=fruits_factory)


class BasketWithListConstructor(msgspec.Struct, omit_defaults=True):
    size: int = msgspec.field(name="size")
    material: str = msgspec.field(name="material", default="wicker")
    fruits: list[Fruit] = msgspec.field(name="fruits", default_factory=list)  # Type error because list is untyped


class BasketWithDefaultValue(msgspec.Struct, omit_defaults=True):
    size: int = msgspec.field(name="size")
    material: str = msgspec.field(name="material", default="wicker")
    fruits: list[Fruit] = msgspec.field(name="fruits", default=[])  # Shared mutable default value, should be avoided


if __name__ == "__main__":
    basket = Basket(size=5)

    print(msgspec.json.encode(basket))
    # Expected output: b'{"size":5}'
    # Actual output: b'{"size":5,"fruits":[]}'

    basket_with_list_constructor = BasketWithListConstructor(size=5)
    print(msgspec.json.encode(basket_with_list_constructor))
    # Output: b'{"size":5}'

    basket_with_default_value = BasketWithDefaultValue(size=5)
    print(msgspec.json.encode(basket_with_default_value))
    # Output: b'{"size":5}'

    print(f"Using msgspec {msgspec.__version__}")
```

Siyet:
`omit_defaults` only recognizes empty-collection defaults when the default is given by value or via a builtin collection constructor (`list`/`dict`/`set`/`tuple`/`frozenset`). A custom `default_factory` (a user function, `lambda`, or a `Struct`/`dataclass`/`attrs` type) is never called during detection, so its field is always encoded, even when the produced value is empty.

This is intentional: detection runs in the encode hot path and stays O(1) without invoking user code. The behavior wasn't documented though, so the `matches_default` section now spells out the `default_factory` limitation and points at `default_factory=list` (which type checks fine via the field annotation) as the supported way to omit an empty collection default.

Closes #1032
Closes #645

### DIFF
--- a/docs/structs.rst
+++ b/docs/structs.rst
@@ -677,6 +677,23 @@ detection logic is as follows:
     ...         return True
     ...     return False
 
+This detection never calls a ``default_factory``. A field configured with a
+custom ``default_factory`` is only omitted when the factory is one of the
+builtin collection constructors (``list``, ``dict``, ``set``, ``tuple``, or
+``frozenset``). Any other callable (a user-defined function, a ``lambda``, or a
+``Struct``/``dataclass``/``attrs`` type) is treated as opaque, so the field is
+always encoded, even when the value it produces is empty. To omit an empty
+collection default, configure the builtin constructor directly:
+
+.. code-block:: python
+
+    >>> class Basket(msgspec.Struct, omit_defaults=True):
+    ...     items: list[int] = msgspec.field(default_factory=list)
+
+The field annotation supplies the element type, so ``default_factory=list``
+still type checks. Specifying ``default=[]`` works too: ``msgspec`` doesn't
+share mutable default values between instances.
+
 
 .. _forbid-unknown-fields:
 


docs/structs.rst
```diff
@@ -677,6 +677,23 @@ detection logic is as follows:
...
+collection default, configure the builtin constructor directly:
+
+.. code-block:: python
+
+    >>> class Basket(msgspec.Struct, omit_defaults=True):
+    ...     items: list[int] = msgspec.field(default_factory=list)
+
+The field annotation supplies the element type, so ``default_factory=list``
+still type checks. Specifying ``default=[]`` works too: ``msgspec`` doesn't
+share mutable default values between instances.
```
provinzkraut, line 695: This isn't true and can be misleading. It *is* true for builtin mutable collections, but it is not true for any mutable object.

with a list:
```python
>>> import msgspec
>>> DEFAULT = []
>>> class A(msgspec.Struct):
...     field: list[int] = msgspec.field(default=DEFAULT)
... 
>>> print(A().field is A().field)
False
```

with a custom class: 
```python
>>> import msgspec
>>> from collections import UserDict
>>> class MyDict(UserDict):
...     pass
... 
>>> DEFAULT = MyDict()
>>> class A(msgspec.Struct):
...     field: dict[str, int] = msgspec.field(default=DEFAULT)
... 
>>> print(A().field is A().field)
True
```

PR #1077, opened 2026-06-16
Fix compilation warnings on Python 3.15

### LINKED ISSUE #1048: Compilation warnings on 3.15
There are three warnings that I get on `3.15-dev`:

```
      src/msgspec/_core.c:3691:24: warning: shifting a negative signed value is undefined [-Wshift-negative-value]
       3691 |         if (x == (-1LL << 63)) {
            |                   ~~~~ ^
      src/msgspec/_core.c:10206:16: warning: '_PyLong_Sign' is deprecated [-Wdeprecated-declarations]
       10206 |     bool neg = _PyLong_Sign(obj) < 0;
             |                ^
      /Users/sobolev/.pyenv/versions/3.15-dev/include/python3.15/cpython/longobject.h:38:1: note: '_PyLong_Sign' has been explicitly marked deprecated here
         38 | Py_DEPRECATED(3.14) PyAPI_FUNC(int) _PyLong_Sign(PyObject *v);
            | ^
      /Users/sobolev/.pyenv/versions/3.15-dev/include/python3.15/pyport.h:261:54: note: expanded from macro 'Py_DEPRECATED'
        261 | #define Py_DEPRECATED(VERSION_UNUSED) __attribute__((__deprecated__))
            |                                                      ^
      src/msgspec/_core.c:12037:19: warning: shifting a negative signed value is undefined [-Wshift-negative-value]
       12037 |     if (x < (-1LL << 53)) return false;
             |              ~~~~ ^
      3 warnings and 1 error generated.
      error: command '/usr/bin/clang' failed with exit code 1
```

Env: 

```
» clang --version
Apple clang version 17.0.0 (clang-1700.6.3.2)
Target: arm64-apple-darwin24.6.0
Thread model: posix
InstalledDir: /Library/Developer/CommandLineTools/usr/bin
```

Siyet:
Silences three compiler warnings seen on 3.15-dev (and reproducible on 3.14):

- `_constr_as_i64`: `(-1LL << 63)` is a left shift of a negative value (UB, `-Wshift-negative-value`). Replaced with `INT64_MIN`, which is the value it was computing.
- `double_as_int64`: `(-1LL << 53)` is the same UB. Replaced with `-(1LL << 53)` (negate a positive shift), keeping the exact `-2**53` bound.
- `ms_passes_big_int_constraints`: `_PyLong_Sign` is `Py_DEPRECATED(3.14)`. Use the public `PyLong_GetSign` on 3.14+ and keep `_PyLong_Sign` on older versions.

All three are behavior preserving. Built and ran the unit suite on both branches: 3.12 (`_PyLong_Sign`) and 3.14 (`PyLong_GetSign`), both green; the three warnings are gone under 3.14 headers.

Note: the reporter's paste also mentions "1 error" alongside the warnings. That is an `SSIZE_MAX undeclared` from CPython's 3.15 `pyport.h` when `_core.c` is compiled in isolation without `_GNU_SOURCE`; it doesn't reproduce under the normal build (CPython's `pyconfig.h` sets the feature-test macros), so it's out of scope here.

Closes #1048

### DIFF
--- a/src/msgspec/_core.c
+++ b/src/msgspec/_core.c
@@ -10171,7 +10171,7 @@ static bool
 ms_passes_big_int_constraints(PyObject *obj, TypeNode *type, PathNode *path) {
 #if PY314_PLUS
     /* obj is always a PyLong here, so PyLong_GetSign can't fail */
-    int sign;
+    int sign = 0;
     PyLong_GetSign(obj, &sign);
     bool neg = sign < 0;
 #else
--- a/src/msgspec/_core.c
+++ b/src/msgspec/_core.c
@@ -3675,7 +3675,7 @@ _constr_as_i64(PyObject *obj, int64_t *target, int offset) {
     }
     /* Do offsets for lt/gt */
     if (offset == -1) {
-        if (x == (-1LL << 63)) {
+        if (x == INT64_MIN) {
             PyErr_SetString(PyExc_ValueError, "lt <= -2**63 is not supported");
             return false;
         }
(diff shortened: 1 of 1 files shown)


src/msgspec/_core.c
```diff
@@ -10169,7 +10169,14 @@ ms_passes_int_constraints(uint64_t ux, bool neg, TypeNode *type, PathNode *path)
 /* Constraint checks for a PyLong that is known not to fit into a uint64/int64 */
 static bool
 ms_passes_big_int_constraints(PyObject *obj, TypeNode *type, PathNode *path) {
+#if PY314_PLUS
+    /* obj is always a PyLong here, so PyLong_GetSign can't fail */
+    int sign = 0;
+    PyLong_GetSign(obj, &sign);
```
sobolevn, line 10175: It is better to use explicit `(void)PyLong_GetSign(obj, &sign);` to tell that we ignore the result value.

PR #1079, opened 2026-06-16
try codspeed runners

provinzkraut:


provinzkraut:
Superseded by #1082 (for now).

provinzkraut:
@ofek ran this with the codspeed runners a couple of times now. The results seem to be stable, but the benchmarks run for 7 minutes now, which also isn't great

ofek:
I'm totally fine with that if it means more reliable measurements, especially since our build/test workflow is about the same anyway https://github.com/msgspec/msgspec/actions/runs/27905578518?pr=1079

ofek:
Do you mind explaining what you mean by the results now being seemingly stable? I haven't looked.

provinzkraut:
> Do you mind explaining what you mean by the results now being seemingly stable? I haven't looked.

I just re-ran them a couple of times and checked if the results were about the same, compared to previous runs on the same branch. Probably not a super reliable way to test this, but it's the best I could some up with other than directly merging to `main` and seeing what happens.

> I'm totally fine with that if it means more reliable measurements, especially since our build/test workflow is about the same anyway https://github.com/msgspec/msgspec/actions/runs/27905578518?pr=1079

True. Just a bit longer turnaround time if there's something to address.

What's more important is that only arm runners are available, so our main arch isn't being tested. Probably not super relevant for *most* parts, but especially in the hot paths there are *some* platform differences.

ofek:
I'm fine with merging and seeing what happens as it would be a clear improvement to what we have now!

provinzkraut:
> I'm fine with merging and seeing what happens as it would be a clear improvement to what we have now!

Alright, let's try it

provinzkraut:
@ofek did you change something about the merge queue? This PR somehow merged with its last commit message, instead of the PR title: 551e2dca463dfb8fe4166fc0a5e8682e97329eba

ofek:
No but I found the issue, the docs [say](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/configuring-commit-squashing-for-pull-requests):

> The default message uses the commit title and message if the pull request contains only 1 commit, or the pull request title and list of commits if the pull request contains 2 or more commits.

We were using `Default message` so I just changed it to `Pull request title and description`.

image

PR #1080, opened 2026-06-17
Fix inspect.type_info crashing on mixed-type Literals

### LINKED ISSUE #1018: inspect.type_info crashes on mixed-type Literals

gaoflow:
Fixes #1018.

### Problem

`msgspec.inspect.type_info` crashes on mixed-type `Literal`s:

```python
msgspec.inspect.type_info(Literal[1, None])      # TypeError: '<' not supported between instances of 'NoneType' and 'int'
msgspec.inspect.type_info(Literal[True, "yes"])  # same
```

`type_info` does `tuple(sorted(args))` on the literal's values, and Python 3 forbids `<` between values of different types. The encoder/decoder already handle these mixed-type literals fine (e.g. `Literal[1, None]` decodes `1`/`null` and rejects `2`), so `type_info` failing on them is a contract violation.

### Fix

Sort the literal values with a `(type name, value)` key instead. This keeps existing same-type ordering identical (`Literal[3,1,2]` → `(1,2,3)`, `Literal[True,False]` → `(False,True)`) while grouping mixed-type literals deterministically rather than crashing, and a `try/except TypeError` falls back to the original order as a safety net. The `LiteralType.values` annotation is widened to allow mixed and `None` members.

### Tests

Added `test_mixed_literal` (`Literal[1, None]`, `Literal[True, "yes"]`) and `test_bool_literal`. Without the fix `test_mixed_literal` raises the `TypeError`; with it the full `test_inspect.py`/`test_schema.py` pass (276 passed, 13 skipped).

I'll add a `docs/changelog.md` entry referencing this PR number in a follow-up commit.

Disclosure: I used AI assistance (Claude) to help locate and draft this fix, under my direction and review.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -17,7 +17,7 @@
 - Fix reference leaks in `typenode_collect_literal`, `Meta.__rich_repr__`,
   `ms_decode_bigint`, and `Encoder.__init__` ({pr}`1021`, {pr}`1022`,
   {pr}`1023`, {pr}`1040`).
-- Fix `msgspec.inspect.type_info` crashing on mixed-type `Literal`s such as `Literal[1, None]` ({pr}`1080`).
+- Fix `msgspec.inspect.type_info` and `msgspec.json.schema` crashing on mixed-type `Literal`s such as `Literal[1, None]` ({pr}`1080`).
 - Fix missing GC traversal and clearing of some module state members ({pr}`1060`).
 - Ensure an exception is always set on allocation failures ({pr}`1044`).
 - Fix compilation warnings on Python 3.15 ({pr}`1077`).
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -17,6 +17,7 @@
 - Fix reference leaks in `typenode_collect_literal`, `Meta.__rich_repr__`,
   `ms_decode_bigint`, and `Encoder.__init__` ({pr}`1021`, {pr}`1022`,
   {pr}`1023`, {pr}`1040`).
+- Fix `msgspec.inspect.type_info` crashing on mixed-type `Literal`s such as `Literal[1, None]` ({pr}`1080`).
 - Fix missing GC traversal and clearing of some module state members ({pr}`1060`).
 - Ensure an exception is always set on allocation failures ({pr}`1044`).
 - Fix compilation warnings on Python 3.15 ({pr}`1077`).
--- a/src/msgspec/_json_schema.py
+++ b/src/msgspec/_json_schema.py
@@ -364,7 +364,11 @@ class _SchemaGenerator:
             else:
                 schema["anyOf"] = options
         elif isinstance(t, mi.LiteralType):
-            schema["enum"] = sorted(t.values)
+            # `t.values` may mix types (e.g. `Literal[1, None]`), which a plain
+            # `sorted` can't order; reuse the same type-aware sort as
+            # `inspect.type_info` so the two entry points agree and neither
+            # crashes.
+            schema["enum"] = list(mi._sort_literal_args(t.values))
         elif isinstance(t, mi.EnumType):
             schema.setdefault("title", t.cls.__name__)
             if doc := _get_doc(t):
--- a/src/msgspec/inspect.py
+++ b/src/msgspec/inspect.py
@@ -308,10 +308,11 @@ class LiteralType(Type):
     ----------
     values: tuple
         A tuple of possible values for this literal instance. Only `bool`,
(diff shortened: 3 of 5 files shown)


Siyet (changes requested):
Thanks for the fix! The `type_info` part is solid: I verified the sort key is total and deterministic for every value the validator lets through, same-type ordering is preserved bit-for-bit, and the full unit suite is green on your branch.

One blocker though - the same crash is still one call above, in `json.schema`:

```python
>>> msgspec.json.schema(Literal[1, None])
TypeError: '<' not supported between instances of 'int' and 'NoneType'
```

`_json_schema.py` (~line 367) does its own plain `schema["enum"] = sorted(t.values)` on the same values. So the motivating examples from the PR description now work in `inspect.type_info` but still fail in the more user-facing `json.schema` entry point with the exact same `TypeError` - the crash just moved one frame down. Since `LiteralType.values` is already sorted by your fix, `schema["enum"] = list(t.values)` is enough; please also add a `test_schema.py` case for `Literal[1, None]` (the schema suite is green today only because no test covers mixed literals there).

Non-blocking notes:

- The `except TypeError` fallback looks unreachable: `multi_type_info` validates types by constructing a decoder first, which rejects anything but None/bool/int/str before the sort ever runs, and within those the key can't raise. Fine to keep as a guard, but it could just as well go.
- The `sorted(e.value for e in t.cls)` for enums a few lines below has the same latent issue for mixed-value enums - out of scope here, just noting it as a possible follow-up.
- For bool/int mixes (`Literal[0, True]`) the ordering changes vs current main (`(0, True)` -> `(True, 0)`). That only affects unreleased bool-literal support (#1004), so no released behavior changes, but worth a line in the PR description.

The changelog entry is already in place, thanks for that.

binggao1230:
Good catch — fixed. `json.schema` did its own `sorted(t.values)` on the same values, so `Literal[1, None]` still crashed there. It now reuses the same type-aware sort (`_sort_literal_args`) as `inspect.type_info`, so both entry points agree and neither crashes:

```python
>>> msgspec.json.schema(Literal[1, None])
{'enum': [None, 1]}
```

Added `test_mixed_literal` to `tests/unit/test_schema.py` (`Literal[1, None]` and `Literal[True, "yes"]`) — it fails at the old `sorted()` and passes now. Since a `LiteralType`'s values are already canonically ordered by `type_info`, the reused helper is idempotent for single-type literals, so existing enum ordering is unchanged. Full schema + inspect suites green.

Siyet (approved):
The follow-up commit resolves the change request: `json.schema` now reuses the same type-aware `_sort_literal_args` as `inspect.type_info`, so both entry points handle mixed-type literals consistently instead of one of them still crashing.

Verified against `main`: `inspect.type_info(Literal[1, None])` and `msgspec.json.schema(Literal[1, None])` both crash before this PR and both work after, while same-type literal ordering (`str`/`int`/`bool`) is unchanged. `test_inspect.py` + `test_schema.py` green. Thanks!

PR #1081, opened 2026-06-17
Clean up module import handling in `PyInit__core`

provinzkraut:
Clean up handling of module imports in `PyInit__core`:

- Unify error handling; Introduce and always use `error` label to ensure temporary objects are cleaned up (previously we often did `return NULL` early, missing `Py_XDECREF` in a few places)
- Introduce a new `SET_MODULE_REF` macro to uniformly handle the "call `PyModule_AddObjectRef`, check for errors" pattern
- Introduce a new `IMPORT_TEMP_MODULE` to uniformly handle the "import module as `temp_module`, get attribute from `temp_module` as `temp_obj`, and assign it to the module state"

Removes 2 leaks found with `LSan` on (checked locally, as it's turned off in CI)

<hr>

Based on @sobolevn's suggestion here: https://github.com/msgspec/msgspec/pull/1059#pullrequestreview-4490929893

### DIFF
--- a/src/msgspec/_core.c
+++ b/src/msgspec/_core.c
@@ -22788,7 +22788,9 @@ PyInit__core(void)
     CACHED_STRING(str_is_safe, "is_safe");
 
     /* Initialize the Struct Type */
-    PyState_AddModule(m, &msgspecmodule);
+    if (PyState_AddModule(m, &msgspecmodule) < 0) {
+        goto error;
+    }
     st->StructType = PyObject_CallFunction(
         (PyObject *)&StructMetaType, "s(O){ssss}", "Struct", &StructMixinType,
         "__module__", "msgspec", "__doc__", Struct__doc__
--- a/src/msgspec/_core.c
+++ b/src/msgspec/_core.c
@@ -22525,7 +22525,8 @@ static struct PyModuleDef msgspecmodule = {
 PyMODINIT_FUNC
 PyInit__core(void)
 {
-    PyObject *m, *temp_module, *temp_obj;
+    PyObject *m;
+    PyObject *temp_module = NULL, *temp_obj = NULL;
     MsgspecState *st;
 
     PyDateTime_IMPORT;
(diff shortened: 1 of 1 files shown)


src/msgspec/_core.c
```diff
@@ -22538,148 +22538,139 @@ PyInit__core(void)
 
     StructMetaType.tp_base = &PyType_Type;
     if (PyType_Ready(&NoDefault_Type) < 0)
-        return NULL;
+        goto error;
```
provinzkraut, line 22542: These were not strictly necessary, as there's no references to clean up yet, but I think it's better to have one clearly defined error path for this function

src/msgspec/_core.c
```diff
@@ -22691,100 +22682,72 @@ PyInit__core(void)
...
+    IMPORT_TEMP_MODULE("enum");
     temp_obj = PyObject_GetAttrString(temp_module, "EnumMeta");
-    Py_DECREF(temp_module);
     if (temp_obj == NULL)
-        return NULL;
+        goto error;
     if (!PyType_Check(temp_obj)) {
-        Py_DECREF(temp_obj);
         PyErr_SetString(PyExc_TypeError, "enum.EnumMeta should be a type");
-        return NULL;
```
provinzkraut, line 22723: These are still handled manually, as they work a bit different from the rest. There's only 2 cases of them though, so I felt introducing a macro wasn't worth it

src/msgspec/_core.c
```diff
@@ -22824,9 +22787,19 @@ PyInit__core(void)
...
+return m;
+
+error:
+    Py_XDECREF(temp_module);
+    Py_XDECREF(temp_obj);
+    return NULL;
+
+#undef SET_MODULE_REF
+#undef IMPORT_TEMP_MODULE
+#undef SET_REF
```
provinzkraut, line 22811: Just some housekeeping, also not strictly necessary, but good practice

src/msgspec/_core.c
```diff
@@ -22538,148 +22538,139 @@ PyInit__core(void)
...
 
     temp_obj = PyTuple_Pack(2, st->MsgspecError, PyExc_ValueError);
-    if (temp_obj == NULL) return NULL;
+    if (temp_obj == NULL) goto error;
     st->DecodeError = PyErr_NewExceptionWithDoc(
         "msgspec.DecodeError",
         "An error occurred while decoding an object",
         temp_obj, NULL
     );
     Py_XDECREF(temp_obj);
```
```suggestion
    Py_DECREF(temp_obj);
```

It can't be `NULL` here. But, I think that we should manage it as well, because it is later reuse. Let's create a macro for it as well. So it will be decrefing the last temp object automatically.

src/msgspec/_core.c
```diff
@@ -22538,148 +22538,139 @@ PyInit__core(void)
...
+    SET_MODULE_REF("ValidationError", st->ValidationError);
 
     /* Initialize the struct_lookup_cache */
     st->struct_lookup_cache = PyDict_New();
-    if (PyModule_AddObjectRef(m, "_struct_lookup_cache", st->struct_lookup_cache) < 0)
-        return NULL;
+    SET_MODULE_REF("_struct_lookup_cache", st->struct_lookup_cache);
+
+#define IMPORT_TEMP_MODULE(MOD_NAME) \
+    do { Py_XDECREF(temp_module); \
```
```suggestion
    do { \
        Py_XDECREF(temp_module); \
```

style nit :)

src/msgspec/_core.c
```diff
@@ -22538,148 +22538,139 @@ PyInit__core(void)
...
-        return NULL;
-    if (PyModule_AddObjectRef(m, "ValidationError", st->ValidationError) < 0)
-        return NULL;
+    SET_MODULE_REF("MsgspecError", st->MsgspecError);
+    SET_MODULE_REF("EncodeError", st->EncodeError);
+    SET_MODULE_REF("DecodeError", st->DecodeError);
+    SET_MODULE_REF("ValidationError", st->ValidationError);
 
     /* Initialize the struct_lookup_cache */
     st->struct_lookup_cache = PyDict_New();
```
sobolevn, line 22664:
```suggestion
    st->struct_lookup_cache = PyDict_New();
    if (st->struct_lookup_cache == NULL) goto error;
```

src/msgspec/_core.c
```diff
@@ -22538,148 +22538,139 @@ PyInit__core(void)
...
+    do { Py_XDECREF(temp_module); \
+        temp_module = PyImport_ImportModule((MOD_NAME)); \
+        if (temp_module == NULL) goto error; \
+    } while (0)
 
 #define SET_REF(attr, name) \
     do { \
     st->attr = PyObject_GetAttrString(temp_module, name); \
-    if (st->attr == NULL) return NULL; \
+    if (st->attr == NULL) goto error; \
```
```suggestion
        if (st->attr == NULL) goto error; \
```

style nit :)

src/msgspec/_core.c
```diff
@@ -22824,9 +22787,19 @@ PyInit__core(void)
...
         "__module__", "msgspec", "__doc__", Struct__doc__
     );
-    if (PyModule_AddObjectRef(m, "Struct", st->StructType) < 0) return NULL;
+    SET_MODULE_REF("Struct", st->StructType);
 #ifdef Py_GIL_DISABLED
     PyUnstable_Module_SetGIL(m, Py_MOD_GIL_NOT_USED);
 #endif
-    return m;
+
+return m;
```
```suggestion
Py_DECREF(temp_module);
Py_DECREF(temp_object);
return m;
```

we should not forget to clean up temp objects on success.

src/msgspec/_core.c
```suggestion
    (void)PyState_AddModule(m, &msgspecmodule);
```

src/msgspec/_core.c
provinzkraut: Yup, already done :)

src/msgspec/_core.c
provinzkraut: This should already handled by `SET_MODULE_REF`: If `PyModule_AddObjectRef` is passed `NULL`, it returns `-1`, in which case we `goto error`

src/msgspec/_core.c
provinzkraut: Not sure I understand why. Can you explain?

sobolevn: It is either `(void)` so we "silence" the compiler warning on unused return value (looks like we are not using this flag in msgspec 🤔) or even better we should check for `if (PyState_AddModule(m, &msgspecmodule) < 0) goto error;`, because it can fail.

src/msgspec/_core.c
provinzkraut: > But, I think that we should manage it as well, because it is later reuse. Let's create a macro for it as well.

What exactly do you mean? Like a macro that's something like
```c
#define SET_TEMP_OBJ(TEMP) \
    do { \
        Py_XDECREF(temp_obj); \
        temp_obj = TEMP;\
        if (temp_obj == NULL) \
            goto error; \
    } while (0)
```
?

src/msgspec/_core.c
provinzkraut: Ah! That makes sense. I'll add an explicit check here, since that seems to be the best option

src/msgspec/_core.c
sobolevn: Oh, right :)

src/msgspec/_core.c
provinzkraut: @sobolevn I'm wondering if we should just remove these checks? They could only ever be reached if someone patched these classes in the stdlib for some reason, where it would then fail further down the road if these aren't types. We're also not doing these kinds of assertions anywhere else in this block

provinzkraut: Removing them would make this much cleaner, because then we could use the same pattern as everywhere else

src/msgspec/_core.c
provinzkraut: Done

src/msgspec/_core.c
provinzkraut: Actually, @ofek it seems you added those. Was there a particular reason for this? If yes, we should probably document it, otherwise I'd say we remove them

ofek: I don't recall but if CI passes with their removal then feel free to merge!

provinzkraut: Tests at least pass locally. I'll try removing them

Siyet (commented):
Did a deep pass on bd95764 (code walk of every failure path + an empirical check). Substance looks good, I'd like to get this in.

**Both claimed leaks are real and fixed.** On the base, the `uuid` and `decimal` blocks are the only two that never `Py_DECREF(temp_module)` (the reference is overwritten by the next `PyImport_ImportModule`), and they're exactly what LSan would flag on a plain import. Verified empirically too: with the merge-base and this branch built from the same parent, `sys.getrefcount(sys.modules["uuid"])` and `...["decimal"]` drop from 4 to 3 while every other module stays identical. Beyond those two, the unified error label also closes the ~20 conditional error paths that used to drop a live `temp_module` on early `return NULL` - walked all ten import blocks, the ownership chain converges on every path, no double-DECREF (the `SET_TEMP_OBJ`/`IMPORT_TEMP_MODULE` XDECREF-before-replace handles reentry, and the error label's XDECREFs are NULL-safe for the early `PyType_Ready` gotos). Full unit suite is green on a local build.

**On the open items:**

- The conflict with main is a single hunk from #962: keep `SET_REF(convert_generic_alias, "convert_generic_alias")` in the `_utils` block and drop the now-redundant `Py_DECREF(temp_module)`. Nothing from #1052/#1097/#1109 touches `PyInit__core`, so it's a one-line resolution.
- The `EnumMeta`/`ABCMeta` `PyType_Check` removal you mentioned trying (and ofek pre-approved with "if CI passes... feel free to merge") never got pushed. Either way works for me - push the removal here, or merge as-is and drop them in a follow-up. I'd just rather not let a green refactor sit for another two weeks.

After the rebase this is an approve from me.

PR #1083, opened 2026-06-17
Support the PyEmscripten platform

### LINKED ISSUE #1089: Publish `pyemscripten_*` wheels
### Description

https://blog.pyodide.org/posts/314-release/#pep-783-is-accepted-what-does-it-mean

ofek:
Resolves #1089

### Notes

- The errors that were resolved by the extension changes can be seen in [this run](https://github.com/msgspec/msgspec/actions/runs/27709337400/job/81966407071?pr=1083#step:5:633).
- Some tests for this platform are skipped to avoid crashes such as the one in [this run](https://github.com/msgspec/msgspec/actions/runs/27705230550/job/81952236488?pr=1083#step:5:521). For context, see:
    - https://github.com/pyodide/pyodide/issues/5959
    - https://github.com/pyodide/pyodide/issues/5987
- I also refactored the matrix to allow for defining the target platform rather than just architecture. After this is merged, #987 should be implemented as a new matrix member rather than an entirely separate job.

### References

- https://pyodide.org/en/stable/development/abi.html
- https://cibuildwheel.pypa.io/en/stable/platforms/#pyodide
- https://pyodide.org/en/stable/usage/packages-in-pyodide.html
- https://peps.python.org/pep-0783/

### AI use

I collaborated with GPT-5.5 on Extra High for the following:

- The initial scaffolding for the CI changes (which are mostly no longer present because I disliked the separate job approach).
- The extension changes, primarily for understanding the root cause in order to simplify the logic and create informative code comments.
- A loop to determine if lowering the recursion limit would remove the need to skip some tests but it didn't help for any of them.

### DIFF
--- a/src/msgspec/_core.c
+++ b/src/msgspec/_core.c
@@ -1151,9 +1151,13 @@ static PyObject *
 IntLookup_GetInt64(IntLookup *self, int64_t key) {
     if (MS_LIKELY(self->compact)) {
         IntLookupCompact *lk = (IntLookupCompact *)self;
-        Py_ssize_t index = key - lk->offset;
-        if (index >= 0 && index < Py_SIZE(lk)) {
-            return lk->table[index];
+        /* Compute the dense-table index in 64-bit space so large offsets
+         * cannot wrap through Py_ssize_t on 32-bit builds. */
+        if (key >= lk->offset) {
+            uint64_t index = (uint64_t)key - (uint64_t)(lk->offset);
+            if (index < (uint64_t)Py_SIZE(lk)) {
+                return lk->table[(Py_ssize_t)index];
+            }
         }
         return NULL;
     }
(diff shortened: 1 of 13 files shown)


ofek:
@msgspec/committers This is ready for review! I was anticipating having to exclude these builds on PRs but they are actually now the fastest platform job. Should we wait to build 3.15?

tests/unit/test_raw.py
```diff
@@ -8,6 +8,11 @@
 
 import msgspec
 
+emscripten = pytest.mark.skipif(
```
sobolevn:
```suggestion
requires_subprocess = pytest.mark.skipif(
```

Maybe it is better to check if we have `subprocesses` support instead of hardcoding the exact platform?

ofek: Done, thanks!

sobolevn (approved):
Great stuff! Thanks!

src/msgspec/_core.c
```diff
@@ -1150,9 +1150,13 @@ static PyObject *
 IntLookup_GetInt64(IntLookup *self, int64_t key) {
     if (MS_LIKELY(self->compact)) {
         IntLookupCompact *lk = (IntLookupCompact *)self;
-        Py_ssize_t index = key - lk->offset;
-        if (index >= 0 && index < Py_SIZE(lk)) {
-            return lk->table[index];
+        /* Compute the dense-table index in 64-bit space so large offsets
+         * cannot wrap through Py_ssize_t on 32-bit builds. */
+        if (key >= lk->offset) {
```
provinzkraut, line 1156: Probably a bit paranoid, but this does kinda sit in a hot path, so I'd be curious to see the impact this has on pure int decoding when measured with #1087

ofek: Let's merge that one first and then I'll rebase.

provinzkraut: Looks good!

ofek:
@provinzkraut I rebased to trigger tests on the larger runners. Should we merge?

provinzkraut:
> @provinzkraut I rebased to trigger tests on the larger runners. Should we merge?

I'm fine with it. The perf tests seem to be quite useless at the moment though, even with the macro-runners. I've re-opened https://github.com/CodSpeedHQ/codspeed/issues/409 :shrug:

PR #1087, opened 2026-06-18
Extend CI benchmarking with microbenchmarks for individual types

provinzkraut:
Before, we only tested overall performance with mixed data. However, this can skew results a bit. 

This introduces targeted, isolated tests for builtin types we support, to make it easier to catch regressions / test improvements for individual type handling.

### DIFF
--- a/tests/prof/perf/test_module.py
+++ b/tests/prof/perf/test_module.py
@@ -1,13 +1,27 @@
 import subprocess
 import sys
 
+import pytest
 
-def test_import_time(benchmark):
+
+@pytest.mark.parametrize(
+    "module_name",
+    [
+        "msgspec",
+        "msgspec.structs",
+        "msgspec.json",
+        "msgspec.msgpack",
+        "msgspec.inspect",
+        "msgspec.yaml",
+        "msgspec.toml",
+    ],
+)
+def test_import_time(benchmark, module_name):
     # also measures interpreter startup etc., but is the best we can do now, since we
     # don't have a fully isolated module, and some per-process global state, so
     # re-importing purely within the current interpreter doesn't get reliable timings
     def do_import():
-        subprocess.run([sys.executable, "-c", "import msgspec"], check=True)
+        subprocess.run([sys.executable, "-c", f"import {module_name}"], check=True)
 
     do_import()  # execute once to reduce possibility of a super cold FS cache
 
--- a/tests/prof/perf/test_encode_decode.py
+++ b/tests/prof/perf/test_encode_decode.py
@@ -1,3 +1,10 @@
+import datetime
+import decimal
+import enum
+import random
+import uuid
+from typing import Literal
+
 import pytest
 from pytest_codspeed import BenchmarkFixture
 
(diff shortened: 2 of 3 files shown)


tests/prof/perf/test_module.py
```diff
@@ -1,13 +1,27 @@
...
+        "msgspec.toml",
+    ],
+)
+def test_import_time(benchmark, module_name):
     # also measures interpreter startup etc., but is the best we can do now, since we
     # don't have a fully isolated module, and some per-process global state, so
     # re-importing purely within the current interpreter doesn't get reliable timings
     def do_import():
-        subprocess.run([sys.executable, "-c", "import msgspec"], check=True)
+        subprocess.run([sys.executable, "-c", f"import {module_name}"], check=True)
```
ofek, line 24: Can you pass this to `hyperfine` instead? Otherwise, we really won't get reliable measurements. I'd recommended tracking just the difference/true overhead.

```nu
❯ (
    uv run --python 3.14 --no-project --isolated --with msgspec
    -- hyperfine
        --export-json -
        --style none
        --shell none
        --min-runs 100
        --warmup 10
        --reference-name baseline
        --reference 'python -c pass'
        --command-name combined
        'python -c "import msgspec"'
    err> null-device
    | from json
    | get results
    | select command median
    | update median {|r| $r.median * 1000 }
    | transpose --header-row --as-record
    | insert overhead {|r| $r.combined - $r.baseline }
)
╭──────────┬───────╮
│ baseline │ 47.02 │
│ combined │ 87.67 │
│ overhead │ 40.65 │
╰──────────┴───────╯
```

ofek (commented):
Very nice!

tests/prof/perf/test_module.py
provinzkraut: Not really, no. That will work for local benchmarking, but not in CI with codspeed, where we use its [CPU instrumentation](https://codspeed.io/docs/instruments/cpu/index), not walltime measurements. 

Unfortunately I haven't found a good way yet to measure the import time without the interpreter startup overhead with codspeed.

ofek: Is there no way to submit a raw value that we can track over time?

provinzkraut: Not that I know of, no. I could look into hacking something together, but iiuc that just isn't the intended use case. codspeeds model is tuned towards running stuff you submit there. A bit unfortunate really

provinzkraut: it should be fine as-is though, as we're only tracking the relative performance, not the absolute numbers, and the stability there is exactly what codspeed is attempting to solve with their CPU simulation. i'd suggest we just go with that for a while and see how it works out. with the latest changes i've made, things seem to be quite stable now though

ofek (approved):
Thanks!

Can you please open an issue to track startup overhead with `hyperfine` as conceptualized [here](https://github.com/msgspec/msgspec/pull/1087#discussion_r3440731704) using either Codspeed's [walltime instrument](https://codspeed.io/docs/instruments/walltime/index) or some other approach? We really just need any rudimentary way to track the `overhead` delta from that comment.

provinzkraut:
> Can you please open an issue to track startup overhead with `hyperfine` as conceptualized [here](https://github.com/msgspec/msgspec/pull/1087#discussion_r3440731704) using either Codspeed's [walltime instrument](https://codspeed.io/docs/instruments/walltime/index) or some other approach? We really just need any rudimentary way to track the `overhead` delta from that comment.

@ofek I think we already got #886 tracking this. I'm planning to tackle that soon

ofek:
Oh that's true, thanks! Sorry to keep talking about startup overhead; it's just that I maintain a great number of CLIs and the pain there is real 😅

PR #1092, opened 2026-06-20
Extend struct performance tests

provinzkraut:
Extend the continuous benchmarks for structs, to include more common struct operations, and complex structs.

### DIFF
--- a/tests/prof/perf/test_structs.py
+++ b/tests/prof/perf/test_structs.py
@@ -219,7 +219,15 @@ def test_pickle_load(benchmark: BenchmarkFixture):
 @pytest.fixture(
     params=[
         "msgspec",
-        pytest.param("copy", marks=[pytest.mark.skipif("sys.version_info < (3, 13)")]),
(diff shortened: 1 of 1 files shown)


tests/prof/perf/test_structs.py
```diff
@@ -136,9 +178,107 @@ def test_equality(benchmark):
...
+        warmup_rounds=10,
+    )
+
+    assert res == items
+
+
+@pytest.fixture(
+    params=[
+        "msgspec",
+        pytest.param("copy", marks=[pytest.mark.skipif("sys.version_info < (3, 13)")]),
```
ofek: Can you please add a `reason`?

ofek (changes requested):

tests/prof/perf/test_structs.py
provinzkraut: Done!

ofek (approved):

PR #1101, opened 2026-06-27
Fix docs builds that run on PRs from forks

ofek:
https://github.com/msgspec/msgspec/actions/runs/28223897460/job/83611202184

> Error: Refusing to check out fork pull request code from a 'pull_request_target' workflow. This workflow runs with the base repository's GITHUB_TOKEN, secrets, default-branch cache scope, and runner access. Fetching and executing a fork's code in that trusted context commonly leads to "pwn request" vulnerabilities. To opt in, review the risks at https://gh.io/securely-using-pull_request_target and set 'allow-unsafe-pr-checkout: true' on the actions/checkout step.

Copilot opened https://github.com/msgspec/msgspec/pull/1100 but I very much dislike the `workflow_run` trigger because the status does not get reflected on the PR. Therefore, either failures go unnoticed or there has to be a job that manually updates the PR's commit status (and if something goes wrong then we're back to failures going unnoticed).

### DIFF
--- a/.github/workflows/docs-deploy.yml
+++ b/.github/workflows/docs-deploy.yml
@@ -23,19 +23,18 @@ concurrency:
 jobs:
   docs:
     name: Build and check docs
+    if: github.event_name == 'push'
     uses: ./.github/workflows/docs.yml
     permissions:
       actions: read
       contents: read
     with:
-      cache-namespace: "${{ github.event_name == 'pull_request_target' && format('pr-{0}', github.event.pull_request.number) || 'main' }}"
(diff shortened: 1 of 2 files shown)


sobolevn:
Looks like it did not fix the issue on https://github.com/msgspec/msgspec/pull/1052
Links:
- https://github.com/msgspec/msgspec/actions/runs/28282291556/job/83800187600?pr=1052
- https://github.com/msgspec/msgspec/actions/runs/28282291608/job/83800187786?pr=1052

ofek:
Try again https://github.com/msgspec/msgspec/pull/1102

PR #1103, opened 2026-06-27
Fix GC internals docs link

notpushkin:
The “Garbage collector design” document [has been moved](https://web.archive.org/web/20250903075149/https://devguide.python.org/internals/garbage-collector/index.html) to CPython main repository: https://github.com/python/cpython/blob/main/InternalDocs/garbage_collector.md


provinzkraut (approved):
Thanks!

PR #1105, opened 2026-07-01
Fix `to_builtins` with `frozendict`

### LINKED ISSUE #1104: Incorrect `to_builtins` convertion with `builtin_types=(frozendict,)`
I did not test it enough. It does not convert the internal objects:

```python
def test_frozendict_builtin_types(self):
    msg = to_builtins(frozendict({"test": 1}), builtin_types=(frozendict,))
    assert type(msg) is frozendict
    assert msg == frozendict({"test": 1})

    msg = to_builtins(
        frozendict(
            {
                "test": frozendict({1: "a"}),
                FruitStr.BANANA: 1,
                "b": [FruitInt.APPLE],
            }
        ),
        builtin_types=(frozendict,),
    )
    assert type(msg) is frozendict
    assert type(msg["test"]) is frozendict
    assert type(list(msg.keys())[1]) is str
    assert msg == frozendict({
        "test": frozendict({1: "a"}),
        "banana": 1,
        "b": [-1],
    })
```

this test fails.

I have a fix for this.

sobolevn:
Now it follows the same rules as `dict` :)
1. Converts nested elements
2. Respects `order` keyword

Closes #1104

### DIFF
--- a/src/msgspec/_core.c
+++ b/src/msgspec/_core.c
@@ -20225,6 +20225,17 @@ cleanup:;
     return out;
 }
 
+#if PY315_PLUS
+static PyObject *
+to_builtins_frozendict(ToBuiltinsState *self, PyObject *obj) {
+    PyObject *out = to_builtins_dict(self, obj);
+    if (self->builtin_types & MS_BUILTIN_FROZENDICT) {
+        return _PyFrozenDict_NewSteal(out);
+    }
+    return out;
+}
+#endif
+
 static PyObject *
 to_builtins_struct(ToBuiltinsState *self, PyObject *obj, bool is_key) {
     if (Py_EnterRecursiveCall(" while serializing an object")) return NULL;
@@ -20506,8 +20517,7 @@ to_builtins(ToBuiltinsState *self, PyObject *obj, bool is_key) {
     }
 #if PY315_PLUS
     else if (PyFrozenDict_Check(obj)) {
-        if (self->builtin_types & MS_BUILTIN_FROZENDICT) goto builtin;
-        return to_builtins_dict(self, obj);
+        return to_builtins_frozendict(self, obj);
     }
 #endif
     else if (ms_is_struct_type(type)) {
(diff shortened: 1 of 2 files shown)


Siyet (dismissed):
LGTM, thanks for the quick follow-up! Did a deep pass on 5a54d59 including an empirical run on 3.15.0a7.

Verified:

- The #1104 repro is fixed: enum keys/values and nested frozendicts convert in both modes, `str_keys` and `order` are honored now, plain-dict parity holds for content and ordering in all three `order` modes, and the type is preserved only with `builtin_types=(frozendict,)` - matching the flag's intent from #1052.
- Refcounting is clean on every path: `_PyFrozenDict_NewSteal`'s NULL-guard covers the `to_builtins_dict` error paths (enc_hook raise, unsortable keys with `order`, `str_keys` TypeError), and 100k-300k iteration loops over both happy and error paths show zero RSS/gc-object growth.
- No dict-path regression (byte-identical behavior for plain dict, with and without the flag), encoders/`convert` untouched as they should be, builds and full suite green on 3.12 (`PY315_PLUS` off) and 3.15, including the PR diff re-applied onto current main.

Three small notes, none blocking:

1. In `test_frozendict_builtin_types_sorted` the `assert_eq` helper only checks ordering under `if isinstance(left, typ)` - if the fix ever regressed to returning a plain `dict`, equality would still pass (`frozendict == dict`) and the order check would silently skip, so the test would stay green without testing anything. An `assert isinstance(left, typ)` (or a `type(res) is frozendict` pin in the test) closes that - same lesson as the `test_order_frozendict` fix in #1107.
2. The PR body has no `Closes #1104`, so the issue won't auto-close on merge - worth adding.
3. Optional: one test line for a frozendict *subclass* input (it now collapses to exact `frozendict`/`dict`, consistent with dict-subclass behavior, but nothing pins that).

sobolevn:
I've updated tests and will merge it once CI passes.

sobolevn:
@Siyet thanks a lot for your review! Really helpful.

PR #1107, opened 2026-07-01
Fix `test_order_frozendict`

sobolevn:
Refs https://github.com/msgspec/msgspec/pull/1052#pullrequestreview-4606876383


sobolevn:
СС @vqko

vqko (commented):
Looks correct to me now

PR #1114, opened 2026-07-03
Fix `__struct_encode_fields__` check in CI

sobolevn:
```python
>>> import msgspec
>>> class S(msgspec.Struct):
...     i: int
...     
>>> S(i=1).__struct_encode_fields__
('i',)
```

For some reason CI in https://github.com/msgspec/msgspec/pull/813 passed 🤔

### DIFF
--- a/src/msgspec/__init__.pyi
+++ b/src/msgspec/__init__.pyi
@@ -90,6 +90,7 @@ def field(*, name: str | None = None) -> Any: ...
 class Struct(metaclass=StructMeta):
     __struct_fields__: ClassVar[tuple[str, ...]]
     __struct_config__: ClassVar[structs.StructConfig]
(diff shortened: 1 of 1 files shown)


sobolevn:
Ok, I know what happened 🤦‍♂️ 

- You can click "Merge when ready" while CI is active
- Since only a single job is required: "[CI capability policy / CI capability policy (pull_request_target)](https://github.com/msgspec/msgspec/actions/runs/28671827809/job/85036607982?pr=1114)"
- If it passes, the PR will be merged
- Even if other jobs fail

I propose to use the same thing we use in CPython for this: https://github.com/python/cpython/blob/1b4135a2c69f5320dbdcb37dcc295a5fd36580b5/.github/workflows/build.yml#L676-L727
The action is called https://github.com/marketplace/actions/alls-green

PR #1116, opened 2026-07-04
Run stubtest in the type-checking CI job

### LINKED ISSUE #1056: Consider using `stubtest` to test the stubs
It is a default tool from `mypy` (we already have, since it is a part of `mypy`).
It used to test that runtime and stubs match: https://github.com/typeddjango/django-stubs/blob/master/scripts/stubtest.sh

It can find missing type definition or ones that we forgot to remove for some reason.
Extra / missing func parameters / names / etc.

It is a really useful tool.

Siyet:
Closes #1056.

Adds a `stubtest` step to the `test-typing` job so stub/runtime drift is caught in CI instead of by hand (as in #1043, #1062). `mypy` is already in the `test-typing` group, so no new dependency.

### Root-cause fixes (instead of allowlisting)

- **Positional-only `__text_signature__`.** `json`/`msgpack` `encode`/`decode` and the `Encoder`/`Decoder` methods take their first arg positional-only at runtime, but their C `__text_signature__` omitted the `/`, so `inspect.signature` reported them as positional-or-keyword and stubtest flagged 10 mismatches. Added the `/` in `_core.c` so introspection matches runtime; behaviour is unchanged (keyword calls already raised). This removes all 10 pos-only allowlist entries.
- **`StructMeta` `@disjoint_base`.** `StructMeta` is a disjoint base at runtime; marked it `@disjoint_base` in the stub (the stub already imports from `typing_extensions`, so no new floor of consequence). Removes that allowlist entry.

### Other stub fixes

- `structs.pyi` was missing `__all__`.
- `json.format`s stub marked `buf` positional-only (`/`), but the runtime accepts it as a keyword, so the `/` is dropped to match (this is the one callable that really takes a keyword).

### Allowlist

What is left in `tests/typing/stubtest_allowlist.txt` is 10 structural false positives stubtest cannot see through: the per-class `__init__` synthesized in C for every `Struct`/`inspect` type, the custom `__new__` on `Meta`/`Ext`/`UnsetType`, `Struct.__init_subclass__` (config kwargs via `dataclass_transform`, surfaces on 3.14+), and the private `_utils` module.

Run with `--ignore-unused-allowlist`. The `test-typing` job runs stubtest on a single Python resolved from `python-version-file: pyproject.toml` (currently 3.14, floats up over time), not a pinned matrix, so a single superset allowlist with `--ignore-unused` is robust to the runner Python moving. Verified green on 3.12, 3.14, 3.15.


tests/typing/stubtest_allowlist.txt
```diff
@@ -0,0 +1,46 @@
...
+# Private, intentionally unstubbed helper module.
+msgspec\._utils\..*
+
+# encode/decode take their first argument positional-only in C, but the
+# generated __text_signature__ omits the "/", so stubtest reads them as
+# positional-or-keyword. The runtime rejects keyword calls, so the stubs (which
+# mark "/") are correct. json.format is the one callable that really accepts a
+# keyword, and its stub is fixed instead of allowlisted.
+msgspec\.json\.encode
+msgspec\.json\.decode
```
sobolevn: It is typed as:

```python
def decode(
    buf: Buffer | str,
    /,
    *,
    type: type[_T],
    strict: bool = True,
    dec_hook: _DecHookSig = None,
) -> _T: ...
```

It's real signature:

```python
>>> msgspec.json.decode('{}')
{}
>>> msgspec.json.decode(buf='{}')
  File "<python-input-5>", line 1, in <module>
    msgspec.json.decode(buf='{}')
    ~~~~~~~~~~~~~~~~~~~^^^^^^^^^^
TypeError: Missing 1 required arguments
```

It's `__text_signature__`: 

```python
>>> inspect.signature(msgspec.json.decode)
<Signature (buf, *, type='Any', strict=True, dec_hook=None)>
>>> msgspec.json.decode.__text_signature__
"(buf, *, type='Any', strict=True, dec_hook=None)"
```

The correct fix would be to add `/` to the text signature:

```c
"json_decode(buf, /, *, type='Any', strict=True, dec_hook=None)\n"
```

And remove these entries from the stubtest ignore.

tests/typing/stubtest_allowlist.txt
```diff
@@ -0,0 +1,46 @@
...
+msgspec\.UnsetType\.__new__
+
+# Struct.__init_subclass__ is stubbed with the config kwargs (via
+# dataclass_transform) so type checkers accept `class Foo(Struct, frozen=True)`.
+# At runtime those kwargs are handled by the metaclass, so __init_subclass__
+# itself takes none. Surfaces on Python 3.14+.
+msgspec\.Struct\.__init_subclass__
+
+# StructMeta is a disjoint base at runtime. Marking it @disjoint_base in the
+# stub is deferred (see #1056) pending a typing_extensions compat decision.
```
sobolevn: Sorry, I don't understand :)
What compat decision are you talking about specifically? 
That not all type checkers support `@disjoint_base`?

In this case it would be just a regular `(T) -> T` function.

sobolevn:
> These go in tests/typing/stubtest_allowlist.txt. Run with --ignore-unused-allowlist since the exact finding set shifts across Python versions (verified green on 3.12, 3.14, 3.15).

If this is the case, you can use per-version ignores. See https://github.com/typeddjango/django-stubs/blob/master/scripts/stubtest.sh

tests/typing/stubtest_allowlist.txt
Siyet: Done in d4560b2: added the `/` to the `__text_signature__` for `json`/`msgpack` `encode`/`decode` and the `Encoder`/`Decoder` methods, so `inspect.signature` reports them positional-only and stubtest agrees. Dropped all 10 pos-only entries from the allowlist. Behaviour is unchanged (they already rejected keyword calls), only the introspection metadata was wrong.

tests/typing/stubtest_allowlist.txt
Siyet: Right, my "compat decision" was just the worry that `disjoint_base` is new enough (typing_extensions 4.13) that pulling it into the stub raises the minimum type-checker/typeshed floor: an older mypy would error on the import rather than treat it as identity. But the stub already imports `Buffer`/`Self`/`dataclass_transform` from `typing_extensions`, so the floor is modern anyway and the bump is small. Added `@disjoint_base` to `StructMeta` in d4560b2 and dropped the allowlist entry. mypy and pyrefly are both happy with it.

Siyet:
@sobolevn On per-version allowlists: django-stubs runs stubtest across a pinned matrix of Python versions, so a `common` + per-version split (each checked for unused entries) fits there. msgspec's `test-typing` job runs stubtest on a single Python resolved from `python-version-file: pyproject.toml` (`requires-python >=3.10`), which currently lands on 3.14 and floats up as new stables release. With one floating version a single superset allowlist + `--ignore-unused-allowlist` is simpler and doesn't break when the runner's default Python moves. After the two fixes above the allowlist is down to 10 structural entries, only one of which is version-dependent (`Struct.__init_subclass__`, 3.14+).

If you'd rather keep the unused-entry safety that `--ignore-unused` gives up, the single-file alternative is to pin the stubtest step to a specific Python and drop the flag: strict mode already passes on 3.14 with this allowlist. That pins the interpreter for one step and would trip local runs on other versions, so I left it out, but happy to go that way if you prefer.

Siyet:
Heads up on the red CI here: the failure is in the `test-typing` stubtest step, but it is not from this PR. It is a pre-existing mypy error in `_json_schema.py` introduced by #1028 (`none_member: Type | None` passed to `to_schema`), which nothing caught before because the msgspec source is not type-checked in CI, only `tests/typing` is. This PR 's stubtest step is the first thing to run mypy over the package, so it surfaces it, which is the point.

Fixed separately in #1121. Once that lands I will rebase this and CI goes green.

.justfile
```diff
@@ -90,6 +90,9 @@ test-typing: (
   env-run "test" "pyright tests/typing"
 ) (
   env-run "test" "pyrefly check tests/typing"
+) (
+  env-run "test"
+  "python -m mypy.stubtest msgspec --allowlist tests/typing/stubtest_allowlist.txt --ignore-unused-allowlist"
```
sobolevn: I still advocate for the version-based allowlists, because otherwise we would have `--ignore-unused-allowlist` option enabled, which can hide problems :(

Siyet:
@sobolevn conceded on the per-version allowlists. `--ignore-unused-allowlist` is gone, so an entry that stops matching now fails the run.

Layout: `tests/typing/stubtest_allowlist.txt` holds the entries that are true on every supported version, `stubtest_allowlist_py<major><minor>_plus.txt` holds the version-gated ones, and `scripts/stubtest.py` passes the common file plus every versioned file at or below the running interpreter. I used a "from this version on" rule rather than django-stubs' exact-version files because `test-typing` resolves its Python from `python-version-file`, so the version floats; exact-version files would go dark the moment it moves. Say the word and I will flatten it to exact versions.

Correction to my earlier comment: `Struct.__init_subclass__` is not a 3.14+ finding, stubtest reports it from 3.13 on, hence `py313_plus`.

Verified on 3.10 through 3.14: clean, and a stale entry, a version-gated entry in the common file, or a missing per-version file each fail the run.

One caveat worth stating: stubtest runs in `test-integration`, which has no matrix, so CI only ever exercises one version's branch of this. Happy to move the step into the matrixed `test` job if you want the mechanism to actually bite.

sobolevn (approved):
I am not 100% sure that `tests/typing/stubtest_allowlist_py313_plus.txt` is a great idea. In my opinion, it better to duplicate the code for each version. But, for the start - it is good enough!

PR #1118, opened 2026-07-06
Fix json.encode of a dict keyed by a plain str Enum

gaoflow:
### Problem

Encoding a dict whose keys are members of a plain `enum.Enum` with `str` values
raises, even though the same key type round-trips through every other path:

```python
import msgspec, enum

class Fruit(enum.Enum):
    APPLE = "apple"
    BANANA = "banana"

msgspec.json.decode(b'{"apple": 1}', type=dict[Fruit, int])  # ok
msgspec.msgpack.encode({Fruit.APPLE: 1})                     # ok
msgspec.to_builtins({Fruit.APPLE: 1})                        # ok -> {'apple': 1}
msgspec.json.encode({Fruit.APPLE: 1})
# TypeError: Only dicts with str-like or number-like keys are supported
```

`dict[Fruit, int]` is an accepted *decode* type and both `msgpack.encode` and
`to_builtins` handle the key, so `json.encode` rejecting it is a cross-codec
inconsistency.

### Cause

`json_encode_enum` routes a key's underlying value through
`json_encode_dict_key_noinline`, which only handles non-`str` keys (`str` keys
are fast-pathed in `json_encode_dict_key`). A plain `str`-valued enum therefore
hits the "unsupported key" error.

### Fix

When encoding an enum as a dict key, write a `str` value straight to the key
with `json_encode_str`; non-`str` values keep going through
`json_encode_dict_key_noinline` as before.

### Tests

`test_encode_dict_plain_enum_str_key` covers a plain str `Enum` used as a dict
key (encode output and encode/decode round-trip), and `FruitStr` is added to the
existing mixed-key-type dict test.

`pytest tests/unit/test_json.py` → 1786 passed.

### DIFF
--- a/src/msgspec/_core.c
+++ b/src/msgspec/_core.c
@@ -14153,9 +14153,20 @@ json_encode_enum(EncoderState *self, PyObject *obj, bool is_key)
     PyObject *value = PyObject_GetAttr(obj, self->mod->str__value_);
     if (value == NULL) return -1;
 
-    int status = (
-        is_key ? json_encode_dict_key_noinline(self, value) : json_encode(self, value)
-    );
+    int status;
+    if (is_key) {
+        /* A str value must be written as a JSON object key directly.
+         * json_encode_dict_key_noinline only handles non-str keys (str keys
+         * are fast-pathed in json_encode_dict_key), so recursing into it with
+         * a str value would wrongly raise "unsupported key" - e.g. for a plain
+         * Enum whose members have str values when used as a dict key. */
+        status = PyUnicode_Check(value)
+            ? json_encode_str(self, value)
+            : json_encode_dict_key_noinline(self, value);
+    }
+    else {
+        status = json_encode(self, value);
+    }
 
     Py_DECREF(value);
     return status;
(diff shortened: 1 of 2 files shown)


Siyet (approved):
Verified against `main`: `json.encode` of a dict keyed by a plain str `Enum` (and the same nested in a Struct, or round-tripped) raises on `main` and works here, while every other key path stays byte-identical: plain int `Enum` keys, `IntEnum` keys, `str`-mixin `Enum` keys, plain str/int keys, and enum-as-value are all unchanged. `json.encode` now agrees with `msgpack.encode`, `to_builtins`, and the `dict[FruitStr, int]` decode type, closing the cross-codec inconsistency.

The C fix mirrors `json_encode_dict_key`'s own `PyUnicode_Check(value) ? json_encode_str : json_encode_dict_key_noinline` dispatch, so an enum key with a str value takes exactly the same path as a plain str key. Refcounting on `value` is unchanged (single DECREF on all paths). `tests/unit/test_json.py` passes (1786). Thanks!

PR #1119, opened 2026-07-10
support msgpack encoding with integer keys

jensbjorgensen:
Add `int_keys`: integer MessagePack map keys for Structs

## Motivation

There's an inherent tension between having descriptive field names on a Struct and
the size of its encoded representation — every field name is repeated, in full, in
every message. Since one of msgspec's biggest draws is its encoding efficiency,
being able to encode a Struct's fields with integer keys instead of their names is
an attractive way to shrink MessagePack messages further.

`rename` to single-character names helps, and is the right tool when you need a
readable/valid JSON representation. But MessagePack can do better: it supports
integer map keys directly, and even in JavaScript the common MessagePack decoders
handle integer keys gracefully. This PR makes that available.

## What this adds

A new Struct configuration option, `int_keys`, mapping field names to integers:

```python
class Point(msgspec.Struct, int_keys={"x": 1, "y": 2}):
    x: int
    y: int

>>> msgspec.msgpack.decode(msgspec.msgpack.encode(Point(1, 2)))
{1: 1, 2: 2}
>>> msgspec.msgpack.decode(msgspec.msgpack.encode(Point(1, 2)), type=Point)
Point(x=1, y=2)
```

## Design

- **MessagePack-only.** JSON object keys must be strings, so the JSON encoder
  ignores `int_keys` and keeps using the (possibly renamed) field names — no error,
  and the existing fast JSON path is untouched.
- **A dedicated option, not an extension of `rename`.** `rename` is string-valued
  and applies to both codecs; keeping `int_keys` separate avoids overloading its
  meaning, and the two compose cleanly (a field can be renamed for JSON *and*
  int-keyed for MessagePack).
- **Reuses existing machinery.** Encoding integer keys uses `mpack_encode_long`;
  decoding uses the same `IntLookup` reverse-lookup that already backs integer
  union tags — so this leans on well-trodden code paths rather than introducing new
  ones.
- **Minimal overhead when unused.** The two new `StructMetaObject` members are `NULL`
  for Structs without `int_keys`, and the encode/decode fast paths are unchanged for
  them. On decode, the key type is only inspected for Structs that opt in.
- **Baked onto the type.** Nested Structs are handled automatically — each encodes
  with its own keys, with nothing extra required at encode/decode time. Partial maps
  are allowed (unlisted fields keep their string keys); keys must be unique within a
  Struct and fit in a signed 64-bit integer.
- **Introspectable.** The assigned key is exposed via
  `msgspec.structs.fields(...).int_key`.

## Compatibility, tests, docs

Fully backward compatible — no behavior change for Structs that don't use
`int_keys`. Includes tests (`tests/unit/test_msgpack.py`, `tests/unit/test_struct.py`)
and documentation (`docs/structs.rst`).

Please consider this addition to msgspec — I'm eager to discuss any amendments or
adjustments that would help it fit the project.

### DIFF
--- a/docs/structs.rst
+++ b/docs/structs.rst
@@ -896,25 +896,41 @@ rather than strings.
     >>> msgspec.msgpack.decode(buf, type=Point)
     Point(x=1, y=2)
 
-Integer keys only make sense for MessagePack — JSON object keys must be strings,
-so the **JSON encoder ignores** ``int_keys`` and continues to use the (possibly
-renamed) field names:
+JSON object keys must be strings, so in JSON the integer keys are encoded as their
+decimal strings (``"1"``, ``"2"``) — the same coercion msgspec uses for integer
+dict keys:
 
 .. code-block:: python
 
     >>> msgspec.json.encode(Point(1, 2))
-    b'{"x":1,"y":2}'
+    b'{"1":1,"2":2}'
+
+    >>> msgspec.json.decode(b'{"1":1,"2":2}', type=Point)
+    Point(x=1, y=2)
+
+Decoding also accepts the original field names, so a message that uses names (for
+example one produced with a plain ``rename``, or by a non-msgspec producer) still
+round-trips:
+
+.. code-block:: python
+
+    >>> msgspec.json.decode(b'{"x":1,"y":2}', type=Point)
+    Point(x=1, y=2)
 
 A few things to note:
 
-- Not every field needs a key; fields missing from ``int_keys`` keep their
-  string names on the wire (producing a message with mixed integer/string keys).
+- Not every field needs a key; fields missing from ``int_keys`` keep their string
+  names on the wire (producing a message with mixed integer and string keys).
 - Integer keys are baked onto the type, so **nested** structs are handled
   automatically — each struct encodes with its own ``int_keys``, with no extra
-  configuration needed at encode/decode time.
-- ``int_keys`` composes with ``rename`` (and ``field(name=...)``): the string
-  name is used for JSON, the integer for MessagePack.
+  configuration needed at encode/decode time. This also holds under
+  ``order="sorted"``, which emits the same integer keys.
+- ``int_keys`` composes with ``rename`` (and ``field(name=...)``): ``int_keys``
+  sets the wire key for the fields it lists, while ``rename`` still controls the
+  string key of any unlisted fields (and the name that JSON decoding falls back to).
 - Integer keys must be unique within a struct and fit in a signed 64-bit integer.
+- ``int_keys`` cannot be combined with ``array_like=True`` (array-encoded structs
+  have no field keys); doing so raises a ``ValueError`` at class definition.
 
 The assigned key for each field is available via introspection through the
 ``int_key`` attribute of `msgspec.structs.FieldInfo`:
--- a/docs/structs.rst
+++ b/docs/structs.rst
@@ -869,6 +869,62 @@ precedence.
     b'{"fieldX":1,"y":2}'
 
 
+Integer Field Keys
+------------------
+
+Descriptive field names make for readable code but bloat the encoded message,
+since the name of every field is repeated in full on the wire. For MessagePack
+you can avoid this by encoding fields with **integer keys** instead of their
+names, using the ``int_keys`` configuration option — a mapping from field name to
+integer. This is like renaming (above), but the wire keys are compact integers
+rather than strings.
+
+.. code-block:: python
+
+    >>> import msgspec
+
+    >>> class Point(msgspec.Struct, int_keys={"x": 1, "y": 2}):
+    ...     x: int
+    ...     y: int
+
+    >>> # MessagePack encodes the fields with integer keys ({1: 1, 2: 2})
+    >>> buf = msgspec.msgpack.encode(Point(1, 2))
+    >>> msgspec.msgpack.decode(buf)
+    {1: 1, 2: 2}
+
+    >>> # Decoding maps the integer keys back to fields
+    >>> msgspec.msgpack.decode(buf, type=Point)
+    Point(x=1, y=2)
+
+Integer keys only make sense for MessagePack — JSON object keys must be strings,
+so the **JSON encoder ignores** ``int_keys`` and continues to use the (possibly
+renamed) field names:
+
+.. code-block:: python
+
+    >>> msgspec.json.encode(Point(1, 2))
+    b'{"x":1,"y":2}'
+
+A few things to note:
+
+- Not every field needs a key; fields missing from ``int_keys`` keep their
+  string names on the wire (producing a message with mixed integer/string keys).
+- Integer keys are baked onto the type, so **nested** structs are handled
+  automatically — each struct encodes with its own ``int_keys``, with no extra
+  configuration needed at encode/decode time.
+- ``int_keys`` composes with ``rename`` (and ``field(name=...)``): the string
+  name is used for JSON, the integer for MessagePack.
+- Integer keys must be unique within a struct and fit in a signed 64-bit integer.
+
+The assigned key for each field is available via introspection through the
+``int_key`` attribute of `msgspec.structs.FieldInfo`:
+
+.. code-block:: python
+
+    >>> [(f.name, f.int_key) for f in msgspec.structs.fields(Point)]
+    [('x', 1), ('y', 2)]
+
+
 Encoding/Decoding as Arrays
 ---------------------------
 
--- a/docs/structs.rst
+++ b/docs/structs.rst
@@ -908,14 +908,45 @@ dict keys:
     >>> msgspec.json.decode(b'{"1":1,"2":2}', type=Point)
     Point(x=1, y=2)
 
-Decoding also accepts the original field names, so a message that uses names (for
-example one produced with a plain ``rename``, or by a non-msgspec producer) still
-round-trips:
+Only the exact decimal string the encoder emits is recognized as an integer key
+when decoding JSON: no leading zeros, no ``"-0"``, and a value within the signed
+64-bit range. Anything else (``"01"``, ``"9223372036854775808"``, ...) is treated
+as an ordinary, unknown field name.
+
+Decoding also accepts the field's encoded (string) name as an alias, so a message
+that uses names (for example one produced by a non-msgspec producer) still
+round-trips. The alias is a decoding convenience only; the encoder never emits it.
 
 .. code-block:: python
 
     >>> msgspec.json.decode(b'{"x":1,"y":2}', type=Point)
     Point(x=1, y=2)
+
+Because both spellings are accepted, the decimal string of an integer key must not
+also be the encoded name of a *different* field, or the ``tag_field`` of a tagged
+struct — otherwise two keys would be indistinguishable on the wire. Such
+collisions raise a ``ValueError`` at class definition:
+
+.. code-block:: python
+
+    >>> class Bad(msgspec.Struct, int_keys={"a": 1}, rename={"b": "1"}):
+    ...     a: int
+    ...     b: int = 0
+    Traceback (most recent call last):
+      ...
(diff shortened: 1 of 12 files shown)


Siyet (commented):
Thanks for the thorough PR, and for building the decode path on the existing `IntLookup` machinery from #135, that's the right foundation.

## One correctness blocker

An int-keyed struct that is also tagged decodes fine as a concrete type, but breaks as a **union** member whenever the string tag is not the first key on the wire. This reproduces entirely in-library through the sorted encoder, which orders int keys ahead of the string tag:

```python
class A(msgspec.Struct, tag=True, int_keys={"a": 1, "b": 2}):
    a: int
    b: int

class B(msgspec.Struct, tag=True, int_keys={"a": 1}):
    a: int

buf = msgspec.msgpack.Encoder(order="sorted").encode(A(1, 2))  # {1:1, 2:2, "type":"A"}
msgspec.msgpack.decode(buf, type=A | B)
# ValidationError: Expected `str`, got `int` - at `key` in `$`
```

So msgspec's own sorted encoder emits bytes its own union decoder then rejects. `mpack_decode_struct_union` scans map keys with `mpack_decode_cstr` to find the tag and can't skip past an int key to reach it; string-keyed unions are order-independent precisely because it can skip string keys, so this is a regression relative to them. A foreign producer that doesn't put the tag first hits the same wall. The `test_sorted_order_tagged` test only decodes as a concrete type, so CI stays green over the hole.

Fix needs one of: int-key handling in the union tag-scan, a guarantee the tag encodes first even under `order="sorted"`, or an explicit `tag + int_keys` restriction. Whichever way, this path wants a union test.

(No memory-safety issues found in the C changes on a careful read: refcounting, GC traversal/clear, NULL-checks, the class-creation validation, and the sorted-path recursion/alloc handling are all balanced. The blocker above is a logic/interaction bug, not a leak.)

## Two silent no-op edges worth a decision

- `int_keys` together with `array_like=True` silently drops `int_keys` (array_like wins), with no error at class creation and no test covering the combo.
- `int_keys` is silently ignored on JSON.

Both are documented and consistent (no crash, no round-trip mismatch), but neither gives the user any signal.

## Scope / design, for a maintainer call

The motivation is smaller MessagePack, but `array_like=True` is already strictly smaller (no keys at all) and applies to both codecs, and `rename` to short names covers the readable-and-evolvable case on both codecs. The genuine niche `int_keys` fills is compact **and** schema-evolvable (protobuf-style stable field numbers: order-independent, tolerant of gaps, sparse with `omit_defaults`) which is real but narrow, and the PR doesn't spell it out. Separately, the MessagePack-only-with-silent-JSON-fallback framing cuts against how integer dict keys were handled in #243, where they were made to work in JSON via string coercion rather than being codec-specific.

This is new public API surface (a class kwarg + `FieldInfo.int_key` + a new `__struct_encode_int_keys__` dunder) in the deliberately-small Struct-config area, so it deserves a design sign-off before it moves forward. @jcrist @provinzkraut could one of you weigh in on whether the niche justifies a standalone option, and on the JSON asymmetry? Happy to do a full correctness/test pass once the shape is settled.

jensbjorgensen:
Thanks for the careful review — especially confirming the C changes are memory-clean.

## Correctness blocker (fixed)

Fixed by teaching the union tag-scan to skip non-string keys. `mpack_decode_struct_union`
now peeks each key's type and skips any non-string key (e.g. an integer key from an
`int_keys` struct) while looking for the string tag field, so the tag is found regardless
of its position — the same order-independence string-keyed unions already have. The
concrete-type decode path already handled int keys via `IntLookup`; only the union scan
needed to become key-type-aware. Nothing changes for string-keyed unions.

Added union tests:
- `test_tagged_union_decode_order_independent` — decodes `A | B` for both the default and
  `order="sorted"` encoders (the sorted case is exactly your repro: `{1:1, 2:2, "type":"A"}`).
- `test_tagged_union_decode_tag_last` — a foreign-producer layout with the tag as the last
  key.

## array_like + int_keys (now errors)

`int_keys` together with `array_like=True` now raises `ValueError` at class creation
("array-encoded structs have no field keys") rather than silently dropping `int_keys`.
Test added (`test_int_keys_array_like_rejected`, class + defstruct forms).

## JSON asymmetry (now implemented via string coercion)

Rather than leave the asymmetry, we implemented the #243-style approach, so `int_keys` is
now codec-agnostic:

- Encode: JSON object keys are the integer ids coerced to decimal strings, e.g.
  `{"1": 3, "2": 4}` (msgpack uses native integer keys, `{1: 3, 2: 4}`). This holds under
  `order='sorted'` too — the sorted JSON output emits the int-string keys in the same order
  as the sorted msgpack output, so the two codecs stay consistent.
- Decode: a JSON string key is resolved by parsing it as an integer and looking it up in the
  struct's int-key table; if it isn't a mapped integer, we fall back to matching it as a
  field name, so a name-keyed (e.g. foreign, or upstream-`rename`d) producer still round-trips.

No overhead for structs without `int_keys` (the JSON encode/decode paths only branch when
the type has an int-key table). Tests added for JSON encode/decode, partial maps, tagged
unions in JSON, and the sorted path (asserting it emits int-string keys, tag included).

Happy to change the shape (back to msgpack-only, or an explicit error on JSON) if you'd
rather — but this removes the asymmetry and keeps `int_keys` consistent with how #243 handled
integer dict keys.

## Scope / the niche it fills

Fair point that the motivation undersold this. The niche isn't just "smaller" — it's
**compact *and* schema-evolvable**, i.e. protobuf-style stable field numbers:

- `array_like=True` is smaller still, but positional — you can't reorder or remove a field,
  there are no gaps, and it's not tolerant of producer/consumer schema skew.
- `rename` to short names stays evolvable but keeps string keys (larger, and single-char
  names collide fast).
- `int_keys` is the intersection: **order-independent, gap-tolerant (retired ids are never
  reused), and sparse under `omit_defaults`** — a field can be added/removed/omitted without
  breaking existing readers, while staying as compact as short integer keys allow. That's
  the case neither existing option covers, and it's what makes it worth a stable per-field id
  map rather than positional or string encoding.

Happy to fold this framing into the PR description, and to do the full correctness/test pass
once the shape (standalone option vs. not, and the JSON question) is settled.

jensbjorgensen:
Hello, just checking in, wondering if anyone maybe had a look at the updates?

jensbjorgensen:
Hello friends! Hope you all had a great summer if you're in the northern hemisphere. Any case we might restart consideration of this PR?

Siyet (changes requested):
The previous tag-order blocker is fixed. Three correctness problems remain in the JSON support:

1. In `json_struct_int_key_index`, valid negative int64 keys with 20 characters are excluded, including `-9223372036854775808`. The unchecked signed accumulation also overflows for positive `9223372036854775808`. On my build, decoding that out-of-range positive key results in assignment to the field mapped to `-2**63`, without an error. Please use checked parsing for the complete signed int64 range and test both boundaries and overflow.
2. Integer IDs and string field names can become identical after JSON key conversion. With `int_keys={"a": 1}`, `rename={"b": "1"}`, and fields `a: int; b: int = 0`, the encoded representation of values `(10, 20)` is `{"1":10,"1":20}`, and the decoded values are `(20, 0)`. Roundtrip also fails when `tag_field="1"` collides with integer key `1`. Please validate these collisions, including name aliases accepted during decoding.
3. The original string field names are still used in JSON Schema. For a required field `a` mapped to ID `1`, the encoded JSON is `{"1":10}`, but the generated schema lists `"a"` as required. The encoded output therefore fails validation against the generated schema. Please align the schema with the wire representation and define the treatment of accepted name aliases.

These issues need to be resolved before approval.

jensbjorgensen:
Ok! Thanks again for you patience and consideration of this PR, I think I have met the concerns raised above with my latest push, details as follows:

Code changes:

1. Checked parsing of JSON int keys over the full int64 range

json_struct_int_key_index now accumulates the key in an unsigned 64-bit value, capped at 19 digits so the accumulation itself cannot overflow, and then range-checks against the signed int64 bounds explicitly. The length limit is 20 characters to allow for the sign, so -9223372036854775808 now resolves correctly. Out-of-range keys such as 9223372036854775808 no longer wrap onto the field mapped to -2**63; they fall through to name matching and are treated as unknown fields, so forbid_unknown_fields=True rejects them.

Tests cover both int64 boundaries, one past each boundary, 2**64 - 1, 2**64 (which would wrap a uint64 accumulator without the digit cap), and several over-long keys. Class creation is also tested at both boundaries and one past each.

Design choice worth flagging: decoding now only recognizes the canonical decimal form that the encoder emits. Leading zeros ("01"), "-0", and a leading + are no longer treated as int keys and are handled as ordinary unknown field names. This goes slightly beyond the literal ask, and the reason is item 2 below: the collision check needs a well-defined set of strings that resolve to each int key. With the previous lenient parsing that set was unbounded ("0", "00", "-0", ‚ ... all mapped to key 0), so a rename to "01" would have passed a str(key) comparison yet still been shadowed at decode time. Restricting decode to the canonical form makes the colliding set for key k exactly str(k), which keeps the check exact and the error message obvious. It also matches item 3, since the generated schema only describes the canonical key. The alternative would be to keep lenient parsing and make the collision check compare numerically; happy to switch if you prefer that.

2. Collision validation between int keys, encoded names, and the tag field

At class creation, the decimal string of every int key is compared against the encoded names of all other fields. This uses the encoded names rather than the Python names, since encoded names are what decoding accepts, so it covers rename mappings, rename callables, field(name=...), and names inherited from base classes. After the tag is constructed, the same string is compared against tag_field for tagged structs. Both cases from the review now raise at class definition:

ValueError: `int_keys` value 1 for field 'a' conflicts with field 'b', whose encoded name is also '1'
ValueError: `tag_field='1' conflicts with the `int_keys` value 1 of field 'a'

Design choice: a field whose own encoded name equals its int key string (int_keys={"a": 1}, rename={"a": "1"}) is still allowed, since both spellings resolve to the same field and there is no ambiguity. Also, the check is protocol-agnostic and runs even though msgpack itself cannot collide (integer and string keys are distinct types there). That keeps the rule simple: a struct definition is either valid for all protocols or invalid.

3. JSON Schema describes the wire representation

inspect.Field gained an int_key attribute, populated from the struct's int_keys, and the schema generator uses str(int_key) as the property name and in required for int-keyed fields. Unlisted fields keep their encoded name. Encoded output now validates against the generated schema; the test suite checks this with jsonschema when it is installed (skipped otherwise, since it is not a declared test dependency).

Treatment of name aliases: the schema describes only the encoded form. The field's encoded name, which decoding still accepts as an alias, is intentionally not added to the schema, since the encoder never emits it and expressing "either "1" or "a" is required" would complicate the schema for little benefit. One consequence is documented: a schema generated with forbid_unknown_fields=True (additionalProperties: false) will reject an alias-keyed message that msgspec itself would still decode. If you would rather the schema list aliases as optional properties, or drop alias acceptance entirely, either is a small change.

Docs:

The int_keys docstring previously stated the option had no effect on JSON, which was stale. It now describes the decimal-string form, the uniqueness, range, and collision rules, and the array_like restriction. The "Integer Field Keys" section in docs/structs.rst gained the canonical-form rule, the collision rule with an example, and a note on schema generation and alias handling.

Tests:

New tests in test_json.py (TestStructIntKeys), test_struct.py (collision and boundary cases), test_schema.py, and test_inspect.py. Full unit suite passes locally.

Siyet:
Thx for the updates! The three issues from my previous review are fixed. Two issues remain:

1. `StructMeta.__struct_encode_int_keys__` is missing from the stub. This causes a new stubtest failure and a source mypy error in `inspect.py`. Please add the declaration and typing coverage for the new API.
2. JSON Schema defaults still use name aliases for nested int-keyed Structs. For a frozen `Child` with `x: int` and `int_keys={"x": 1}`, a parent field default of `Child(5)` appears as `{"x": 5}`, while the encoded value and Child schema use `{"1": 5}`. The default therefore does not validate against its schema. Pls preserve the encoded keys in nested defaults and add a regression test

PR #1126, opened 2026-07-19
Fixed an issue decoding `Dataclass` and `TypedDict` instances

mishmish66:
To reproduce try this snippet (in a fresh process):

```python
from dataclasses import dataclass
import msgspec

@dataclass
class Foo:
    π: float
    μ: int

foo = msgspec.json.decode(b'{"\xcf\x80":3.14,"\xce\xbc":0}', type=Foo)
print(foo)
```
Can't decode π because it is unicode and hasn't been loaded into the unicode cache yet

But this:
```python
from dataclasses import dataclass
import msgspec

@dataclass
class Foo:
    π: float
    μ: int

print(msgspec.json.encode(Foo(3.1416, 1)))
foo = msgspec.json.decode(b'{"\xcf\x80":3.14,"\xce\xbc":0}', type=Foo)
print(foo)
```
works fine since the cache is populated during encoding.
This PR should have a fix for this and also a test that fails the old unfixed version.

### DIFF
--- a/src/msgspec/_core.c
+++ b/src/msgspec/_core.c
@@ -8734,6 +8734,9 @@ TypedDictInfo_Convert(PyObject *obj) {
         Py_INCREF(key);
         info->fields[i].key = key;
         info->fields[i].type = type;
+        /* Prime the UTF8 cache so later unicode_str_and_length_nocheck can be
+        * safely used (it assumes the UTF8 is cached) */
+        if (PyUnicode_AsUTF8(key) == NULL) goto cleanup;
         int contains = PySet_Contains(required, key);
         if (contains == -1) goto cleanup;
         if (contains) { type->types |= MS_EXTRA_FLAG; }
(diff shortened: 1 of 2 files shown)


provinzkraut (commented):
Mhm. I'm not sure if this is the best way to fix the issue. Invoking `PyUnicode_AsUTF8` purely for its side effect doesn't seem to great. We're not doing this elsewhere for UTF8 either, so I would suggest not introducing this pattern for this particular path only. 

Have you checked if other types (`Struct`, `NamedTuple`, `attrs` classes) are also affected? If not, dataclasses / TypedDicts should handle this the same way these do.

Siyet:
Pls run each of these tests in a fresh subprocess. If an earlier test encodes a dataclass with `π` and `μ` fields, both new tests pass even without this fix. The UTF-8 representation of those field names is already cached, and `gc.collect()` does not clear it. Running each test in a fresh process would prevent earlier tests from affecting the result.

PR #1127, opened 2026-07-20
Fix dec_hook exception type in json.decode docstring

### LINKED ISSUE #774: What should dec_hook raise if the type is unsupported?
### Question

There seems to be a (deliberate?) inconsistency in the docs regarding what exception `dec_hook` should raise if the expected message `type` is unsupported.

`json.decode` states that `TypeError` should be raised.
All other instances state that `NotImplementedError` should be raised.

Can you clarify which is correct?
- `dec_hook` should raise `TypeError` for unsupported types
- `dec_hook` should raise `NotImplementedError` for unsupported types
- `dec_hook` should raise `NotImplementedError` for unsupported types, unless it is being passed to `msgspec.json.decode` when it should raise `TypeError`.
- other?

rvaccone:
Fixes #774.

`json.decode` was the only `dec_hook` docstring telling users to raise a
`TypeError` for an unsupported type. `json.Decoder`, `msgpack.decode`,
`msgpack.Decoder`, and `convert` all say `NotImplementedError`, which is the
intended contract you confirmed in #774 (`TypeError`/`ValueError` are for a
value that doesn't match the type). This changes the one line to match.

Grepped the rest of the tree for the wording and this docstring was the only
copy, so there's nothing to fix in the docs or stubs. Added an `## Unreleased`
changelog entry. I pointed it at the issue with `{issue}`774`` since there was
no PR number when I wrote it; happy to swap it to `{pr}` with this PR's number.

Disclosure: I used AI assistance (Claude) to draft this change and reviewed the
result.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -27,6 +27,8 @@
   `msgspec.to_builtins` ({pr}`1025`).
 - Document that `omit_defaults` ignores fields with a custom
   `default_factory` ({pr}`1076`).
+- Fix the `msgspec.json.decode` docstring to say `dec_hook` should raise
+  `NotImplementedError` for unsupported types, not `TypeError` ({issue}`774`).
 - msgspec moved to the [msgspec GitHub organization](https://github.com/msgspec/msgspec);
   documentation now lives at [msgspec.dev](https://msgspec.dev) (repository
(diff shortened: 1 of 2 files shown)


provinzkraut (approved):
Thanks!

PR #1128, opened 2026-07-20
fix: raise ValidationError for out-of-range int→float in convert

### LINKED ISSUE #1122: `convert` leaks `SystemError` on an out-of-range `int` → `float`, where `json.decode` cleanly raises `ValidationError`
## Summary

## Reproduction

```python
import msgspec

msgspec.convert(big, float)

# nested, same result:
# SystemError: ...

```

## Expected behavior

## Actual behavior

## Environment

- msgspec 0.21.1
- Python 3.11.10
- macOS 12.7.6 (x86_64)

## Notes

takerukondo:
## Summary

Fixes #1122.

`msgspec.convert(10**400, float)` was leaking `SystemError` because `PyLong_AsDouble` sets `OverflowError` while `convert` still returned a float object. `json.decode` already reports this as `ValidationError: Number out of range`.

Clear the overflow error and raise `ValidationError` (same message as the JSON path). Also reject non-finite results as a belt-and-suspenders check.

## Test plan

- [x] `pytest tests/unit/test_convert.py::TestFloat::test_float_from_out_of_range_int tests/unit/test_convert.py::TestFloat::test_float`
- [ ] CI

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/msgspec/_core.c
+++ b/src/msgspec/_core.c
@@ -20937,7 +20937,18 @@ convert_int(
         return ms_decode_int_enum_or_literal_pyint(obj, type, path);
     }
     else if (type->types & MS_TYPE_FLOAT) {
-        return ms_decode_float(PyLong_AsDouble(obj), type, path);
+        double x = PyLong_AsDouble(obj);
+        /* PyLong_AsDouble sets OverflowError and may return ±inf for ints that
+         * cannot be represented as a finite C double. Surface that as a
+         * ValidationError (matching json.decode) instead of leaking SystemError. */
+        if (x == -1.0 && PyErr_Occurred()) {
+            PyErr_Clear();
+            return ms_error_with_path("Number out of range%U", path);
+        }
+        if (MS_UNLIKELY(!isfinite(x))) {
+            return ms_error_with_path("Number out of range%U", path);
+        }
+        return ms_decode_float(x, type, path);
     }
     else if (
         type->types & MS_TYPE_DECIMAL
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -4,7 +4,7 @@
 
 - Fix `msgspec.convert` raising `SystemError` when converting an out-of-range
   `int` to `float`; it now raises `ValidationError` like `json.decode`
-  ({issue}`1122`).
+  ({issue}`1122`, {pr}`1128`).
 - Add `frozendict` support on Python 3.15+ ({pr}`1052`, {pr}`1105`).
 - Support passing a callable as `decimal_format` to `msgspec.json.Encoder` and
   `msgspec.msgpack.Encoder` for custom `Decimal` encoding ({pr}`978`).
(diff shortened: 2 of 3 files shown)


src/msgspec/_core.c
```diff
@@ -20937,7 +20937,18 @@ convert_int(
         return ms_decode_int_enum_or_literal_pyint(obj, type, path);
     }
     else if (type->types & MS_TYPE_FLOAT) {
-        return ms_decode_float(PyLong_AsDouble(obj), type, path);
+        double x = PyLong_AsDouble(obj);
+        /* PyLong_AsDouble sets OverflowError and may return ±inf for ints that
+         * cannot be represented as a finite C double. Surface that as a
```
provinzkraut, line 20942: Where are you getting the 

> return ±inf for ints that cannot be represented as a finite C double

part from? It's at least not stated in the documentation: https://docs.python.org/3/c-api/long.html#c.PyLong_AsDouble

Siyet:
Thanks for this, and sorry for the long silence. We went with #1162, so I am closing this one. You were here first, so you are owed the actual reasons rather than a bare "duplicate".

Two of them are on us, not on you:

- Nobody approved the workflow runs on this PR, so in 49 days it never got a real CI run beyond the capability policy check. Your tests have never actually been executed by our CI. That is a maintainer miss, and it is why the other PR looks green and this one does not.
- @provinzkraut's question from 2026-07-20 was left hanging and you were never pinged again.

Answering that question now, since it is the substantive one: he was right, and the code comment was wrong. `PyLong_AsDouble` cannot return `+/-inf`. In `Objects/longobject.c` it either takes the compact single digit fast path, which is always finite, or calls `_PyLong_Frexp`, which normalises the significand into `[0.5, 1.0)` and increments the exponent when rounding would push it to `1.0`. `PyLong_AsDouble` then rejects `exponent > DBL_MAX_EXP` with `OverflowError` and returns exactly `-1.0`; otherwise it returns `ldexp(x, exponent)`, at most `(1 - 2**-53) * 2**1024`, which is `DBL_MAX`. So the `isfinite` guard is unreachable and the `-1.0` plus `PyErr_Occurred()` test is sufficient on its own. I checked this rather than taking it on faith: instrumenting that branch on your patch and running 22100 values around the `DBL_MAX` boundary, it never fired.

The `PyErr_Clear()` is a no-op too, though a harmless one. `ms_error_with_path` reports through `PyErr_Format`, and `_PyErr_SetObject` clears the pending exception itself before building the new one, while implicit chaining reads the handled exception stack rather than a merely pending error. The three existing sites in `_core.c` that do this same thing (the integer struct tag check and the `Ext.code` check among them) also set the new error without clearing first.

On the changelog entry, to be clear: writing one was not wrong, several merged PRs here carry their own. Only the placement broke, and recently. Since you opened this, `## Unreleased` became `## Version 0.22.0 (2026-08-11)`, so your hunk still merges cleanly but drops the entry inside an already released section.

Your diagnosis was right and your test case was a good one. Thank you for the time, and please do send more.

PR #1130, opened 2026-07-22
Fix NameError resolving generic PEP 695 TypedDicts under `from __future__ import annotations`

michaelbilow:
Besides this short paragraph, the rest of this PR is entirely AI-generated. I saw no policy on the repo and I hope this isn't annoying—it fixes an actual problem for me (and I'm a big msgspec fan). I've checked the code by hand and tried to mirror the style of the rest of the repo. If what's here is annoying, feel free to reject the PR and please accept my personal apologies. Thanks.

### Problem (Why)

Decoding into a generic `TypedDict` defined with PEP 695 syntax raises `NameError: name 'T' is not defined` on Python 3.12 and 3.13 when postponed annotations are enabled:

```python
from typing import TypedDict
import msgspec

class Ex[T](TypedDict):
    x: T

msgspec.json.decode(b'{"x": 1}', type=Ex[int])
```

`TypedDict` eagerly creates module-bound `ForwardRef` objects. During evaluation, these references may replace the class namespace with the owning module's globals, losing the class's PEP 695 type parameters. A same-named module global could also shadow the class parameter and silently weaken validation.

Python 3.12.0–3.12.3 add another compatibility constraint: the fourth argument to `typing._eval_type` is `recursive_guard`. The `type_params` argument was added in Python 3.12.4 and became mandatory in Python 3.13.

### Solution (What + How)

The class's `__type_params__` are now threaded through annotation evaluation. On Python 3.12.4+, they are passed using the `type_params` keyword. Earlier versions retain the three-argument `_eval_type` call.

Type parameters are also copied into the local evaluation namespace so module-bound forward references cannot discard them or resolve a same-named legacy module `TypeVar` instead.

Tests cover direct annotation inspection, end-to-end JSON and MessagePack decoding, generic inheritance, the Python 3.12.0–3.12.3 signature, and module-level name collisions.

## Test Plan

- Full unit suite on Python 3.12.4: 6,387 passed, 143 skipped.
- Full unit suite on Python 3.14: 6,387 passed, 143 skipped.
- Focused regression tests on Python 3.12.0, 3.12.4, 3.12.12, 3.13, 3.14, and 3.15.
- Ruff lint and formatting checks pass.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -10,6 +10,9 @@
 - Support the PyEmscripten (Pyodide) platform ({pr}`1083`).
 - Fix handling of PEP 695 type parameter syntax (`class Foo[T]`) and of
   `types.GenericAlias` instances in type annotations ({pr}`962`).
+- Fix a `NameError` when resolving a generic `TypedDict` defined with PEP 695
+  type parameter syntax (`class Foo[T](TypedDict)`) under
+  `from __future__ import annotations` on Python 3.12/3.13.
 - Fix a crash on incorrect `typing.ClassVar` annotations ({pr}`1097`).
 - Fix an `AttributeError` when converting to a `Struct` type defined in a
   namespace without a `__name__` ({pr}`1072`).
--- a/src/msgspec/_utils.py
+++ b/src/msgspec/_utils.py
@@ -23,8 +23,6 @@ def get_type_hints(obj):
     return _get_type_hints(obj, include_extras=True)
 
 
-PY_312PLUS = sys.version_info >= (3, 12)
-
 # The `is_class` argument was new in 3.11, but was backported to 3.9 and 3.10.
 # It's _likely_ to be available for 3.9/3.10, but may not be. Easiest way to
 # check is to try it and see. This check can be removed when we drop support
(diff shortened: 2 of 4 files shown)


Siyet (changes requested):
On Python 3.12.3, this breaks decoding an ordinary TypedDict with postponed annotations. This example works on main but raises `TypeError: unsupported operand type(s) for |: 'tuple' and 'set'` on this PR:

```python
from typing import TypedDict
import msgspec

class Example(TypedDict):
    x: int

print(msgspec.json.decode(b'{"x": 1}', type=Example))
```

The `type_params` argument was added to `typing._eval_type` in Python 3.12.4. In earlier 3.12 releases, the fourth positional argument is `recursive_guard`. Please preserve compatibility with that signature and add a regression test

michaelbilow:
Thanks. This is fixed for Python 3.12.0...3.12.3 and a regression test has been added.

Siyet:
Thanks, this addresses both points. I checked the new version on a real 3.12.3: the `TypeError` is gone for postponed TypedDict, Struct and dataclass annotations, and `test_postponed_annotations` fails on the previous version there while passing from 3.12.4 on, so it guards exactly that boundary.

One thing worth extending before merge is the changelog entry. It describes only the `NameError`, but the second commit fixes a separate and quieter problem: when a module-level `TypeVar` shares its name with a PEP 695 parameter, the parameter was silently captured, so `Ex[int]` did not substitute and `decode(b'{"x": 1}', type=Ex[str])` was accepted instead of raising. That one also affects 3.14, where NameError never occurred, so a reader of the changelog would not learn that their 3.14 decoding may have been affected.

michaelbilow:
@Siyet good point, done.

Siyet:
The early 3.12 compatibility problem is fixed. The `>= (3, 12, 4)` gate lines up with where the fourth parameter of `typing._eval_type` changes from `recursive_guard` to `type_params`, and the postponed-annotation cases that broke on 3.12.0 through 3.12.3 all pass again.

One correction to the changelog entry, and it is mine rather than yours: I pointed you at 3.14, so that is what the entry say, but the silent case is not limited to it. On a build of main with string annotations, a module-level `TypeVar` sharing its name with the class type parameter leaves the field resolved as `any`, and data that should be rejected is decoded without error, on 3.12, 3.13, 3.14 and 3.15. Since 3.11 cannot parse the syntax at all, "Python 3.12+" covers it without naming versions that will age. The `NameError` half is accurate as written, including its restriction to 3.12 and 3.13.

PR #1133, opened 2026-07-30
Fix overriding an inherited field alias back to the field's own name

### LINKED ISSUE #654: Field Alias Overrides in Subclasses Not Reflected in __struct_encode_fields__
### Description

When subclassing a msgspec.Struct and redefine a field with a new alias using msgspec.field(name="new_name"), the new alias does not appear to update the __struct_encode_fields__ attribute of the subclass. Instead, the subclass retains the parent's field alias in __struct_encode_fields__. This behavior seems inconsistent with the expectation that subclass field redefinitions should override parent definitions, including aliases used for encoding.

```
from msgspec import Struct, field

class Parent(Struct):
    field1: str = field(name="_field1")

assert Parent.__struct_encode_fields__ == ("_field1",)  # This passes

class NewParent(Parent):
    field1: str = field(name="field1")

assert NewParent.__struct_encode_fields__ == ("field1",)  # This fails with AssertionError
```

**Expected Behavior:**

The __struct_encode_fields__ attribute of NewParent should reflect the new alias "field1" instead of inheriting "_field1" from Parent, considering the explicit redefinition of field1 in NewParent with a new alias.

**Actual Behavior:**

The assertion that NewParent.__struct_encode_fields__ == ("field1",) fails, indicating that __struct_encode_fields__ still contains the parent's alias ("_field1",).

rvaccone:
Fixes #654.

`structmeta_process_rename` skips writing to `renamed_fields` when the requested encode name equals the field's own name, leaving any alias inherited from a base class in place. So `field(name="x")` on a redeclared field `x` was silently ignored, and the same held for a `rename=` policy whose result is the field's own name (e.g. `rename="lower"` under a parent's `rename="upper"`):

```python
class Base(Struct):
    x: int = field(name="_x")

class Sub(Base):
    x: int = field(name="x")

Sub.__struct_encode_fields__  # ('_x',), expected ('x',)
```

The fix clears the inherited entry at both sites instead of skipping the write. Redeclared fields already re-apply the effective naming config, including overriding a base's `field(name=...)` alias with a policy result; this extends that to identity results.

Two behaviors deliberately unchanged: a redeclaration expressing no naming intent (`x: int = 1`, no policy in effect) keeps the inherited alias, and so does `rename=None`, which acts as "no policy in this class" rather than an identity policy. Clearing per field is spelled `field(name="x")`. Both new tests fail before the fix.

### DIFF
--- a/src/msgspec/_core.c
+++ b/src/msgspec/_core.c
@@ -5908,7 +5908,12 @@ structmeta_process_rename(
         ((Field *)default_value)->name != NULL
     ) {
         Field *field = (Field *)default_value;
-        if (PyUnicode_Compare(name, field->name) == 0) return 0;
+        if (PyUnicode_Compare(name, field->name) == 0) {
+            if (PyDict_GetItem(info->renamed_fields, name) != NULL) {
+                return PyDict_DelItem(info->renamed_fields, name);
+            }
+            return 0;
+        }
         return PyDict_SetItem(info->renamed_fields, name, field->name);
     }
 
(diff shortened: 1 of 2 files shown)


Siyet:
Thanks for the fix

Issue #1134, opened 2026-08-07
New release: 0.22.0


sobolevn:
Hi 👋 

@Siyet @provinzkraut @ofek Can we please make a new release? I need some of the changes in `django-modern-rest`, which depends on `msgspec` :) 

I've never released a new version of `msgspec` yet, but I can see that trusted publishing is setup here: https://github.com/msgspec/msgspec/blob/593ec549a99cf3928f1763f92798fe444d3e5c03/.github/workflows/ci.yml#L192-L224

So, if this is fine, I can make one myself :)

sobolevn:
Ok, since I did not recieve any "no", I will take an action and release a new version. I really need some of the bugfixes in my project :) 

Wish me luck!

sobolevn:
Ok, it did not work :(

```
Building cp310-musllinux_riscv64 wheel
CPython 3.10 musllinux riscv64
Setting up build environment...
      + mkdir -p /
      + /opt/python/cp39-cp39/bin/python -c 'import sys, json, os; json.dump(os.environ.copy(), sys.stdout)'
      + which python
      + python -V -V
  Python 3.10.20 (main, Jun  4 2026, 13:38:41) [Clang 22.1.6 ]
      + which uv
Error: cibuildwheel: Command ['which', 'uv'] failed with code 1. 
                                                              ✕ 1.59s

```

Fails consitently: https://github.com/msgspec/msgspec/actions/runs/31476259980/job/93742042390

sobolevn:
I propose skipping `riscv` on `musllinux` for now. The use-case seems VERY rare.

gounthar:
Sorry, my matrix entry is what broke your release.

The manylinux riscv64 wheels all built and tested fine in that run. Only musllinux fails, and I think it's a stale image pin rather than anything riscv-specific.

pypa/manylinux used to skip installing uv in the musllinux riscv64 image, with the comment "uv doesn't provide musl riscv64 wheels due to Rust issues" in `finalize.sh`. That was removed in pypa/manylinux#1969 on 11 July, and their image tests exercise uv there now. cibuildwheel 4.1.0 pins the `2026.06.04-1` image, which predates the fix, so `build[uv]` has no uv to find. 4.2.0 pins `2026.08.04-1`, which does.

[...]

PR #1146, opened 2026-08-11
Bump cibuildwheel to 4.2.0 to fix musllinux riscv64 wheels

gounthar:
The 0.22.0 release run died on `cp310-musllinux_riscv64` with `Command ['which', 'uv'] failed with code 1`, from my riscv64 matrix entry. All six manylinux riscv64 wheels in that same job built, repaired and tested fine, so it is specific to the musllinux image.

pypa/manylinux used to skip installing uv in the musllinux riscv64 image. At the commit the pinned image was built from (`8dbfa8e`, 4 June), `docker/build_scripts/finalize.sh` had:

```
musllinux*_riscv64-uv) continue;;  # uv doesn't provide musl riscv64 wheels due to Rust issues
```

That was removed by pypa/manylinux#1969 on 11 July, and their image test suite now runs `uv venv` / `uv pip install` inside musllinux riscv64. The comment was already out of date when it was written: uv has published a `manylinux_2_31_riscv64.musllinux_1_1_riscv64` wheel since 0.11.18 (1 June).

The pins differ across cibuildwheel releases:

| cibuildwheel | `musllinux_1_2_riscv64` image | uv present |
|---|---|---|
| 4.1.0 | `2026.06.04-1` | no |
| 4.2.0 | `2026.08.04-1` | yes |

So `build-frontend = "build[uv]"` has nothing to find on 4.1.0. This bumps both `uvx --from` pins to 4.2.0.

4.2.0 builds CPython 3.15 by default, but `.cibuildwheel.toml` sets an explicit `build = "cp310-* cp311-* cp312-* cp313-* cp314-* cp314t-*"` list, so no new target appears. The rest of that release is Pyodide 314.0.4, a GraalPy pip hold, and download retries.

I have not run this. The reasoning is from the failing job log and the manylinux source, not from a green build. CI here will not test it either unless someone adds the `build-riscv64` label, since the riscv64 job is excluded on unlabelled PRs, and even then `CIBW_SKIP` drops musllinux outside a release build. So the honest state is that only a release run proves it. If you would rather not move cibuildwheel right before a release, `musllinux-riscv64-image` in `.cibuildwheel.toml` pins just that one image and leaves everything else alone.

Two things worth knowing either way:

The riscv64 job is gated behind the `build-riscv64` label on PRs, and `CIBW_SKIP` drops musllinux unless `build-all` is set. So nothing had ever exercised musllinux riscv64 until the release turned `build-all` on. Whatever you decide here, that combination will keep hiding riscv64 musllinux breakage until release day.

Dependabot covers `github-actions` and `uv`, but the cibuildwheel version is an inline `uvx --from` argument rather than a tracked dependency, which is why it sat at 4.1.0. Moving it into the uv ecosystem somehow would stop the next one, though I do not have a clean suggestion for how.

Refs #1134, #987

### DIFF
--- a/.github/workflows/build-wheels.yml
+++ b/.github/workflows/build-wheels.yml
@@ -91,7 +91,7 @@ jobs:
     - name: Build & Test Wheels
       if: runner.os != 'Windows'
       run: >-
-        uvx --from cibuildwheel==4.1.0
+        uvx --from cibuildwheel==4.2.0
         cibuildwheel
         --config-file .cibuildwheel.toml
         --output-dir wheelhouse
@@ -100,7 +100,7 @@ jobs:
     - name: Build & Test Wheels
       if: runner.os == 'Windows'
       run: >-
-        uvx --from cibuildwheel==4.1.0
+        uvx --from cibuildwheel==4.2.0
         cibuildwheel
         --config-file .cibuildwheel.toml
         --output-dir wheelhouse


sobolevn (approved):
Thank you! It works now 🎉

PR #1148, opened 2026-08-11
bump ruff

chirizxc:


.ruff.toml
```diff
@@ -1,4 +1,5 @@
 line-length = 88
+output-prefer-rule-codes = true
```
chirizxc, line 2: See: https://docs.astral.sh/ruff/settings/#output-prefer-rule-codes | https://github.com/astral-sh/ruff/issues/26320

sobolevn (approved):
Thanks!

Issue #788, opened 2024-12-28
Add support for prebuild armv7l


return42:
msgspec is resource-efficient and therefore predestined for "small setups" (architectures) / it would be nice if we had the pre-builds for such architectures ..

https://github.com/jcrist/msgspec/blob/dd965dce22e5278d4935bea923441ecde31b5325/.github/workflows/ci.yml#L86

About `armv7l` 

- https://cibuildwheel.pypa.io/en/stable/changelog/#v2212
- https://cibuildwheel.pypa.io/en/stable/options/#archs

PR #1149, opened 2026-08-11
[WIP] try to support more wheels

chirizxc:
The diff is large because of:

```console
actions-up --yes --min-age 0
```

I'll revert these changes later

### DIFF
--- a/.github/workflows/build-wheels.yml
+++ b/.github/workflows/build-wheels.yml
@@ -67,7 +67,6 @@ jobs:
           os: ubuntu-latest
           archs: wasm32
           cibw_platform: pyodide
-
         exclude:
         - wheel:
             target: Linux (riscv64)
--- a/.github/workflows/build-wheels.yml
+++ b/.github/workflows/build-wheels.yml
@@ -59,39 +59,39 @@ jobs:
       PACKAGE_DIR: dist/${{ inputs.sdist-name }}
 
     steps:
-      - name: Checkout code
-        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0  # v7.0.0
-        with:
-          persist-credentials: false
+    - name: Checkout code
+      uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0  # v7.0.0
+      with:
+        persist-credentials: false
 
-      - name: Download source distribution
-        uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c  # v8.0.1
-        with:
-          name: artifact-sdist
-          path: dist
+    - name: Download source distribution
+      uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c  # v8.0.1
+      with:
+        name: artifact-sdist
+        path: dist
 
-      - name: Install uv
-        uses: astral-sh/setup-uv@fac544c07dec837d0ccb6301d7b5580bf5edae39  # v8.2.0
-        with:
-          enable-cache: false
+    - name: Install uv
+      uses: astral-sh/setup-uv@fac544c07dec837d0ccb6301d7b5580bf5edae39  # v8.2.0
+      with:
+        enable-cache: false
 
-      - name: Set up QEMU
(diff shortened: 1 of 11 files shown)


.github/workflows/build-wheels.yml
```diff
@@ -22,93 +22,75 @@ jobs:
...
-    - name: Build & Test Wheels
-      if: runner.os != 'Windows'
-      run: >-
-        uvx --from cibuildwheel==4.2.0
-        cibuildwheel
-        --config-file .cibuildwheel.toml
-        --output-dir wheelhouse
-        "${PACKAGE_DIR}"
+      - name: Set up QEMU
+        if: runner.os == 'Linux' && contains(fromJSON('["ppc64le","s390x","armv7l"]'), matrix.wheel.archs)
```
sobolevn: Do we really need these platforms? There was 1 real user request for these platforms: https://github.com/msgspec/msgspec/issues/788

I am asking, because this is not free for us:
- We have to maintain all the build infra
- We have to run CI for this regularly

I think that `msgspec` can be compiled locally quite easily and quite efficently. So, proving wheels is not 100% required for it to work.

Moreover, there's not much platform specific code in `msgspec` for us to test it heavily.

.github/workflows/build-wheels.yml
```diff
@@ -22,93 +22,75 @@ jobs:
...
-      CIBW_SKIP: "${{ !inputs.build-all && '*-musllinux_*' || '' }}"
       ARTIFACT_NAME: artifact-wheels-${{ matrix.wheel.cibw_platform || matrix.wheel.os }}-${{ matrix.wheel.archs }}
       PACKAGE_DIR: dist/${{ inputs.sdist-name }}
 
     steps:
-    - name: Checkout code
-      uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0  # v7.0.0
-      with:
-        persist-credentials: false
+      - name: Checkout code
```
sobolevn: Can you please revert all whitespace and style changes? They create useless code churn.

tests/unit/test_msgpack.py
```diff
@@ -1416,6 +1421,7 @@ def test_timestamp32_lower(self):
         msg = b"\xd6\xff\x00\x00\x00\x00"
         self.check(dt, msg)
 
+    @skip_if_is_32bit
```
sobolevn, line 1424: If we are to support 32-bit archs, we need special tests for them.

.github/workflows/build-wheels.yml
```diff
@@ -22,93 +22,75 @@ jobs:
...
-          os: windows-11-arm
-          archs: ARM64
-        - target: Windows (AMD64)
-          os: windows-latest
-          archs: AMD64
-        - target: Pyodide (wasm32)
-          os: ubuntu-latest
-          archs: wasm32
-          cibw_platform: pyodide
+          - { target: Linux (aarch64),    os: ubuntu-24.04-arm,   archs: aarch64 }
```
sobolevn: Can you please revert this change as well? I mean style only.

sobolevn (commented):
I think that x86_64 macos would be the only reasonable thing to support?

I am also not a fan of the CI speed. All jobs take 10+ minutes to test these platforms. Which is a long time :(

.github/workflows/build-wheels.yml
chirizxc: Yes, this will increase the CI time, but for people who use these targets, installing the wheel will be much faster than compiling it on the spot

chirizxc: I guess we can compile manylinux and musslinux on different GitHub Runners

provinzkraut: Agree with @sobolevn. 

I'd also be cautious to add more platforms on a hunch. If there's a real demand for it, by which I mean at least a ticket asking for a specific platform with a good reason, I'm all for it. But otherwise, maybe let's hold off with it a bit.

chirizxc: In theory, without `cibuildwheel`, we could skip using `QEMU`, use `zig cc` instead, and compile the wheels on an x86 platform, but we would still need to test them, which takes about ~100+ seconds per wheel.

If we remove the musllinux wheels for `s390x`, `ppc64le`, and `armv7l` and keep what's in this PR, the build time will remain roughly the same - less than ~10 minutes per GitHub Runner.

PR #1151, opened 2026-08-11
more classifiers for PyPI metadata

chirizxc:


Siyet (approved):
Thx for taking care of this! Small packaging improvements like this are appreciated.

Issue #958, opened 2025-11-26
C-API Capsule and Cython Bindings


Vizonex:
### Description

I think this project is super cool and could benefit users who want to build off msgspec but at a lower level in C or in Cython. The Encoder/Decoder Classes would be a very cool thing to have in Cython and C for anyone who wants to implement more obscure protocols but at lower levels.

jacopoabramo:
I've been thinking a lot about this, and as an off-topic it would be amazing to be able to integrate `msgspec` in custom written python packages which can be fully compiled with `mypyc`. Although they're not quite there yet, there have been mentions in the following issues: [#1137](https://github.com/mypyc/mypyc/issues/1137), [#1098](https://github.com/mypyc/mypyc/issues/1098).

A possible approach to this would be to move most of the content of `_core.c` into an header file, so that C and Cython extensions can import it, and then provide some utility methods to import the include files for compilation, similar to how `numpy` does. Then the regular `msgspec` package can simply export the header as a module.

I'm just wondering how much of a breaking change would that be.

Vizonex:
> I've been thinking a lot about this, and as an off-topic it would be amazing to be able to integrate `msgspec` in custom written python packages which can be fully compiled with `mypyc`. Although they're not quite there yet, there have been mentions in the following issues: [#1137](https://github.com/mypyc/mypyc/issues/1137), [#1098](https://github.com/mypyc/mypyc/issues/1098).
> 
> A possible approach to this would be to move most of the content of `_core.c` into an header file, so that C and Cython extensions can import it, and then provide some utility methods to import the include files for compilation, similar to how `numpy` does. Then the regular `msgspec` package can simply export the header as a module.
> 
> I'm just wondering how much of a breaking change would that be.

@jacopoabramo 
I actually have a fork that I am attempting to do if this idea was accepted. I actually have a bit of experience with brining in C-API Capsules in cython/cpython and python since me and avestlov were coming up with one for [multidict](https://github.com/aio-libs/multidict/pull/1190) so that the http-parser and writer would run smoother in Cython for aiohttp. 
I'll probably try using a similar setup used with that project in this one so that things can remain organized and so that the `_core.c` file doesn't get too overwhelmed with C-API Capsule bindings. I'll probably have it marked simillar to how `_core.c` currently has things marked.


I might be able to start a pull request soon for this depening on my IRL work schedule. But if I feel fully motivated to develop msgspec a c-api the quickest I can get to doing it is tomorrow or sometime next week. 

[...]

PR #961, opened 2025-11-28
Add C-API Capsule

Vizonex:
> [!WARNING]
> This pull request is a still a being worked on and is based off previous work I did for a different library based off #958

I have not added a whole lot yet in order to be sure that this is what other maintainers had in mind otherwise I will simply wait for input. It should be noted that this concept I had was based off unfinished work I had previously done for another library but works nevertheless. 
> [!IMPORTANT]
> It should also be noted that getting this implemented __will take time and planning__ on what the best approach would be to take and to do it all slow and steady. I have not implemented more objects besides Fields and Factories because I wanted to ensure that what gets added makes sense and is in the best interest of the maintainers in order to not make the process anymore stressful than it may already be to maintain.

Edit: Here is the objects I am planning to add to the C-API

# Checklist
- [x] Field
- [x] Factory
- [x] Ext types
- [ ] Raw Types
- [ ] Attempting to Subclass `StructMetaType` Safely in C (Crashes for me upon attempting to bind at `PyType_Type.tp_new(...)`)
- [x] StructMetaTypes (I'm planning to subclass SturctMeta to implement an SQLModel inspired class and Instrument attributes via subclassing Field Types (I'll make a PR for subclassing fields again since I closed the previous one))
 - [x]  Struct Types
 - [ ]  Encoder/Decoder Types (Tools for extra customization would be nice)
- [ ] 1:1 Msgspec C Structures but renamed as Msgspec... Example: `Field is named MsgspecFieldObject` in `msgspec.h` so when someone compiles the code something Generic like `Field` (which could theoretically be used in a different C library or another Dev's implementation) it isn't being mixed and causing developers headaches and compilation issues.
- [ ] Cython Bindings

fixes #958

PR #1153, opened 2026-08-12
Add narrow Struct builder C API

goblinmode2700:
## What this draft tests

This is a deliberately narrow alternative/complement to #961: one opaque,
versioned C-API capability for constructing an already-validated `msgspec.Struct`
from declared-order owned references. It does not expose msgspec's private
layouts, offsets, metaclass construction, encoders, or decoders.

The concrete consumer is a native TOON decoder. Its parser already knows the
target schema and has final Python field values, but today it must call the
public Struct constructor, which repeats argument binding. The proposed call
lets msgspec retain ownership of allocation, defaults/factories, GC decisions,
abstract checks, and `__post_init__` while removing that duplicate binding work.

## Safety and fallback

- The returned builder is an opaque Python object, not a raw private-layout
  handle.
- It participates in cyclic GC. The test suite covers the class -> token ->
  class cycle found during external review.
- The build call has an explicit consume-on-success-and-failure ownership
  contract.
- Unsupported class/metaclass/interpreter configurations return a normal
  fallback status without an exception.
- V1 advertises no capability on free-threaded builds.
- The public header and table are size/capability/version checked.

## Evidence

- CPython 3.13 upstream unit suite: 6,387 passed, 143 skipped.
- CPython 3.14 free-threaded upstream unit suite: 6,369 passed, 155 skipped.
- Focused API/lifecycle tests: 7 passed.
- TOON 4.1.1 corpus through the consumer: 538/538; strict errors 84/84.
- Matched current-main A/B, same codec wheel and same msgspec base commit:
  S1 about 3% faster (resolved only after doubled sampling), S6 about 3% faster
  but unresolved on confirmation, and S12 about 5% faster and resolved.

This is a draft to discuss whether this small producer-owned capability is a
maintainable first C-API boundary. Production `msgspec-toon` remains pinned to
stock `msgspec==0.21.1`; no patched dependency is shipped.

Thank you to @Vizonex and the participants in #958/#961 for establishing the
capsule/test-module prior art and surfacing the broader extension use cases.

### DIFF
--- a/src/msgspec/msgspec.h
+++ b/src/msgspec/msgspec.h
@@ -1,0 +1,83 @@
+#ifndef MSGSPEC_CAPI_H
+#define MSGSPEC_CAPI_H
+
+#include <Python.h>
+#include <stddef.h>
+#include <stdint.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/*
+ * msgspec native C API, ABI major 1.
+ *
+ * This header exposes no msgspec object layouts. The capsule table is an
+ * opaque, versioned capability boundary implemented by msgspec.
+ */
+#define MSGSPEC_CAPI_CAPSULE_NAME "msgspec._core._C_API_v1"
+
+#define MSGSPEC_CAPI_ABI_MAKE(major, minor) \
+    ((((uint32_t)(major)) << 16) | ((uint32_t)(minor) & 0xffffu))
+#define MSGSPEC_CAPI_ABI_VERSION MSGSPEC_CAPI_ABI_MAKE(1, 0)
+#define MSGSPEC_CAPI_ABI_MAJOR(v) ((uint32_t)(v) >> 16)
+#define MSGSPEC_CAPI_ABI_MINOR(v) ((uint32_t)(v) & 0xffffu)
+
+#define MSGSPEC_CAPI_OK          0
+#define MSGSPEC_CAPI_UNSUPPORTED 1
+#define MSGSPEC_CAPI_ERROR      -1
+
+#define MSGSPEC_CAPI_CAP_STRUCT_BUILD_OWNED_V1 (1ull << 0)
+
+/*
+ * struct_builder_prepare:
+ * - OK returns a new opaque Python token in builder_out.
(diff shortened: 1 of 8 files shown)


ofek (changes requested):
Please provide a concrete example of code that illustrates how you intend for this to be used.

src/msgspec/msgspec.h
```diff
@@ -0,0 +1,83 @@
...
+extern "C" {
+#endif
+
+/*
+ * msgspec native C API, ABI major 1.
+ *
+ * This header exposes no msgspec object layouts. The capsule table is an
+ * opaque, versioned capability boundary implemented by msgspec.
+ */
+#define MSGSPEC_CAPI_CAPSULE_NAME "msgspec._core._C_API_v1"
```
Vizonex, line 18: I really do not believe a `_v1` is needed in this naming and it confuses me quite a bit. You have the right approach was we can slowly add more things in when this first PR has been merged but a recommendation I have is to remove this `_v1` it isn't needed here.

src/msgspec/msgspec.h
```diff
@@ -0,0 +1,83 @@
...
+ *
+ * struct_builder_build_owned:
+ * - values contains exactly nslots declared-field positions.
+ * - each non-NULL entry is one owned reference transferred by the call.
+ * - NULL means absent and carries no ownership.
+ * - for a valid builder, all non-NULL entries are consumed on success or
+ *   failure. The caller must treat every entry as moved after the call.
+ * - success returns a new reference; failure returns NULL with an exception.
+ */
+typedef struct Msgspec_CAPI_v1 {
```
Vizonex, line 46: More or a less a ditto to what I explained a bit earlier.

Vizonex (changes requested):
@goblinmode2700 while I love your approach to this just know that I stopped working on #958 due to my IRL job taking up more chunks of my time away and wanting to explore and invent new things. I am however willing to take the time to review and help you this this PR. 
Be sure to comment that it closes #958 Just know that I am not inline with the whole `_v1` thing and believe this suffix needs dropping. As long as you provide some code examples you should be all set to go.

PR #1155, opened 2026-08-14
Support Optional[Raw] when decoding JSON and MessagePack

### LINKED ISSUE #659: Optional[Raw] results in an unexpected validation error at json.decode time
### Description

it seems that the json parser finds a "object" type in the JSON stream but can't assign it to a fields that expects "null" (air-gapped system, i can't copy paste).

```
class Wrapper(Struct):
  result: Optional[Raw]
```
.. and decode some JSON into it, like `{ "result" : { "foo": "bar" } }` - and that will repro the error

DSeaStar:
Fixes #659.

`msgspec.Raw` was collected with a type mask of 0, so `Optional[Raw]` / `Raw | None` never recorded `MS_TYPE_NONE`. JSON and MessagePack then treated `null` as a `Raw` payload (`b"null"`) instead of `None`, which is the reproduction in #659.

This follows the design from the issue: `Raw` is allowed in a union only with `None`. `null` decodes as `None`; any other value decodes as `Raw`. Unions such as `Raw | int` now raise `TypeError`, matching the existing custom-type restriction.

Changes:
- Give `Raw` its own `MS_TYPE_RAW` flag and reject unsupported `Raw` unions at type-collection time.
- Peek for JSON `null` / MessagePack nil when decoding `Raw | None`.
- Accept `Raw | None` in `convert`.
- Add JSON, MessagePack, and `convert` tests for the `null` and non-null cases.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+- Support `Optional[msgspec.Raw]` / `Raw | None` when decoding JSON and MessagePack ({issue}`659`).
+
 ## Version 0.22.0 (2026-08-11)
 
 - Add `frozendict` support on Python 3.15+ ({pr}`1052`, {pr}`1105`).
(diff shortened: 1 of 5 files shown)


Siyet (changes requested):
Thx for working on this! Pls rebase on main and preserve the `raw` diagnostic from #1169, including both `raw` and `null` for `Raw | None`. Also document null handling and supported unions, and add a nullable Raw example to the typing checks.

PR #1157, opened 2026-08-14
Fix serializing `list` subclasses with external storage

### LINKED ISSUE #1156: `list` subclasses that store data elsewhere silently encode as `[]`
### Description

When a type subclasses both `list` and a container that keeps its items somewhere other than the built-in list storage (e.g. `collections.UserList`), `msgspec` encodes it as an empty array. No error is raised and `enc_hook` is never called, so the data is silently lost.

`json.dumps` from the standard library encodes the same object correctly.

### Reproduction

```python
import collections
import json
import msgspec

class L(collections.UserList, list):
    pass

x = L(["a", "b"])

print(list.__len__(x), len(x), list(iter(x)))
# 0 2 ['a', 'b']

print(json.dumps(x))
# ["a", "b"]

print(msgspec.json.encode(x))
# b'[]'

print(msgspec.msgpack.encode(x))
# b'\x90'

def hook(obj):
    print("hook called")   # never printed
    return list(obj)

print(msgspec.json.encode(x, enc_hook=hook))
# b'[]'
```

### Cause

The encoder dispatches on `PyList_Check` and then reads the built-in list storage directly via `PyList_GET_ITEM`/`PyList_GET_SIZE`. For these types that storage is empty — the items live in `UserList.data` and are only reachable through `__len__`/`__getitem__`/`__iter__`.

### Why it matters

I made [django-msgspec](https://django-msgspec.readthedocs.io/en/latest/) to provide Django integrations to msgspec. In rolling it out for a client, I found code that dumps Django’s [`ErrorList`](https://docs.djangoproject.com/en/stable/ref/forms/api/#django.forms.ErrorList) class into a JSON response. Since this class inherits from both `UserList` and `list`, the migration from stdlib json to msgspec causes error messages to be silently dropped.

Inheritance chain:

```python
>>> from django.forms.utils import ErrorList
>>> [c.__name__ for c in ErrorList.__mro__]
['ErrorList', 'UserList', 'MutableSequence', 'Sequence', 'Reversible',
 'Collection', 'Sized', 'Iterable', 'Container', 'list',
 'RenderableErrorMixin', 'RenderableMixin', 'object']
```

A plain `UserList` (without `list` in the bases) raises `TypeError: Encoding objects of type L is unsupported`, which is fine. It's the combination that turns a loud failure into silent data loss.

The equivalent `UserDict` + `dict` case also produces `{}`, but there `json.dumps` agrees, so it's less clear-cut.

### Expected behaviour

Encode the actual contents, or raise `TypeError`. Anything but silently producing `[]`.

Possible approaches, in rough order of preference:

1. Use the generic sequence path (`__len__`/`__getitem__` or iteration) when `type(obj) is not list` and the type overrides those slots.
2. Treat such types as unsupported so `enc_hook` gets a chance to handle them.

Option 2 alone would be enough for downstream libraries to work around this, since currently `enc_hook` is unreachable for these types.

### Versions

- msgspec 0.21.1
- Python 3.14.6
- macOS (arm64)

adamchainz:
Fixes #1156.

The new `ms_list_is_plain` function checks whether a `list` subclass stores its items in the internal list storage, and if not, copies the items out through the sequence protocol before serializing. This ensures that `list` subclasses are not serialized to an empty array.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## Unreleased
+
+- Fix `list` subclasses that store their items outside of the internal list
+  storage (such as classes inheriting from both `list` and
+  `collections.UserList`) silently encoding as an empty array. They are now
+  encoded via the sequence protocol ({issue}`1156`).
+
 ## Version 0.22.0 (2026-08-11)
 
 - Add `frozendict` support on Python 3.15+ ({pr}`1052`, {pr}`1105`).
--- a/src/msgspec/_core.c
+++ b/src/msgspec/_core.c
@@ -3137,6 +3137,21 @@ ms_is_struct_inst(PyObject *o) {
     return ms_is_struct_meta(Py_TYPE((PyObject *)Py_TYPE(o)));
 }
 
+static MS_INLINE bool
+ms_list_is_plain(PyObject *obj) {
+    /* Checks whether a `list` instance stores its items in the internal list
+     * storage. Some subclasses of `list`, such as those combining it with
+     * `collections.UserList`, may store items elsewhere, only reachable
+     * through the sequence protocol. */
+    PyTypeObject *type = Py_TYPE(obj);
+    if (MS_LIKELY(type == &PyList_Type)) return true;
+    return (
+        type->tp_iter == PyList_Type.tp_iter
+        && type->tp_as_sequence->sq_length == PyList_Type.tp_as_sequence->sq_length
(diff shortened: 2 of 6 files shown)


provinzkraut (changes requested):
I don't think special casing lists here is the solution, and the current behaviour is documented.

I left a more detailed response here: https://github.com/msgspec/msgspec/issues/1156#issuecomment-5301543305

PR #1162, opened 2026-08-20
Fix SystemError when convert()'ing an out-of-range int to float

### LINKED ISSUE #1122: `convert` leaks `SystemError` on an out-of-range `int` → `float`, where `json.decode` cleanly raises `ValidationError`

STiFLeR7:
## Summary

`msgspec.convert(obj, float)` converts a Python `int` to a C `double` via `PyLong_AsDouble` without checking for overflow. For an `int` too large to represent as a finite `float` (e.g. `10**400`), `PyLong_AsDouble` sets `OverflowError` internally but still returns `-1.0`, and that value was passed straight through to `ms_decode_float` and returned to the interpreter with the exception still set — which surfaces as:

```
SystemError: <built-in function convert> returned a result with an exception set
```

This bypasses the documented `ValidationError` contract entirely, so a caller wrapping the call in `except msgspec.ValidationError` doesn't catch it.

`json.decode` already handles the equivalent case correctly, raising `ValidationError: Number out of range`. This PR applies the same "check `PyErr_Occurred()` after a lossy C conversion, then report via `ms_error_with_path`" pattern already used elsewhere in this file (e.g. `_constr_as_f64`), so `convert()` reports the overflow the same way — for both bare (`convert(big, float)`) and nested (`convert({"x": big}, dict[str, float])`) targets, under both `strict=True` and `strict=False`.

Fixes #1122.

## Testing

- Reverted the fix locally and reproduced the exact reported `SystemError` before reapplying it.
- Added `test_float_from_int_out_of_range` to `TestFloat` in `tests/unit/test_convert.py`, covering bare and nested targets under both strict modes.
- Ran the full `tests/unit/` suite: 6052 passed, 473 skipped, 0 failures.

### DIFF
--- a/src/msgspec/_core.c
+++ b/src/msgspec/_core.c
@@ -20937,7 +20937,14 @@ convert_int(
         return ms_decode_int_enum_or_literal_pyint(obj, type, path);
     }
     else if (type->types & MS_TYPE_FLOAT) {
-        return ms_decode_float(PyLong_AsDouble(obj), type, path);
+        double val = PyLong_AsDouble(obj);
+        if (val == -1.0 && PyErr_Occurred()) {
+            /* `obj` is out of range for a C double (PyLong_AsDouble sets
+             * OverflowError but still returns -1.0); without this check
+             * that error leaks past this function as a SystemError. */
+            return ms_error_with_path("Number out of range%U", path);
+        }
+        return ms_decode_float(val, type, path);
     }
     else if (
         type->types & MS_TYPE_DECIMAL
--- a/tests/unit/test_convert.py
+++ b/tests/unit/test_convert.py
@@ -429,6 +429,17 @@ class TestFloat:
         with pytest.raises(ValidationError, match="Expected `float`, got `null`"):
             convert(None, float)
 
+    @pytest.mark.parametrize("strict", [True, False])
+    def test_float_from_int_out_of_range(self, strict):
+        # A python int too large to represent as a C double must raise a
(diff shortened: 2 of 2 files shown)


Siyet (approved):
Verified locally with current `main` (7c2f473) merged into this branch, Python 3.14.6, `-X dev`:

- `convert(10**400, float)`, the nested `dict[str, float]`, `tuple[float, ...]` and Struct-field cases, `strict=False`, and the negative value all raise `ValidationError: Number out of range` with the right path suffix (`$[...]`, `$[1]`, `$.x`).
- `2**1024` is rejected, `2**1023` and `int(sys.float_info.max)` still convert.
- The `OverflowError` that `PyLong_AsDouble` sets does not leak: `__context__` and `__cause__` are `None`, `sys.exc_info()` is clean afterwards.
- Full unit suite: 6387 passed, 143 skipped.

The check mirrors what `_constr_as_f64` already does a few thousand lines up, and the message matches the JSON path. Bug fix with no API change, so I will put it in the merge queue once the CI runs (just approved for the first-time-contributor gate) come back green. The changelog entry gets added in the release-prep batch, same as the other fixes.

PR #1164, opened 2026-08-26
Fix defstruct with bases using a StructMeta subclass

### LINKED ISSUE #974: msgspec.defstruct is not compatible with msgspec.StructMeta
### Description

Structs that were defined with custom metaclass (subclass of `msgspec.StructMeta`) can't be subclassed with `msgspec.defstruct`. (msgspec 0.20.0)

Following code:
```python
import msgspec

class _FooStructMeta(msgspec.StructMeta):
    def __new__(mcls, name, bases, namespace, **struct_config):
        return super().__new__(mcls, name, bases, namespace, **struct_config)

class FooStruct(msgspec.Struct, metaclass=_FooStructMeta):
    pass

BarStruct = msgspec.defstruct(name="BarStruct", fields=[("id", int)], bases=(FooStruct,))
```
Raises this error:
```
  File "/home/jirka642/.config/JetBrains/PyCharm2025.3/scratches/scratch_1.py", line 13, in <module>
    BarStruct = msgspec.defstruct(name="BarStruct", fields=[("id", int)], bases=(FooStruct,))
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jirka642/.config/JetBrains/PyCharm2025.3/scratches/scratch_1.py", line 6, in __new__
    return super().__new__(mcls, name, bases, namespace, **struct_config)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: Struct types cannot define __slots__
```

I have been able to hotfix it by running `namespace.pop("__slots__", None)` in the metaclass, but I'm not sure if that is safe.

abhijeet117:
Fixes #974

## Summary

`defstruct(bases=...)` raised `TypeError: Struct types cannot define __slots__` when a base used a custom `StructMeta` subclass, while equivalent class syntax worked. `defstruct` always built the class through `StructMeta`, so type creation re-dispatched through the base's more derived metaclass and re-entered construction with `__slots__` already injected into the namespace. We now resolve the most derived metaclass among the bases (mirroring CPython's calculation) and construct through it, matching class definition behavior.

## Testing

Added `test_defstruct_custom_metaclass_base` (constructs, instantiates, JSON round-trips) plus `test_defstruct_default_metaclass`. Full unit suite: 6380 passed, 145 skipped (CPython 3.13, Windows). `ruff check`, `ruff format --diff`, and `codespell` clean on changed files.

## Checklist

- [x] Bug reproduced before fix
- [x] Root cause identified
- [x] Bug fixed
- [x] Tests passed

### DIFF
--- a/src/msgspec/_core.c
+++ b/src/msgspec/_core.c
@@ -6795,6 +6795,33 @@ StructMeta_new(PyTypeObject *type, PyObject *args, PyObject *kwargs)
     );
 }
 
+/* Determine the most derived metaclass among `metatype` and the metaclasses
+ * of `bases`. This mirrors the metaclass calculation CPython performs when
+ * creating a class through `type`, so that classes created by `defstruct`
+ * resolve the same metaclass as equivalent class definitions. Returns NULL
+ * with an exception set if the metaclasses conflict. */
+static PyTypeObject *
+structmeta_calculate_metaclass(PyTypeObject *metatype, PyObject *bases) {
+    PyTypeObject *winner = metatype;
+    for (Py_ssize_t i = 0; i < PyTuple_GET_SIZE(bases); i++) {
+        PyObject *base = PyTuple_GET_ITEM(bases, i);
+        if (!PyType_Check(base)) continue;
+        PyTypeObject *base_meta = Py_TYPE(base);
+        if (PyType_IsSubtype(winner, base_meta)) continue;
+        if (PyType_IsSubtype(base_meta, winner)) {
+            winner = base_meta;
+            continue;
+        }
+        PyErr_SetString(
+            PyExc_TypeError,
+            "metaclass conflict: the metaclass of a derived class must be a "
+            "(non-strict) subclass of the metaclasses of all its bases"
+        );
+        return NULL;
+    }
+    return winner;
+}
+
 
 PyDoc_STRVAR(msgspec_defstruct__doc__,
 "defstruct(name, fields, *, bases=None, module=None, namespace=None, "
@@ -6956,8 +6983,15 @@ msgspec_defstruct(PyObject *self, PyObject *args, PyObject *kwargs)
     }
     if (PyDict_SetItemString(namespace, "__annotations__", annotations) < 0) goto cleanup;
 
+    /* Create the class through the most derived metaclass among the bases.
+     * Otherwise type creation re-dispatches through that metaclass, running
+     * this logic a second time on a namespace we've already injected
+     * __slots__ into, which would fail the namespace check. */
+    PyTypeObject *metatype = structmeta_calculate_metaclass(&StructMetaType, bases);
+    if (metatype == NULL) goto cleanup;
+
     out = StructMeta_new_inner(
-        &StructMetaType, name, bases, namespace,
+        metatype, name, bases, namespace,
         arg_tag_field, arg_tag, arg_rename,
         arg_omit_defaults, arg_forbid_unknown_fields,
         arg_frozen, arg_eq, arg_order, arg_kw_only,
--- a/src/msgspec/_core.c
+++ b/src/msgspec/_core.c
@@ -6641,8 +6641,15 @@ StructMeta_new_inner(
     PyObject *args = Py_BuildValue("(OOO)", name, bases, info.namespace);
     if (args == NULL) goto cleanup;
     cls = (StructMetaObject *) PyType_Type.tp_new(type, args, NULL);
-    Py_CLEAR(args);
     if (cls == NULL) goto cleanup;
+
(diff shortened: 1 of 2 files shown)


fallenmi (changes requested):
The selected metaclass is assigned to the generated type, but its construction hooks are never invoked. `StructMeta_new_inner` calls `PyType_Type.tp_new(metatype, ...)` directly, bypassing the selected metaclass's `__new__` and `__init__`. That breaks the documented `StructMeta` subclass pattern for supplying default struct configuration.

At exact head `b11a57d52c3329fa460a1a1d60e7059106796119`, using the documented pattern with a metaclass that does `config.setdefault("kw_only", True)` produces these divergent results:

```text
defstruct: calls=[]; signature=(child: int, *, base: int); Dynamic(2, base=1) succeeds
class syntax: calls=[("new", "Static"), ("init", "Static")]; signature=(*, base: int, child: int); Static(2, base=1) raises TypeError
```

The added custom-metaclass test only delegates from `__new__` and checks the resulting metaclass, so it cannot detect that the hook was skipped. Please preserve the selected metaclass's hooks exactly once (including their configuration/validation behavior) and add a regression covering a hook with an observable effect.

Verification: exact base `f51f378335b01dc0026dc6553a0b9e1915a8edae` reproduces the reported `Struct types cannot define __slots__` failure; the two added tests pass on the head; the complete unit suite passes (`6382 passed, 143 skipped`). The hook/configuration probe above remains red on the head.

Reviewed with OpenAI Codex assistance

abhijeet117:
@fallenmi Thanks for the feedback. I've pushed the requested changes, addressed the review comments, and added the relevant regression coverage. The updated implementation and tests have been verified. Please take another look when you get a chance.

fallenmi (changes requested):
The follow-up still does not preserve the selected metaclass's construction semantics, and the direct `tp_init` call introduces a regression for ordinary class syntax. At exact head `995ca0b6ffc0ae4cdfeff1de0babce5b3347bc72`, `PyType_Type.tp_new(type, ...)` still bypasses `CustomMeta.__new__`; `type->tp_init` is then invoked inside `StructMeta_new_inner`, before the struct fields/configuration are installed.

For `defstruct` I observe only `[("init", "Dynamic")]`, with `__struct_fields__` still missing during that hook, signature `(child: int, *, base: int)`, and a positional child accepted. Equivalent class syntax observes `[("new", "Static"), ("init", "Static"), ("init", "Static")]`, signature `(*, base: int, child: int)`, and rejects the positional child. A metaclass whose `__new__` raises for the generated name is likewise bypassed by `defstruct` but honored by class syntax.

The added regression test is also red as written because it retains the hooks recorded while creating `Base`; the complete local unit run is `1 failed, 6382 passed, 143 skipped` on CPython 3.13.5. Clearing those setup calls would still expose the missing `__new__` and unapplied `kw_only` policy. The public Build and Test, docs-pr, and CodSpeed workflows for this head are currently `action_required` with zero jobs.

There is also a deterministic failure-path reference leak in the new code. `Py_CLEAR(args)` was moved below both the `tp_new` and `tp_init` failure branches, while `cleanup:` never decrefs `args`. Repeating 25 custom-`__init__` validation failures increases the refcount of a namespace sentinel by 25 on this head; exact base increases it by 0.

Please dispatch construction through the selected metaclass on the original namespace/configuration (or otherwise preserve the same lifecycle), so `__new__` and `__init__` each run exactly once at the class-syntax-equivalent point, validation/config hooks are honored, and all failure-owned references are released. The compatible-MRO and metaclass-conflict behavior did match class syntax in my probes.

Reviewed with OpenAI Codex assistance.

Siyet (commented):
The previous class-construction and error-handling issues are fixed. One difference from ordinary class creation remains: configuration flags are converted to booleans twice in `defstruct`, first during argument parsing and then in `StructMeta`. With a stateful `__bool__`, the resulting configuration can differ:

```python
class Flag:
    def __init__(self):
        self.calls = 0
    def __bool__(self):
        self.calls += 1
        return self.calls == 1

flag = Flag()
D = msgspec.defstruct("D", ["x"], frozen=flag)
print(flag.calls, D.__struct_config__.frozen)
# This PR: 2, False. Main and equivalent class syntax: 1, True.
```

The documented flags are booleans, so this is an edge case. Please remove the duplicate conversion while preserving the distinction between omitted and supplied arguments, so metaclass defaults still apply, and add a regression test.

PR #1168, opened 2026-09-01
Fix unquoted Struct forward references on Python 3.14

### LINKED ISSUE #1165: Python 3.14: unquoted forward references fail during Struct class creation
### Description

```
### Description

On Python 3.14, an unquoted forward reference in a `msgspec.Struct`
field raises `NameError` while the Struct class is being created.

Python 3.14's default deferred-annotation behavior normally permits this
without quoting. The equivalent example using ordinary classes works.

Tested with:

- CPython 3.14.0
- msgspec 0.21.1 (currently the latest release on PyPI)

### Minimal reproducer

```python
import msgspec


class Outer(msgspec.Struct):
    inner: Inner


class Inner(msgspec.Struct):
    value: int
```

### Actual result

```text
  File "<string>", line 6, in <module>
    class Outer(msgspec.Struct):
        inner: Inner
  File "<string>", line 7, in __annotate__
    inner: Inner
           ^^^^^
NameError: name 'Inner' is not defined
```

### Expected result

I expected `Outer` to be created successfully, with `Inner` resolved after
the module finishes defining it, as happens with an ordinary Python 3.14
class:

```python
class Outer:
    inner: Inner


class Inner:
    value: int


assert Outer.__annotations__ == {"inner": Inner}
```

### Workarounds

Either of the following prevents the error:

```python
class Outer(msgspec.Struct):
    inner: "Inner"
```

or:

```python
```

Both workarounds use stringified annotations instead of Python 3.14's
default deferred-annotation behavior.

### Additional context

This appears related to:

- #651
- PR #852

PR #852 added Python 3.14 annotation support, but Struct field collection
currently invokes the class annotate function with
`annotationlib.Format.VALUE`:

https://github.com/msgspec/msgspec/blob/main/src/msgspec/_core.c#L6110-L6133

Python's `annotationlib` documentation recommends that metaclasses
inspecting annotations during class construction generally use
`Format.FORWARDREF`, because names may not yet be resolvable:

https://docs.python.org/3.14/library/annotationlib.html#using-annotations-in-a-metaclass

Would using `Format.FORWARDREF`, followed by the existing delayed
forward-reference resolution, be appropriate here?

A practical consequence is that Ruff's UP037 rule considers the explicit
quotes unnecessary on Python 3.14, but removing them can make modules
containing `msgspec.Struct` schemas fail to import.
```

vyrnsynx:
Fixes #1165.

On Python 3.14, Struct field collection invoked the class __annotate__ function with annotationlib.Format.VALUE. That evaluates annotations immediately, so an unquoted forward reference raises NameError while the class is still being created. Ordinary 3.14 classes do not do this.

This change evaluates annotations through annotationlib.call_annotate_function(..., Format.FORWARDREF), matching the annotationlib metaclass guidance. Compiler-generated annotate functions often raise NotImplementedError if FORWARDREF is passed directly; call_annotate_function implements that format.

Field names (and ClassVar detection) still run at class creation time; types continue to resolve later through the existing delayed annotation path.

Quoted annotations and from __future__ import annotations are unchanged. Older Python versions are unaffected because they never take the __annotate__ path.

Tests cover quoted forward refs, future-annotations forward refs, unquoted forward refs and self-recursive structs on 3.14+, JSON encode/decode after both types are defined, and existing ClassVar cases on 3.14.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -30,6 +30,8 @@
 - Fix the `msgspec.json.decode` docstring to say `dec_hook` should raise
   `NotImplementedError` for unsupported types, not `TypeError` ({issue}`774`).
 - Fix backing type declaration of `Ext.code` ({pr}`1135`).
+- Fix `NameError` when creating a `Struct` with an unquoted forward
+  reference on Python 3.14 ({issue}`1165`).
 - msgspec moved to the [msgspec GitHub organization](https://github.com/msgspec/msgspec);
   documentation now lives at [msgspec.dev](https://msgspec.dev) (repository
   references updated in {pr}`1045`).
--- a/src/msgspec/_core.c
+++ b/src/msgspec/_core.c
@@ -6120,7 +6120,12 @@ structmeta_collect_fields(StructMetaInfo *info, MsgspecState *mod, bool kwonly) 
             Py_DECREF(annotate);
             return 0;
         }
-        PyObject *format = PyLong_FromLong(1);  /* annotationlib.Format.VALUE */
+        /* Format.FORWARDREF (3) rather than Format.VALUE (1): unresolved
+         * names must not raise NameError while the class body is still
+         * executing. See PEP 649 / annotationlib metaclass guidance:
+         * https://docs.python.org/3/library/annotationlib.html#using-annotations-in-a-metaclass
+         */
+        PyObject *format = PyLong_FromLong(3);  /* annotationlib.Format.FORWARDREF */
         if (format == NULL) {
             Py_DECREF(annotate);
             return -1;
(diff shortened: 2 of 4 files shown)


provinzkraut (dismissed):
Thanks!

Siyet (approved):
Re-approving after the dismissal. The only change since @provinzkraut's approval is the `ruff I001` import regrouping in `_utils.py`; the fix, the module state ref, and the tests are untouched.

Checked on my side: `call_annotate_forwardref` is added to `msgspec_traverse`, `msgspec_clear` and `SET_REF` consistently with the neighbouring refs, and the pre-3.14 branch keeps the old `Format.VALUE` behavior. Tests cover the quoted, `from __future__ import annotations`, and 3.14-only unquoted cases including a self-referencing struct.

The red `Deploy preview` is the fork-PR docs preview job, unrelated to this change.

Thanks for the fix and for the changelog entry.

PR #1170, opened 2026-09-02
Clarify msgspec.Raw unsupported with yaml/toml decode

### LINKED ISSUE #1136: `msgspec.Raw` appears not to work with `msgspec.yaml.decode`
### Question

I couldn't find any note in documentation suggesting `msgspec.Raw` cannot be used with YAML for decoding, so i assumed it is supported, but i get `ValidationError`. Is this expected behavior?

Minimal reproduction:

```py
import msgspec

class Repro(msgspec.Struct):
    tools: dict[str, msgspec.Raw]

for value in (
    b"tools:\n  x: 1\n",
    b"tools:\n  x: hello\n",
    b"tools:\n  x: [a, b]\n",
    b"tools:\n  x: {a: 1}\n",
    b"tools:\n  x:\n    a: 1\n",
):
    msgspec.yaml.decode(value, type=Repro)
```

Result:

```text
msgspec.ValidationError: Expected `any`, got `int` - at `$.tools[...]``
msgspec.ValidationError: Expected `any`, got `str` - at `$.tools[...]`
msgspec.ValidationError: Expected `any`, got `array` - at `$.tools[...]`
msgspec.ValidationError: Expected `any`, got `object` - at `$.tools[...]`
msgspec.ValidationError: Expected `any`, got `object` - at `$.tools[...]`
```

Environment:
- Python: 3.12.13 (main, May 10 2026, 19:30:01) [Clang 22.1.3 ]
- msgspec: 0.21.1

Brohammad:
## Summary
- Raise a clear `ValidationError` when `convert` targets `msgspec.Raw` but receives an already-parsed Python object (as happens with `yaml.decode` / `toml.decode`)
- Document that delayed decoding with `msgspec.Raw` requires the native `json` or `msgpack` decoders

Fixes #1136

## Root cause
`msgspec.Raw` requires substrings of the original serialized bytes. The native `json` and `msgpack` decoders capture these via `*_decode_raw()`. The `yaml` and `toml` modules parse messages fully with PyYAML/tomllib before calling `convert`, so the original byte spans are unavailable.

Previously, `convert` dispatched on the source Python type and produced a confusing error like `Expected \`any\`, got \`int\``.

## Test plan
- [x] Added `test_raw_from_parsed_object_error` in `test_convert.py`
- [x] Added `test_decode_raw_field_unsupported` in `test_yaml.py`
- [x] Existing Raw tests for json/msgpack/convert still pass (805 tests)


Siyet (commented):
I found no correctness blocker for the field-value case in #1136. Please coordinate this change with #1155 and #1169:

* In #1155, `MS_TYPE_RAW` is used for Raw, so the `type->types == 0` condition will no longer match this type. Whichever PR is merged second will need an updated condition, preserving Raw/None union behavior.
* String dictionary keys are handled through `convert_str()` without a call to `convert()`. Therefore, the error for `convert({"x": 1}, dict[msgspec.Raw, int])` remains `Expected any, got str` in this version. That diagnostic is corrected in #1169. The two fixes cover different execution paths.

Issue #1110, opened 2026-07-02
CodSpeed: test_pickle_load[arm] is bimodal and randomly flags unrelated PRs


Siyet:
`test_pickle_load[arm]` (added in #1092) is bimodal: it flips between ~1.6 ms and ~1.8 ms across commits that don't touch the pickle path at all, so with the 2% threshold it randomly paints unrelated PRs red.

Evidence from this week:

- #978 (decimal_format encoder option, head 045a25c vs main b36a4a1): flagged **-10.05%** (1.6 -> 1.8 ms), check red. The diff cannot affect struct pickling.
- #1105 (to_builtins frozendict fix, head 5a54d59 vs main 8f7ac06): same benchmark reported **+10.28%** (1.8 -> 1.6 ms), check green (improvements don't gate).
- The baseline itself flipped between those two comparisons: main measured 1.6 ms at b36a4a1 (06-27) and 1.8 ms at 8f7ac06 (06-30), with exactly one commit in between (#1052, frozendict) that doesn't touch the pickle path either.

Since this is simulation mode (instruction counting), it's not run-to-run measurement noise: the benchmark appears sensitive to code layout/alignment in `_core.c`, so any code addition can toggle it between the two states. Related to the general variance discussion in #1075 and CodSpeedHQ/codspeed#409.

Options (all live in the CodSpeed dashboard, admin only - there's no repo-side config for this):

1. Set a per-benchmark regression threshold for `test_pickle_load[arm]` above the flip amplitude (say 15%). Keeps measuring, stops the false gating.
2. Ignore the benchmark entirely in CodSpeed.
3. The repo-wide "informational status check" toggle, which would also address the "CI looks perpetually broken" complaint from #1097.

Option 1 seems the most targeted. Worth checking whether the sibling `test_pickle_*` benchmarks from #1092 show the same behavior before picking.

provinzkraut:
This isn't specific to that benchmark. It's an issue with running the benchmarks in different envs. We need to revisit the continuous benchmarking approach, since codspeed could not delivery on its promises regarding stability.

Siyet:
Might well be both. For the #978 flag specifically the report had no "different runtime environments" warning, and both baseline points (b36a4a1 at 1.6 ms, 8f7ac06 at 1.8 ms) were regular post-#1079 pushes to main, with the flip caused by a commit that doesn't touch the pickle path at all - which looks like code-layout sensitivity of simulation mode rather than env variance.

[...]

PR #1172, opened 2026-09-08
Correct error messages for out-of-range Meta length bounds

Siyet:
Calling `Meta(min_length=10**400)` results in a `ValueError` with the incorrect message "must be >= 0" for a positive integer. With this fix, out-of-range `min_length` and `max_length` values result in an error identifying the range issue. Accepted values and diagnostics for representable negative values are unchanged.

This is the length-bound fix discussed in #1122, separate from the numeric constraints work in #1006.


Siyet:
All 22 workflow checks passed, with four publishing/deployment steps skipped. The remaining red status is CodSpeed Performance Analysis: only `test_pickle_load[arm]` moves from 1.6 ms to 1.8 ms (-10.31%), matching the two states documented in #1110. The changed helper is called only by `Meta.__new__` for min_length/max_length validation; it is not executed by this pickle benchmark.

As an additional check I ran the same 1000-Struct pickle-load workload with pyperf on separate freshly built main and patched environments, CPython 3.14.6 on macOS. Main: 98.6 us +/- 4.7 us; patched: 98.1 us +/- 1.1 us. pyperf reports the difference as not significant. This does not establish the cause of a Linux simulation result, but it does not reproduce a 10% walltime regression. I am leaving the PR open with that check explicitly outstanding rather than treating the whole check suite as green.

PR #1173, opened 2026-09-08
Declare Struct instance defaults in the type stub

### LINKED ISSUE #779: `__struct_defaults__` is missing from Python type specification
### Description

I am attempting to work around #778 by writing a recursive Struct traversal in Python as an alternative to `asdict` (this is going to be slow, but slow-and-working is better than not working).

From looking at the C code, `__struct_defaults__` is available alongside `__struct_fields__` (aligning from the end of the two sequences).

However, the Python typing stub for the C code is missing this attribute, so mypy et al complain about accessing the field.

vzer200:
`__struct_defaults__` is available on both Struct classes and instances, but only the metaclass stub declares it. Access through an instance therefore fails type checking even though it works at runtime.

Add the missing Struct annotation and cover class and instance access in the existing typing examples. Runtime behavior is unchanged.

Fixes #779.

Validation: the new instance assertion fails with mypy, pyright, and pyrefly before the change; all three typing suites and stubtest pass afterward. On CPython 3.12, all 6,412 unit tests and 7 package doctests pass, with 118 skips; the repository's lint, format, and spelling checks pass. Windows validation used UTF-8 mode for the existing C-source text checks and a private LLVM-MinGW compiler adapter to build the unchanged extension. Neither is included in this patch.

Implemented and independently reviewed by automated agents before submission.

### DIFF
--- a/src/msgspec/__init__.pyi
+++ b/src/msgspec/__init__.pyi
@@ -96,6 +96,7 @@ def field(*, name: str | None = None) -> Any: ...
 @dataclass_transform(field_specifiers=(field,))
 class Struct(metaclass=StructMeta):
     __struct_fields__: ClassVar[tuple[str, ...]]
+    __struct_defaults__: ClassVar[tuple[Any, ...]]
     __struct_config__: ClassVar[structs.StructConfig]
     __struct_encode_fields__: ClassVar[tuple[str, ...]]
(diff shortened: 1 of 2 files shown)


Siyet (approved):
Thx for the fix! The type stub now matches the existing runtime behavior.

PR #1174, opened 2026-09-08
Build the full wheel matrix weekly, not only at release

### LINKED ISSUE #1150: musllinux wheels only build on release, so breakage surfaces at release time
Coming out of #1134, where the 0.22.0 release stopped on a wheel job. The specific breakage is fixed by #1146, but the reason nobody saw it coming is still there, so I wanted to write it down.

## The gap

`build-wheels.yml` sets:

```yaml
CIBW_SKIP: "${{ !inputs.build-all && '*-musllinux_*' || '' }}"
```

`build-all` is only true on `release`. So no musllinux wheel, on any architecture, is ever built outside a release. The first time a musllinux problem can appear is the run that is supposed to publish.

riscv64 has a second gate on top of that. The job is excluded on pull requests without the `build-riscv64` label, so an unlabelled PR touching the wheel workflow builds no riscv64 wheels at all.

The two compose badly. #987 added riscv64 back in March, and `cp310-musllinux_riscv64` first executed on the 0.22.0 release build ([run 31476259980](https://github.com/msgspec/msgspec/actions/runs/31476259980/job/93742042390)), where it failed on its first step. #1146 could not be validated by its own CI either, for the same reason, so it was merged on a reading of the logs and confirmed only by the next release run.

This is not really a riscv64 complaint. musllinux on x86_64 and aarch64 has the same exposure, it just has not broken yet.

## Options

I do not have a favourite, and you know the cost tradeoffs here better than I do:

- A scheduled run of the wheel workflow with `build-all: true`, weekly or so. Catches image and toolchain drift between releases, at the cost of one long job on a timer.
- Build musllinux for one architecture on PRs that touch packaging files, rather than all of them all the time.
- Leave it, and treat a failed release build as an acceptable way to find out. That is a legitimate answer if the extra CI time is not worth it, given releases are infrequent.

## A smaller related thing

The cibuildwheel version is an inline argument:

```yaml
uvx --from cibuildwheel==4.1.0
```

`dependabot.yml` covers the `github-actions` and `uv` ecosystems, and neither sees that string, which is how it stayed on 4.1.0 for the two months after the image it pins was superseded. I have not thought of a tidy way to make it trackable, so this is a note rather than a proposal.

For context, I am the one who added the riscv64 entry, so this is me flagging the blind spot my own change has been sitting behind rather than pointing at anyone else's.

Yasser-Ameur:
`build-wheels.yml` skips every musllinux wheel unless `build-all` is true, and `ci.yml` only passes `build-all` on release. So no musllinux wheel, on any architecture, is built outside a release, and the first run that can show a musllinux breakage is the one meant to publish. #1134 hit exactly that when `cp310-musllinux_riscv64` failed on the 0.22.0 release build, and #1146 could not be exercised by its own CI for the same reason.

**Change.** `ci.yml` gains a weekly `schedule` trigger (Monday 04:17 UTC) and passes `build-all` on `schedule` as well as on `release`, so the complete matrix, musllinux and riscv64 included, runs on a timer between releases. Pull request and push runs are unchanged, and so is the `build-riscv64` label gate. Five lines, no new jobs.

**What proves it.** A workflow cannot run locally; the file parses as YAML. The first scheduled run showing the musllinux and riscv64 legs outside a release is the proof. If PR-time coverage is wanted as well, a single `cp312-musllinux_x86_64` smoke build gated on changes to the wheel workflow, `setup.py`, `pyproject.toml` or `src/` is a small follow-up; I left it out of this change so the schedule can land on its own.

One question for the reviewer: is weekly the right cadence for the project's CI budget, or would you rather tie the full build to a pre-release tag?

Closes #1150

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,6 +35,10 @@ on:
   release:
     types:
     - published
+  schedule:
+    # Weekly full wheel build (build-all), so a musllinux or riscv64 breakage
+    # surfaces before the release run that would otherwise be the first to try it.
+    - cron: '17 4 * * 1'
 
 permissions: {}
 
@@ -201,7 +205,7 @@ jobs:
       contents: read
     with:
       sdist-name: "${{ needs.prepare.outputs.sdist-name }}"
-      build-all: ${{ github.event_name == 'release' }}
+      build-all: ${{ github.event_name == 'release' || github.event_name == 'schedule' }}
 
   upload-pypi:
     name: Upload artifacts to PyPI


Siyet (approved):
Thx for the fix! Weekly sounds good to me. This should help catch wheel build failures before a release

PR #1179, opened 2026-09-13
Fix Meta.tz stub annotation to `bool | None`

vmphase:
The C implementation validates it with `ensure_is_bool` and the docstring documents it as `tz: bool, optional`.

```python
import msgspec

m = msgspec.Meta(tz=True)

is_tz_aware: bool = m.tz
# Type "int | None" is not assignable to declared type "bool"
# "int" is not assignable to "bool" (basedpyright reportAssignmentType)

print(f"m.tz = {m.tz!r}, runtime = {type(m.tz).__name__}")
# m.tz = True, runtime = bool
```


Siyet (commented):
Thx for the fix! Please add `assert_type(c.tz, bool | None)` to `check_meta_attributes` so the getter type is covered by a regression test

Siyet (approved):

PR #1183, opened 2026-09-15
Fix json schema defaults for set / bytearray default_factory fields

### LINKED ISSUE #1182: json.schema() generates non-serializable default values
### Description

`msgspec.json.schema()` puts a raw `set()` / `bytearray()` object into the schema's `default` for `Struct` fields using `default_factory=set` or `default_factory=bytearray`, so the result isn't json-serializable:

```python
import json

import msgspec
from msgspec import field


class Config(msgspec.Struct):
    features: set[str] = field(default_factory=set)


json.dumps(msgspec.json.schema(Config))
# TypeError: Object of type set is not JSON serializable
```

I would expect the default to be a json value (e.g. `[]`), like it already is for `default_factory=list/dict`.

vmphase:
Fixes #1182
Now values are json-serializable: `[]` / `""` instead of raw `set()` / `bytearray()`.


Siyet (changes requested):
Thanks for the quick fix. The change looks right to me.

Two things before this can be merged.

The format check fails on the new line, so CI is red and the test jobs were skipped entirely. Pls run `just hooks` and push the reformatted version.

The change also needs a regression test. There is currently no coverage for `set` or `bytearray` factory defaults in `tests/unit/test_schema.py`, so a future regression would go unnoticed. A test parametrized over `list`, `dict`, `set` and `bytearray` asserting the emitted `default` and calling `json.dumps` on the resulting schema would cover it, with `list` and `dict` as controls.

vmphase:
> Thanks for the quick fix. The change looks right to me.
> 
> Two things before this can be merged.
> 
> The format check fails on the new line, so CI is red and the test jobs were skipped entirely. Pls run `just hooks` and push the reformatted version.
> 
> The change also needs a regression test. There is currently no coverage for `set` or `bytearray` factory defaults in `tests/unit/test_schema.py`, so a future regression would go unnoticed. A test parametrized over `list`, `dict`, `set` and `bytearray` asserting the emitted `default` and calling `json.dumps` on the resulting schema would cover it, with `list` and `dict` as controls.

Hello. Thanks for the review, addressed both changes in the two commits above.

Siyet (approved):
Thx for the fix. Both points are addressed.

Worth noting the fix covers more than the test shows: a bare `tags: set[str] = set()` becomes factory internally, so it produced the same unusable value before this change.

PR #1190, opened 2026-09-17
Document PEP 695 type parameter syntax for generic types

Siyet:
The "Generic Types" section describes only the `TypeVar` plus `Generic[T]` form, and the instructions present it as the way to define a generic type, even though msgspec also accepts Python 3.12's type parameter syntax.

This adds a short paragraph and example for the `class Paginated[ItemT](msgspec.Struct)` form after the existing instructions, and extends the references at the end of the section with a link to PEP 695. The existing steps are left as they are: the `Generic[T]` base is required for the `TypeVar` form, and both names there link into the Python docs.

Documentation only, no behavior change.


docs/supported-types.rst
```diff
@@ -1284,6 +1284,15 @@ To define a generic type:
 - When annotating the field types, use the relevant type variables instead of
   "concrete" types anywhere you want to be generic.
 
+On Python 3.12+ the type parameters may instead be declared with the type
```
provinzkraut, line 1287: IMO this should be above the other section, and not mentioned as an alternative, since it's the default for generics in newer Python versions. We can then label the existing section as the alternative

Siyet: Done, the 3.12 form now comes first and the TypeVar steps are labeled as the older spelling.

I kept the running example on the TypeVar form, since it has to work on 3.10 and 3.11 too, and said so in one line right after steps. Happy to convert the example as well if you would rather have it the other way.

Issue #1191, opened 2026-09-17
Type argument of a parametrized generic `TypedDict` base is ignored when subclassed


Siyet:
### Description

When a `TypedDict` inherits from a parametrized generic `TypedDict`, the type argument is dropped. The inherited field keeps the type parameter instead of the type it was parametrized with, so values that should be rejected are accepted.

```python
from typing import Generic, TypedDict, TypeVar

import msgspec

L = TypeVar("L")

class Base(TypedDict, Generic[L]):
    b: L

class Sub(Base[int]):
    n: str

print(msgspec.json.decode(b'{"b": "not an int", "n": "a"}', type=Sub))
```

I expected a `ValidationError` for `b`, because `Base[int]` declares it as `int`. What I get is:

```
{'b': 'not an int', 'n': 'a'}
```

The same shape built on a `Struct`, a `dataclass` or an `attrs` class raises ``ValidationError: Expected `int`, got `str` - at `$.b` ``, so this looks specific to `TypedDict`. Decoding `Base[int]` directly, without the subclass, also rejects the value correctly, so the inheritance step is what drops the argument.

The PEP 695 spelling behaves the same way, and under `from __future__ import annotations` it fails with a `NameError` instead:

```python
from __future__ import annotations

from typing import TypedDict

import msgspec

class Base[Q](TypedDict):
    b: Q

class Sub(Base[int]):
    n: str

msgspec.json.decode(b'{"b": 1, "n": "a"}', type=Sub)
```

```
Traceback (most recent call last):
  File "mcve.py", line 13, in <module>
    msgspec.json.decode(b'{"b": 1, "n": "a"}', type=Sub)
    ~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../msgspec/_utils.py", line 241, in get_typeddict_info
    raw_hints = get_class_annotations(obj)
  File ".../msgspec/_utils.py", line 193, in get_class_annotations
    value = _eval_type(value, cls_locals, cls_globals)
  File ".../msgspec/_utils.py", line 49, in _eval_type
    return typing._eval_type(t, globalns, localns, ())
           ~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../python3.14/typing.py", line 461, in _eval_type
    return evaluate_forward_ref(t, globals=globalns, locals=localns,
                                type_params=type_params, owner=owner,
[...]

PR #1192, opened 2026-09-19
Fix inherited generic TypedDict field validation

Neallin-917:
## Summary

Addresses the inherited generic TypedDict validation case in #1191.

TypedDict copies inherited annotations into the subclass and removes the original bases from its runtime MRO. The existing annotation walker therefore sees the inherited TypeVar without the base's specialization. Resolve original TypedDict bases in their own generic scopes and reuse those hints for unchanged inherited fields, preserving explicit field overrides and separate bindings across multiple bases.

Regression coverage includes JSON and MessagePack, stdlib and typing_extensions TypedDict, scalar and nested list fields, multi-level inheritance, independent base scopes, and an explicit field override.

This is scoped to inherited specialization; it does not implement the PEP 695 postponed-annotation resolution changes proposed in #1130. References #1191 rather than closing the entire report.

## Validation

- New regression tests against the unchanged baseline annotation helper: 6 failed (DID NOT RAISE); with this patch they pass.
- Python 3.12.13: `just test-all -q`: 6454 passed, 116 skipped; doctests 7 passed.
- Python 3.14.0: full unit suite: 6457 passed, 113 skipped; focused TypedDict/generic utility tests: 199 passed.
- Python 3.11.15 TypedDict tests: 45 passed.
- Python 3.10.20 TypedDict tests: 41 passed, 4 skipped (unsupported stdlib features).
- `just check`: lint, format and spelling checks passed; `git diff --check` passed.

## AI assistance

Codex assisted with investigation, implementation, regression tests, and automated validation. No independent human review is claimed.


Siyet:
Thanks for this. The fix itself holds up: on a build of current main the inherited binding is dropped in both the `TypeVar` plus `Generic` spelling and the PEP 695 one, and on this branch the invalid value is rejected in both.

The seven wheel jobs failed, all of them on one thing: `test_inherited_generic_typeddict_scopes[json]` and `[msgpack]`, with `ModuleNotFoundError: No module named 'typing_extensions'`.

That test builds a module source starting with `from typing_extensions import TypedDict`, and `temp_module` execs that source unconditionally, while the cibuildwheel test environment installs only the `test-unit` group, which does not list `typing_extensions`. Every other `typing_extensions` site in this file sits behind `pytest.importorskip("typing_extensions")`, including the other test you added and `test_recursive_generic_typeddict` just below this one, which uses the same `temp_module` shape. Adding that one guard line at the top of the test body should be enough.

This one is easy to miss locally: the dev test environment pulls in `typing_extensions` as a transitive dependency of the type checkers, so the test runs and passes there.

The red `Deploy preview` is unrelated to your change. It ran before the docs workflow did, and only needs a re-run.

Neallin-917:
Thanks for identifying the wheel-test environment difference. Added the optional-dependency guard in 8a681cb. With only the test-unit dependency group, both cases failed before the change; the full unit suite now passes (6,420 passed, 150 skipped). With the full development dependencies, both cases execute and pass, and test-all reports 6,454 passed, 116 skipped plus 7 doctests. Lint, formatting and spelling checks pass. The seven-platform wheel matrix still needs the upstream run; these local results are not a claim that all wheel jobs passed. This follow-up was implemented and validated with Codex assistance.

